PR #1321, opened 2015-02-19
Exporttocellh5

sommerc:
Hey Lee,

I was working on the exportToCellH5 module. The basic version is running. Please, pull the current cellh5 master to get the preleminary cellh5write.py API.

I used the example pipeline you forwarded me and ignored time so far.

Before I continue, would be happy to hear your opinion about it!

Cheers, Christoph


sommerc:
to test the output you can use a preliminary fiji version containing cellh5 reader in bioformats:

http://gerlichlab.imba.oeaw.ac.at/software/cellh5/Fiji.app-cellh5.zip

Note: the master file does not work with that fiji version yet

cellprofiler/modules/exporttocellh5.py
```diff
@@ -366,35 +392,84 @@ def keep(column):
...
-                #
-                # store a measurement with feature name "feature_name"
-                # and value, "value"
-            else:
+        
+        ### 0) extract object information (i.e. object_label_no) 
+        ### 1) extract all single cell features and write it as feature matrix (for e.g. classification)
+        ### 2) extract Center
+        ### 3) create artifical Bounding box... usefull for displaying it in fiji lateron
+        ### 4) Don't see the point of features extracted on "Image" the only real and useful feature there is "Count" which can be deduced from single cell information
```
LeeKamentsky: There are often useful image-wide features, for instance, there is a module that measures image Q/A parameters to detect images that are out of focus or have a large number of saturated pixels (indicating debris in well). 

I am not sure why you're excluding some object measurements. There may be some worth excluding, but I'd pick the ones to leave out, not the ones to accept in case some user adds their own custom module. Also, the UI for the module lets users pick which measurements to save, so that would be the preferred way of cutting the measurement list down.

cellprofiler/modules/exporttocellh5.py
```diff
@@ -366,35 +392,84 @@ def keep(column):
...
+            
+            c5_feature_writer = c5_pos.add_object_feature_matrix(object_name=object_name, feature_name="object_features", n_features=n_features, dtype=numpy.float32)
+            c5_feature_writer.write(feature_matrix)
+            c5_feature_writer.write_definition(feature_names)
+            c5_feature_writer.finalize()
+            
+            ### iterate over Location  to create bounding_box and center
+            c5_bbox = c5_pos.add_object_bounding_box(object_name=object_name)
+            
+            location_x = m[object_name, "Location_Center_X"][:, numpy.newaxis]
```
LeeKamentsky: After this is merged, I'll amend it to do a more accurate bounding box using the actual segmentation.

LeeKamentsky:
I'm having problems installing cellh5. The setup script doesn't install
cellh5write or hmm_wrapper, hope you can fix.

On Thu, Feb 19, 2015 at 12:39 PM, Christoph Sommer <notifications@github.com

cellprofiler/modules/exporttocellh5.py
```diff
@@ -306,21 +310,19 @@ def handle_interaction(self, master_file, pid, path):
         master_dict = self.get_dictionary().setdefault(master_file, {})
         if pid not in master_dict:
             md_head, md_tail = os.path.splitext(master_file)
-            subfile = "%s_%s%s" % (md_head, str(pid), md_tail)
```
LeeKamentsky: I am worried that this will create thousands of small files in a single directory. My goal here was to have one .h5 file per worker and let the worker write multiple records to that file. Also, the user may not have site metadata and can choose a site metadata item of "None". The code fails in that case because the path has only two items. Would everything work if this line were subfile="_".join([md_head] + path + [md_tail])?

LeeKamentsky:
I've merged this with some of my changes, let me know what you think.

cellprofiler/modules/exporttocellh5.py
sommerc: Q/A paramteres are indeed usefull, I think of how to add this to cellh5. We haven't had a good use case for this, so far. Definitely, I want to avoid storing "all ones" pixel images to it

cellprofiler/modules/exporttocellh5.py
sommerc: Yes. One file per worker should be the goal. So far, several channels can be written by a worker. The next would be to add temporal support, and - if available - write a movie. How is it with z-slices? 

In case, there is no meta data (for plate, well, site), what should I use? Cellh5 requires this information as a key for the image... I could just take Constant_place holders plus an auto-incrementing digit!?

If only site is missing, this should default to sth constant "1" or "0"..., I suppose

sommerc:
updated cellh5/setup.py to include the missing files

cellprofiler/modules/exporttocellh5.py
LeeKamentsky: If there's no metadata for a category, I think I'm going to display a setting that lets the user type in a custom metadata entry. I'll default to "1" for site, "A01" for well and "Plate" for plate.

LeeKamentsky:
Christoph, I think we need to reorganize things a bit. The definitions should happen at an earlier stage and should be done in the master file - otherwise CellH5 as it is raises exceptions when it tries to overwrite them. I can change the code to write the defs to both the master and subfiles so that each subfile is useable by itself, but my gut tells me that the official defs should be in the master file.

As it is, on the second cycle, each worker dies as it tries to overwrite the defs. I can do the work, just not today.

PR #1353, opened 2015-05-04
ljosa -> cellprofiler as the real docker repo

purohit:
It appears https://registry.hub.docker.com/u/cellprofiler/cellprofiler/ is the real repository, no longer the one belonging to `ljosa`. I've updated that in the README.


LeeKamentsky:
Thanks ashwin. I'm out until Friday, will merge then. If you run into any
problems let me know.

PR #1367, opened 2015-05-24
Yit

sstoma:
Dear Lee,

better late than never, we finished with the devs. The code contains tests, was tested and affects only minimally CP core (TrackObjects is touched and should be reviewed by you with extra care; some necessary changes to integrate YeastSegmentation module were added to core e.g. registering modules; lastly or our helpers are in contrib, though it should be more relaxed). We would be very happy to see our module in one of the nightly build versions :) In case you see some issues, please let me know, we will solve them. Cheers, s/


LeeKamentsky:
Starting work on pull request
@0x00B1 @sstoma - We're refactoring CellProfiler into smaller packages. There are a lot of reasons why contrib/cell_star should be its own package (authorship is more directly that of the contributors / it holds together as a package / we want contrib to disappear). I'll make that happen if everyone is OK with that.

Similarly, move neighmovetrack.py to the [centrosome](https://github.com/CellProfiler/centrosome) package.

@sstoma My initial take is that the quality is higher now than when you visited.

sstoma:
We are super flexible on the final integration scheme - please feel free to do as it suits you guys. We simply would like to see the plugin in the builds, since we can really test it then on different platforms and ask people for ideas and contributions.

cellprofiler/modules/yeast_cell_segmentation.py
```diff
@@ -0,0 +1,892 @@
...
+    def ground_truth_editor( self ):
+        '''Display a UI for GT editing'''
+        from cellprofiler.gui.editobjectsdlg import EditObjectsDialog
+        from wx import OK
+        import wx
+        #title = "%s #%d, image cycle #%d: " % (self.module_name,
+        #                                     self.module_num,
+        #                                     image_set_number)
+        
+        ### opening file dialog
```
LeeKamentsky: wx.FileDialog is not destroyed. Maybe use a "with" block:

    with wx.FileDialog(...) as dlg:

contrib/cell_star/utils/image_util.py
```diff
@@ -0,0 +1,402 @@
...
+    @returns [ystart,ystop), [xstart,xstop) or None, None
+    """
+    non_zero_points = argwhere(image_mask)
+    if len(non_zero_points) == 0:
+        return None
+    (ystart, xstart), (ystop, xstop) = non_zero_points.min(0), non_zero_points.max(0) + 1
+    return (ystart, ystop), (xstart, xstop)
+
+
+def get_circle_kernel(radius):
```
LeeKamentsky: I have an instance where get_circle_kernel is being called with a value of 1.36. The kernel should be symmetric around the center point, but instead the value is:

    [[ False, False, False],
     [False, True, True],
     [False, True, True]]

The problem is that `np.ogrid[-radius:radius+epsilon]` returns the values, -1.36, -.36, .64, consistent with adding 1 to -1.36. The grid should instead be
 
    np.mgrid[np.floor(-kernel_size):np.ceil(kernel_size + .001) + 1,
        np.floor(-kernel_size):np.ceil(kernel_size + .001) + 1]

contrib/cell_star/utils/image_util.py
```diff
@@ -0,0 +1,402 @@
...
+    """
+    nr = 1
+    morphology_element = get_circle_kernel(kernel_size)
+    while True:
+        new_mask = mask.copy()
+        # find connected components
+        components, num_components = sp.ndimage.label(np.logical_not(new_mask), np.ones((3, 3)))
+        slices = sp.ndimage.find_objects(components)
+        for label, slice in zip(range(1, num_components + 1), slices):
+            slice = extend_slices(slice,kernel_size+1000)
```
LeeKamentsky: I think that

    slice = extend_slices(slice, morphology_element.shape[0] * 2)

should be sufficient and should speed this up dramatically?

contrib/cell_star/utils/cluster_util.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
+__author__ = 'Adam Kaczmarek, Filip Mróz'
+
+# External imports
+import numpy as np
+# Internal imports
+from calc_util import euclidean_norm
+from contrib.cell_star.core.seed import Seed
+
+
```
LeeKamentsky: pdist is too slow. You can do the cross-product distance like this

    def pdist(list_points, _):
          pts = np.asanyarray(list_points)
          return np.sqrt(np.sum(pts[:, np.newaxis, :] - pts[np.newaxis, :, :], 2))
        return

contrib/cell_star/core/seeder.py
```diff
@@ -0,0 +1,309 @@
...
+
+    def __init__(self, images, parameters):
+        self.images = images
+        random.seed(abs(np.sum(images.image)))
+        self.parameters = parameters
+        self.cluster_min_distance = self.parameters["segmentation"]["seeding"]["minDistance"] \
+            * self.parameters["segmentation"]["avgCellDiameter"]
+        self.random_radius = self.parameters["segmentation"]["seeding"]["randomDiskRadius"] \
+            * self.parameters["segmentation"]["avgCellDiameter"]
+
```
LeeKamentsky: I entered this routine while segmenting a sample yeast image with ~1500 seeds, all of which were initially less than the cluster_min_distance from some adjacent seed. I am guessing that the internal loop is going to run at least 750 times before it converges on a cluster. Ideally, a Delaunay Triangulation could be used to get the adjacent neighbors of a seed and a merge could be done just considering those neighbors and could be done in-place on that data structure. That would be fastest since only a few minimum distances would need to be recomputed instead of 1500 * 2 per step and the memory moving (in this case, ~ 1Mb per iteration) required by the delete step on line 55 could be avoided.
An intermediate approach is to process the array of 1500 at once, coalescing all seed pairs that are closest to each other and less than the cluster_min_distance apart from each other. That can all be done in Numpy and hopefully the first iteration of the loop. In my case, 400 pairs are closest to each other (i is closest to j and j is closest to i), reducing the array on the first step to 1100 cases. Hopefully, the clustering should be complete after a few iterations and I am pretty sure you are guaranteed to converge.

contrib/cell_star/utils/image_util.py
```diff
@@ -0,0 +1,402 @@
...
+    component = np.zeros(mask.shape, dtype=bool)
+    for label in range(1, num_components + 1):
+        np.equal(components, label, component)
+        if np.count_nonzero(component) > min_area_size:
+            tmp_mask = image_dilate(component, dilate_radius)
+            mask = mask | tmp_mask
+
+    return mask
+
+
```
LeeKamentsky: Consider using centrosome.cpmorphology.fill_labeled_holes after labeling.

.gitignore
```diff
@@ -38,3 +38,4 @@ killjavabridge.egg-info
 plugins
 output
 windows/**
+.idea/**
```
0x00B1: Exclude .gitignore changes.. They’re unrelated.

cellprofiler/gui/editobjectsdlg.py
```diff
@@ -2068,20 +2069,18 @@ def close_label(self, label, display = True):
                 self.display()
                 
 if __name__== "__main__":
```
0x00B1: This is an artifact and should be removed.

contrib/cell_star/config/__init__.py
```diff
@@ -0,0 +1,2 @@
+# -*- coding: utf-8 -*-
+__author__ = 'Adam Kaczmarek, Filip Mróz'
```
0x00B1: This should be broken out into a separate package and repository.

contrib/cell_star/core/snake_filter.py
```diff
@@ -0,0 +1,89 @@
...
+                        curr_snake.area = in2.sum() + Snake.epsilon
+                        avg_inner_darkness = float(np.count_nonzero(tmp)) / float(curr_snake.area)
+                        if avg_inner_darkness < self.parameters["segmentation"]["minAvgInnerDarkness"]:
+                            logger.debug(log_message.format(snakes_indices[i], 'too low inner darkness', '...'))
+                        else:
+                            if curr_snake.area > (self.parameters["segmentation"]["maxArea"] * self.parameters["segmentation"]["avgCellDiameter"]**2 * math.pi / 4):
+                                logger.debug(log_message.format(snakes_indices[i], 'too big area', str(curr_snake.area)))
+                            else:
+                                if curr_snake.area < (self.parameters["segmentation"]["minArea"] * self.parameters["segmentation"]["avgCellDiameter"]**2 * math.pi / 4):
+                                    logger.debug(log_message.format(snakes_indices[i], 'too small area:', str(curr_snake.area)))
```
LeeKamentsky: All of my snakes are being filtered out here at the first step in run_segmentation. I am wondering if they are being filtered out before they have a chance to grow larger.

contrib/cell_star/core/snake_filter.py
```diff
@@ -0,0 +1,89 @@
...
+                                else:
+                                    max_free_border = self.parameters["segmentation"]["stars"]["points"] * self.parameters["segmentation"]["maxFreeBorder"]
```
LeeKamentsky: If I set my maxArea to be very small, all my snakes are filtered out here.

contrib/cell_star/core/snake.py
```diff
@@ -0,0 +1,470 @@
...
+        #
+        # Wygładzanie konturu
+        #
+        #
+
+        # Posortuj punkty konturu (długości promieni) względem kolejnych kątów
+        s = sorted(range(len(ymins)), key=lambda k: ymins[k])
+        xmins = xmins[s]
+        # Wyznacz maksymalne różnice pomiędzy kolejnymi punktami konturu
+        max_diff = np.array(abs(smoothness) * np.arange(1, steps+1) / points_number + 0.5, dtype=int)
```
LeeKamentsky: I am getting a max_diff of zero here. Are you sure about dtype=int? I think this is where I lose all my snakes.

LeeKamentsky:
@sstoma do you have an example image for me to test this on? I am not having much luck on a yeast example image that looks like it should be reasonable.

I'm having problems both with the length of time it takes the yeast segmentation to run and with its ability to properly segment. My initial take is that the method is using some very inefficient loops which increase processing time by perhaps a factor of 100 to 1000, that it isn't finding very good seeds and that several factors cause the method to discard snakes before they are properly grown.

sstoma:
I am sorry for inconvenience. 

This algorithm is very sensitive for the parameter choice (these parameters are used for a cost function describing the border of a cell and cell surface characteristics; optimization of this function is leading to fitting the contour). I would say that I am not surprised that on a random image it does not work using default parameters. Since parametrization is not trivial task, we added parameter search procedure (autoadapt parameters button; later outline cell or two and run the adaptation of parameters to specific images; the longer you wait the more likely that it will work; to get the algorithm to start working one - two minutes should be enough). I am sorry to describe it here, we will have a clear webpage and screencasts to explain all of this, but we first wanted to see how the plugin looks after integration (in case they were some necessary changes in the GUI).

I attach an image from our benchmark, it should be possible to make the algorithm work on it. I had to convert it to png. In case they are problems with image, please download the benchmark and grab an image or two: https://drive.google.com/file/d/0BwJ7pEc_q5STQ0tyOFJTRGswT3c/edit?usp=sharing

Concerning inefficiency, we will look into this. Thanks for reviewing the code, we appreciate it. Some of the reasons is that the algorithm should be "identical" to Matlab code. Adding optimizations makes  it really difficult to follow correspondence (and we had some issues that both algorithms were giving different outputs for same images and parameters...).

screen shot 2015-10-14 at 20 53 01


bf_frame006

0x00B1:
@sstoma @Fafa87 @adamjankaczmarek moved pull request to #1988.

PR #1398, opened 2015-07-05
Update for wxWidgets 3.x 

radioxoma:
Hello! There are small changes for wxWidgets3.x compatibility and one bugfix for preferences dialog.

* Help window `cellprofiler.gui.htmldialog.HTMLDialog` available by click on *?* button in preferences dialog `cellprofiler.gui.preferencesdlg.PreferencesDlg` was inactive (impossible click *OK* or even just close window), because parent (`PreferencesDlg`) was modal.
    * Now `PreferencesDlg` is nonmodal with separate `save_preferences` slot called by both *OK* and *Apply* buttons.

Tested with Archlinux x64, Windows 7 x64 (Anaconda + [gohlke](http://www.lfd.uci.edu/~gohlke/pythonlibs/) libs).

### DIFF
--- a/cellprofiler/cellprofilerapp.py
+++ b/cellprofiler/cellprofilerapp.py
@@ -40,8 +40,6 @@ class CellProfilerApp(wx.App):
         # The wx.StandardPaths aren't available until this is set.
         from cellprofiler.utilities.version import dotted_version
         self.SetAppName('CellProfiler%s' % dotted_version)
-        
-        wx.InitAllImageHandlers()
 
         if self.show_splashbox:
             # If the splash image has alpha, it shows up transparently on
--- a/cellprofiler/cpmath/otsu.py
+++ b/cellprofiler/cpmath/otsu.py
@@ -302,7 +302,6 @@ if __name__=='__main__':
     M_LOG_TRANSFORM = wx.NewId()
     class MyApp(wx.App):
         def OnInit(self):
-            wx.InitAllImageHandlers()
             self.frame = F.CPFigureFrame(title="Otsu",subplots=(2,1))
             file_menu = self.frame.MenuBar.Menus[0][0]
             file_menu.Append(FILE_OPEN,"&Open")
--- a/cellprofiler/cpmath/zernike.py
+++ b/cellprofiler/cpmath/zernike.py
@@ -334,7 +334,6 @@ if __name__ == "__main__":
     
     class MyApp(wx.App):
         def OnInit(self):
-            wx.InitAllImageHandlers()
             self.frame = MyFrame()
             self.SetTopWindow(self.frame)
             self.frame.Show()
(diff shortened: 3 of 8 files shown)


LeeKamentsky:
Sorry - believe me, moving to WX 3.0 is on our list and it does seem very compatible with both Windows and Linux and, if things went well, it lets us use 64 bit mode and Cocoa on OS/X. Sadly, Cocoa and the Java UI don't play well together - I did some exploration and ran into some severe problems with the OS/X menus as well as what seemed to be inevitable segfaults.

In order to accept the pull request, CellProfiler would have to be able to show its menu when CellProfiler's windows were active and ImageJ's windows when ImageJ was active (to test, start CellProfiler, add the RunImageJ module and press the button marked "Show" with the label "Show ImageJ" to the right.

If you can figure out how to get things to work flawlessly on the Mac, we would be deeply indebted to you. Our current plans are to rewrite the UI, probably in Java so that both the ImageJ and CellProfiler UIs would run on the same event loop (see https://github.com/CellProfiler/python-javabridge and https://github.com/LeeKamentsky/backend_swing).

radioxoma:
Latest CP from master falls on my Archlinux, because it's trying to use v3 API, but I installed both [wxpython](https://www.archlinux.org/packages/extra/x86_64/wxpython/) and [wxpython2.8](https://www.archlinux.org/packages/extra/x86_64/wxpython2.8/). May be it's possible to set v2.8 API explicitly with [`wxversion`](http://www.wxpython.org/docs/api/wxversion-module.html) module during packaging (without changing code in repo).

Are you planning totally remove Qt and wx? What about python 3 support?
Seems like I have to use virtual machine. Yosermite x64 is ok? 

P.S. I'll be without computer for few weeks.

LeeKamentsky:
I would go ahead and run with your patch for now. Offhand I don't know if wxversion works with 3.0 - it's a good guess that you'd be able to pick and choose using wxversion.

Regarding QT, we'll probably build it so that the distro can run Ilastik (the pixel classifier) which will run without any Java UI conflict. If we did implement a Java UI, we would remove WX completely. Python 3.x isn't in our immediate plans - it looks like we have until 2020 to do that one :-p jk. We'd like the unicode, but with our use of strings, a big overhaul.

Yosemite x64 is an excellent target and a virtual machine is an excellent choice if you can make it happen. We will probably start doing some exploratory programming on OS/X after the release slated for September - I might pick up the patch on a branch and see if I can hack OS/X myself at that point.

radioxoma:
[This Arch build script](https://projects.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/python2-pyface&id=ce0d5874ecd56c247bbbd4d0fbbf0a3fd148d9d5#n25) (line 25) successfully uses `wxversion`. Seems promising.

UPD. Now pull request is compatible with both wx2.8 and 3.x

LeeKamentsky:
Great. I'll find time to merge this later this week. Thanks for taking the time to make it backwards-compatible. It's at least a first step to 3.0 and perhaps there will be an easy solution somewhere to the OS/X issue.

radioxoma:
Please, merge [same PR](https://github.com/CellProfiler/CellProfiler-Analyst/pull/71) for CellProfiler Analyst.

PR #1464, opened 2015-08-07
Use objgraph from PyPI

### LINKED ISSUE #1438: remove objgraph
Let’s use [objgraph](https://pypi.python.org/pypi/objgraph/2.0.1) from PyPI and remove contrib.objgraph.

0x00B1:


LeeKamentsky:
The Python package, "objgraph", is missing a few things that Ray added when he took it as a contribution: [`get_objs()`](https://github.com/CellProfiler/CellProfiler/blob/master/contrib/objgraph.py#L72), [`by_instance_of()`](https://github.com/CellProfiler/CellProfiler/blob/master/contrib/objgraph.py#L182) [`numpy_label()`](https://github.com/CellProfiler/CellProfiler/blob/master/contrib/objgraph.py#L414) and [`numpy_size()`](https://github.com/CellProfiler/CellProfiler/blob/master/contrib/objgraph.py#L421) are all missing (and probably added by Ray) from the 2.0.1 version I downloaded via PIP.

So, I'm holding off before taking this. I'd copy the functions above to some place that made sense, run the "test_pipeline" unit test which uses objgraph and then try the "Numpy Memory Usage..." command from the Test menu in the CP GUI.

PR #1472, opened 2015-08-10
Replaced utf16encode and utf16decode with str.encode and str.decode 

0x00B1:


cellprofiler/modules/tests/test_saveimages.py
```diff
@@ -656,7 +656,7 @@ def test_00_10_load_v10(self):
         pipeline = cpp.Pipeline()
         image_folder_text = pipeline.encode_txt(
             "%s|%s" % (cpprefs.ABSOLUTE_FOLDER_NAME, 
-                       cpp.utf16encode(cpmt.example_images_directory())))
+                       cpmt.example_images_directory().encode("utf-16")))
```
LeeKamentsky: Unfortunately, cpp.utf16encode wasn't named so correctly. I think this should be "utf-8", as inappropriate as that might seem. There are a lot of unit tests that inadvertently fail because of this and I hope a few that purposefully test for the encoding and fail.

0x00B1: oops! Thanks!

0x00B1:
Fixed UTF-8 issue

LeeKamentsky:
Ugh - I can't believe I wrote those :expletive-deleted: routines. If you run the tests in test_saveimages:

    python cpnose.py --with-javabridge --no-headless -m "(?:^)test_(saveimage|\d).*"

you will see the failures. Maybe the patch to this should be to rename utf16encode and decode to "lees_stupid_encode" and "lees_stupid_decode". The translation should have been `if 0x20 <= c < 0x7f then c else "\u%04x" % c` but it's `if c=="\\" then "\\\\" else...`

PR #1485, opened 2015-08-12
fixes #1417 + tests

### LINKED ISSUE #1417: Export Pipeline Comments
I would find it useful to have a way to export a simple pipeline listing with the Module Notes associated with each module.  This would be useful for CP Forum posts, to list a summary of our posted pipeline, or even for publication, as a nice summary of the analysis steps.  

Something like:

[Module #]. [Module name]: [Module Notes]
[Module #]. [Module name]: [Module Notes]
...

It could live under File > Export > Pipeline Comments

LeeKamentsky:
@0x00B1 here's a pull request for issue #1417. Review and merge?

### DIFF
--- a/cellprofiler/pipeline.py
+++ b/cellprofiler/pipeline.py
@@ -1295,15 +1295,17 @@ class Pipeline(object):
         
         indent - indent of the notes relative to module header.
         '''
+        lines = []
         for module in self.modules(exclude_disabled=False):
             if module.enabled:
-                fmt = "[%4.d] [%s]\n"
+                fmt = "[%4.d] [%s]"
             else:
-                fmt = "[%4.d] [%s] (disabled)\n"
-            fd.write(fmt % (module.module_num, module.module_name))
+                fmt = "[%4.d] [%s] (disabled)"
+            lines.append(fmt % (module.module_num, module.module_name))
             for note in module.notes:
-                fd.write("%s%s\n" %("".join([" "] * indent), note))
-            fd.write("\n")
+                lines.append("%s%s" %("".join([" "] * indent), note))
+            lines.append("")
+        fd.write("\n".join(lines))
         
     def save_measurements(self, filename, measurements):
         """Save the measurements and the pipeline settings in a Matlab file
--- a/cellprofiler/pipeline.py
+++ b/cellprofiler/pipeline.py
@@ -1287,6 +1287,23 @@ class Pipeline(object):
             write_file_list(fd, self.__file_list)
         if needs_close:
             fd.close()
+    
+    def save_pipeline_notes(self, fd, indent = 2):
+        '''Save pipeline notes to a text file
+        
+        fd - file descriptor of the file.
+        
+        indent - indent of the notes relative to module header.
+        '''
+        for module in self.modules(exclude_disabled=False):
+            if module.enabled:
+                fmt = "[%4.d] [%s]\n"
+            else:
+                fmt = "[%4.d] [%s] (disabled)\n"
+            fd.write(fmt % (module.module_num, module.module_name))
+            for note in module.notes:
+                fd.write("%s%s\n" %("".join([" "] * indent), note))
+            fd.write("\n")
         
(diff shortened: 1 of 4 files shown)


cellprofiler/pipeline.py
```diff
@@ -1287,6 +1287,23 @@ def savetxt(self, fd_or_filename,
             write_file_list(fd, self.__file_list)
         if needs_close:
             fd.close()
+    
+    def save_pipeline_notes(self, fd, indent = 2):
```
0x00B1: Why haven’t you synchronized and closed the file? Does this happen elsewhere? If so, why?

cellprofiler/pipeline.py
```diff
@@ -1287,6 +1287,23 @@ def savetxt(self, fd_or_filename,
...
+        fd - file descriptor of the file.
+        
+        indent - indent of the notes relative to module header.
+        '''
+        for module in self.modules(exclude_disabled=False):
+            if module.enabled:
+                fmt = "[%4.d] [%s]\n"
+            else:
+                fmt = "[%4.d] [%s] (disabled)\n"
+            fd.write(fmt % (module.module_num, module.module_name))
```
0x00B1: I’d consider creating a string before writing to save expensive write calls.

cellprofiler/tests/test_pipeline.py
```diff
@@ -1308,6 +1308,28 @@ def test_07_01_metadata(self):
...
+        module = ATestModule()
+        module.module_num = 1
+        module.notes.append("Hello")
+        module.notes.append("World")
+        pipeline.add_module(module)
+        module = ATestModule()
+        module.module_num = 2
+        module.enabled = False
+        pipeline.add_module(module)
+        expected = """[   1] [ATestModule]
```
0x00B1: Have you considered making this a fixture (i.e. a plaintext file that’s read by the test)?

It might be cleaner and prevent anyone from creating an odd spacing bug.

cellprofiler/pipeline.py
LeeKamentsky: I've often written one of these intending to just write to a file, then later found that I had to embed the output as a section in another file. It's more flexible to leave the decision about whether to close to the caller - for instance, you might use this as part of a web page. And you can see the clumsiness of it in line 1289 - unfortunate logic which resulted from savetxt being called from different places with different requirements for the file's handling.

It also makes testing easier if you can write to StringIO.

0x00B1: How do you prevent Python’s garbage collector from closing the file?

0x00B1: Have you considering using a with block?

```Python
with open(filename, 'w') …
```

cellprofiler/tests/test_pipeline.py
LeeKamentsky: Thx - the test case is done as an array of strings, one string per line to eliminate this possibility.

cellprofiler/pipeline.py
LeeKamentsky: https://waymoot.org/home/python_string/ method number 6.

cellprofiler/pipeline.py
```diff
@@ -1287,6 +1287,25 @@ def savetxt(self, fd_or_filename,
...
+        for module in self.modules(exclude_disabled=False):
+            if module.enabled:
+                fmt = "[%4.d] [%s]"
+            else:
+                fmt = "[%4.d] [%s] (disabled)"
+            lines.append(fmt % (module.module_num, module.module_name))
+            for note in module.notes:
+                lines.append("%s%s" %("".join([" "] * indent), note))
+            lines.append("")
+        fd.write("\n".join(lines))
```
LeeKamentsky: The list comprehension came in as the fastest way to build lines into strings.

0x00B1:
:+1:

LeeKamentsky:
@0x00B1 K feel free to push the green button.

PR #1493, opened 2015-08-13
refactored cellprofiler.cpmath into a unique package

0x00B1:
Refactored into https://github.com/cellprofiler/centrosome


0x00B1:
@LeeKamentsky This is ready for review. :fireworks: :

LeeKamentsky:
Centrosome looks good - nice to have it prebuilt. Changes to files look like a straightforward redirect of the imports. The tutorials probably break as do any 3rd party plugins using cellprofiler.cpmath. Does it make sense to have some sort of legacy support for `import cellprofiler.cpmath.cpmorphology` and similar (with a deprecation warning)?

0x00B1:
> Does it make sense to have some sort of legacy support for import cellprofiler.cpmath.cpmorphology and similar (with a deprecation warning)?

Great idea. I’ll do this.

LeeKamentsky:
@0x00B1  look at https://github.com/CellProfiler/CellProfiler/tree/leek/1490 and 73000c3. The idea is to use __name__ and the fact that the module is in the sys.module dictionary from the start to find the module and copy stuff from the similarly-named module in centrosome. It works is all I'll say for it.

Merge my changes if you want and after that, fine with me to merge to master if it is with you.

0x00B1:
I need to fix centrosome’s packaging. Go figure. :laughing:

0x00B1:
@LeeKamentsky What do you think?

Issue #1508, opened 2015-08-19
wx ValueError


braymp:
Forgot to mention after #1499: Starting CP from the cmd line after 5b3567c4ed94b84f9fd4150528eda362bc13c06b also produces the following error (just the salient part here):
```
Bioformats version does not support directory cacheing
Uncaught exception in CellProfiler.py
Traceback (most recent call last):
  File "CellProfiler.py", line 253, in main
    pipeline_path = pipeline_path)
  File "c:\Trunk\CellProfiler\cellprofiler\cellprofilerapp.py", line 37, in __init__
    super(CellProfilerApp, self).__init__(*args, **kwargs)
  File "C:\Python27\lib\site-packages\wx-2.8-msw-unicode\wx\_core.py", line 7981, in __init__
    self._BootstrapApp()
  File "C:\Python27\lib\site-packages\wx-2.8-msw-unicode\wx\_core.py", line 7555, in _BootstrapApp
    return _core_.PyApp__BootstrapApp(*args, **kwargs)
  File "c:\Trunk\CellProfiler\cellprofiler\cellprofilerapp.py", line 49, in OnInit
    CellProfilerSplash.GetHeight(), 255, 255, 255, 255)
  File "C:\Python27\lib\site-packages\wx-2.8-msw-unicode\wx\_gdi.py", line 922,in EmptyBitmapRGBA
    return _gdi_._EmptyBitmapRGBA(width, height, red, green, blue, alpha)
ValueError: Width and height must be greater than zero
Failed to stop Ilastik
```
Do I need to bump up my version of wx as well?

braymp:
Might be a hint here: http://www.cellprofiler.org/forum/viewtopic.php?f=18&t=3281

braymp:
This is now the only error now remaining after f2da6c90d1c5f3aab3c2714da3650a41bb6d75c0

braymp:
Also see this: http://wxpython-users.wxwidgets.narkive.com/H7X05Fg4/aquabutton-exception-on-resize

Sounds like it was fixed in 2.9.2? And is it the case that we need 2.9+ to work with 64-bit OSX?

0x00B1:
Nice work!

It appears it’s a bug in the super useful splash screen code. LOL

braymp:
Using wx 3.0.2.0 fails with a similar error:
```
Traceback (most recent call last):
  File "CellProfiler.py", line 253, in main
    pipeline_path = pipeline_path)
[...]

PR #1510, opened 2015-08-24
fixes issue #1508. The icon was being loaded from the wrong place.

LeeKamentsky:
This isn't an ideal fix - it doesn't deal with a proper `pip install` and where the artwork goes in that case. But it restores CellProfiler to a working state.


LeeKamentsky:
Breaking rules a little - needs to be fixed to fix trunk build.

PR #1604, opened 2015-09-30
Fixes #1603 by returning a nonzero error code on pipeline failure

### LINKED ISSUE #1603: CellProfiler fails and exits with an error code of 0
CellProfiler should exit with a negative error code if it fails. Running CellProfiler headless with the -p switch, using the file, http://broadinstitute.org/~leek/bpissue6/Batch_data.h5, will recreate the problem.

LeeKamentsky:
CellProfiler 1.0 wrote a "Done" file that either had the text, "Complete" or "Failure". CellProfiler 2.x continued doing this, but it might be more appropriate to return an error code from the process (and for backwards compatibility, we do both).

BatchProfiler is currently erroneously checking the error code, so this fixes https://github.com/CellProfiler/BatchProfiler/issues/6 as well.


LeeKamentsky:
This predates CellProfiler.py -> cellprofiler/__main__.py and is so stale that its unmergeable. But the PR shows how to do it, so don't close until there is a replacement.

0x00B1:
@LeeKamentsky Cool. I’ll do this now.

PR #1610, opened 2015-10-01
Removed missing import, but it remains untested

LeeKamentsky:
@0x00B1 I removed the imports and I think it will work but somehow my Maven stuff has gone off the deep end and so I can't run the run_imagej tests.

If you check out this branch and it works, I'm glad I could help, but no guarantees.


LeeKamentsky:
It works... try merging.

PR #1622, opened 2015-10-07
Issues/1616 create synthetic images for ExampleImages tests

LeeKamentsky:


cellprofiler/modules/loadsingleimage.py
```diff
@@ -402,11 +402,10 @@ def run(self, workspace):
                 image_name, pathname, filename, rescale)
             image = provider.provide_image(image_set)
             pixel_data = image.pixel_data
-            digest = hashlib.md5()
             if wants_images:
-                digest.update(np.ascontiguousarray(pixel_data).data)
+                md5 = provider.get_md5_hash(m)
                 m.add_image_measurement("_".join((C_MD5_DIGEST, image_name)), 
-                                        digest.hexdigest())
+                                        md5)
```
LeeKamentsky: I discovered that loadsingleimage isn't calculating the MD5 hash of the file, but using the older MD5 hash of the image within the file. This change makes things uniform with the other image loading modules

LoadSingleImage is pretty heavily deprecated, so not a big deal.

PR #1556, opened 2015-09-11
Issue #1550 - new morphological code to distinguish lines from dots

tischi:
Hi Lee,
this adds code to add a morphological filter that can separate linear structures from dot-like structures.
Best, Tischi

PR #1623, opened 2015-10-07
morphological operation to distinguish lines from dots (fixes #1556)

0x00B1:


0x00B1:
@LeeKamentsky

0x00B1:
requires https://github.com/CellProfiler/CellProfiler/pull/1493

LeeKamentsky:
How do the tags to centrosome work? Do we need to tag centrosome and update the tags in requirements.txt?

0x00B1:
@LeeKamentsky I updated requirements.txt to use `centrosome==1.0.2`.

PR #1629, opened 2015-10-08
PyInstaller (WIP)

0x00B1:


0x00B1:
This was prioritized because it's blocking the 2.2.0 release.

0x00B1:
Mac support is close. It's very close. However, there're some familiar and unfamiliar remaining issues:

- cellh5 and cellh5's dependencies (lxml and pandas) aren't packaged;
- vigra isn't packaged;
- and javabridge's JARS aren't packaged.

0x00B1:
lxml packaging is straightforward.

While pandas packaging is tricky, I've found some functional examples online.

cellh5 is a mess. It requires an editable install from their GitHub repository and their setup.py doesn't properly require lxml or or pandas. I'm concerned about the project's future since they have been non-communicative about issues and pull requests.

0x00B1:
I can't install Vigra on OS X Yosemite with Homebrew.

LeeKamentsky:
I think it's OK to leave Vigra / Ilastik / CellH5 out of this pull request and once things are working, start another branch to add them to the Mac build. The 32-bit build did not support Ilastik, so it's not a regression to leave it out. CellH5 is relatively new and I am guessing there may not be any users yet.
@sommerc what software depends on the CellH5 package right now? Would you be open to a little reorganization of it? I think it might be easier to package if the cellh5 and cellh5write namespaces were combined and if the directory structure were changed so that `import cellh5` loaded a cellh5/__init__.py file which imported functions, classes and variables from cellh5/cellh5.py and cellh5/cellh5write.py. That would avoid having to place cellh5.py and cellh5write.py in the site-packages root directory and be a bit more standard.

sommerc:
> @sommerc <https://github.com/sommerc> what software depends on the 
> CellH5 package right now? Would you be open to a little reorganization 
> of it? I think it might be easier to package if the cellh5 and 
> cellh5write namespaces were combined and if the directory structure 
> were changed so that |import cellh5| loaded a cellh5/*init*.py file 
> which imported functions, classes and variables from cellh5/cellh5.py 
> and cellh5/cellh5write.py. That would avoid having to place cellh5.py 
> and cellh5write.py in the site-packages root directory and be a bit 
> more standard.
>
I agree, this would make things easier. I hope I can have a look on this 
this week and will collect other changes to make a cellh5 1.3

sommerc:
I update the cellh5 structure as you indicated, please have a look at the updated master. It still should be backwards compatible giving a deprecation warning

LeeKamentsky:
@0x00B1 I tried running `pyinstaller CellProfiler.spec` on Windows on the issues/1625 branch. It generated CellProfilerApp.exe. I ran the executable and it returned immediately without showing the UI. Not sure how to proceed - the docs for pyinstaller are a bit sketchy and I'm not sure why nothing happened when I tried running the executable.

PR #1637, opened 2015-10-09
Fixes #1636 + test

### LINKED ISSUE #1636: ClassifyObjects: VisibleDeprecationWarning
Running test_classifyobjects yields this error:

../home/travis/build/CellProfiler/CellProfiler/cellprofiler/modules/classifyobjects.py:586: VisibleDeprecationWarning: boolean index did not match indexed array along dimension 0; dimension is 7 but corresponding boolean dimension is 6

It turns out that there is an off-by-one error in the code that should mask out the wrong objects if there are nan values.

LeeKamentsky:
Old and unnecessary code was operating incorrectly.

### DIFF
--- a/cellprofiler/modules/classifyobjects.py
+++ b/cellprofiler/modules/classifyobjects.py
@@ -582,8 +582,6 @@ class ClassifyObjects(cpm.CPModule):
             nobjects = len(class_1)
             mapping = np.zeros(nobjects+1, int)
             mapping[1:] = np.arange(1, nobjects+1)
-            for i in range(2): 
-                mapping[np.isnan(saved_values[i])] = 0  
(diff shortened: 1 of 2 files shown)


cellprofiler/modules/classifyobjects.py
```diff
@@ -582,8 +582,6 @@ def run_two_measurements(self, workspace):
             nobjects = len(class_1)
             mapping = np.zeros(nobjects+1, int)
             mapping[1:] = np.arange(1, nobjects+1)
-            for i in range(2):
```
LeeKamentsky: Handling display of NaN values is taken care of elsewhere  (a few lines above this code) and this code has an off-by-one error that showed up when I added the test below.

PR #1645, opened 2015-10-13
prokaryote integration

0x00B1:


external_dependencies.py
```diff
@@ -1,334 +1,18 @@
...
+import os
+import urllib
 
-Copyright (c) 2003-2009 Massachusetts Institute of Technology
-Copyright (c) 2009-2015 Broad Institute
-All rights reserved.
+VERSION = "1.0.1"
 
-Please see the AUTHORS file for credits.
+prokaryote = '{}/prokaryote-{}.jar'.format(os.path.abspath("imagej/jars"), VERSION)
```
LeeKamentsky: @0x00B1 it looks like you might need to make the imagej/jars directory if it doesn't exist - mine was missing after a git clean -f -d -x

LeeKamentsky:
@0x00B1 So the big problem is a file in several of the jars: META-INF/json/org.scijava.plugin.Plugin. If you unzip imagej-plugins-commands.jar and imagej-plugins-tools.jar, you'll see both of them have this file. I think it's the list of discoverable plugins for the jar. We would need to run some tool to concatenate the files in each of the jars. I would bet there's some Maven magic to do that - it's unfortunate to have to layer magic on top of magic.

LeeKamentsky:
@0x00B1 Fixed prokaryote with the Maven magic. If you make a new version of prokaryote and update external_dependencies.py to take that version and to create the imagej/jars folder if it doesn't exist, I'll take the pull request and fix the one test that breaks (imagej.tests.test_imageplus.TestImagePlus.test_01_01_load_imageplus).

PR #1674, opened 2015-10-29
setup.py refactor

0x00B1:
@LeeKamentsky thoughts?

- it’ll permit installation with `$ pip install cellprofiler`
- we can remove mac_setup.py and windows_setup.py once we create binaries from setup.py
- external_dependencies.py is moved into setup.py
- added a progress bar for the prokaryote download
- remaining --build-and-exit code show be moved into setup.py

e.g.

```sh
$ python setup.py install
$ python setup.py test
```

```sh
$ python -m cellprofiler
```


BatchProfiler/bputilities.py
```diff
@@ -420,7 +420,7 @@ def build_cellprofiler(
     touch_job = "CellProfiler-touch-%s" % git_hash
```
LeeKamentsky: The BatchProfiler directory should be removed from the CellProfiler repo at this point (didn't I do that?). The BatchProfiler repo has all the changes. So no worries here, but BP will need to be changed to do different things to build depending on the version. I will fix when this goes in.

cpnose.py
```diff
@@ -139,7 +139,7 @@ def finalize(self, result):
             logging.root.warn("Failed to stop zmq boundary")
 
     def wantFile(self, filename):
```
LeeKamentsky: If we were to run the tests after installation, then setup.py wouldn't need to be excluded because it wouldn't be there. If CellProfiler is installed using pip, we should move the logic in cpnose into a nose plugin and specify the switch settings we need in setup.cfg.

setup.py
```diff
@@ -1,114 +1,148 @@
...
-    #call('lipo dist/CellProfiler.app/Contents/MacOS/CellProfiler -thin i386 -output dist/CellProfiler.app/Contents/MacOS/CellProfiler', shell=True)
-    call('rm dist/CellProfiler.app/Contents/Resources/lib/python2.7/cellprofiler/artwork/*.png', shell=True)
+        if not os.path.isfile(dependencies):
+            file = open(dependencies, "w")
+
+            file.write(prokaryote)
+
+            file.close()
+
+
```
LeeKamentsky: This is a better idea than cpnose. Need to add code to start the javabridge and stop the javabridge and Ilastik somewhere in here.

LeeKamentsky:
Overall, big :+1:. Maybe stylistically, `python setup.py prokaryote` should be part of the build phase, similar to how build_ext is? But that is minor.
**Testing** The reasons we currently use nose:
* it can output an XML version of the test output compatible with junit which was very helpful when integrating with CruiseControl. No longer an issue.
* nose's plugin framework is used to start and stop the javabridge.
IMHO, if setup can perform the startup and shutdown, no need for nose.
The approach is the right one, the devil will be in the details and many of these will probably be discovered along the way to completion.

BatchProfiler/bputilities.py
0x00B1: Nice! :+1: :+1: :+1:

setup.py
0x00B1: I knew you’d like this! :smile: I’ll add the missing javabridge and Ilastik components.

0x00B1:
`python setup.py install` works, but `pip install dist/cellprofiler-2.2.0.tar.gz` doesn’t.

While prokaryote is successfully retrieved, it isn’t copied into `.site-packages/imagej/jars`.

0x00B1:
@LeeKamentsky `javabridge` doesn’t have this problem, so what’s the solution? :stuck_out_tongue_closed_eyes:

0x00B1:
If you want to test this:

$ python setup.py sdist
$ pip install dist/cellprofiler-2.2.0.tar.gz

LeeKamentsky:
This is my top priority this afternoon. I'll review r.e. Javabridge. It was pretty tough convincing setuptools and PIP to put Javabridge's jars in a place where the code can find them. So first job is to look at the setup and try to remember what I had to do. And yeah, :stuck_out_tongue_closed_eyes: I feel your pain.

cellprofiler/analysis_worker.py
```diff
@@ -176,20 +186,20 @@ def main():
...
     cp_start_vm()
-    
+
     deadman_start_socket = the_zmq_context.socket(zmq.PAIR)
     deadman_start_socket.bind(DEADMAN_START_ADDR)
-    
+
     # Start the deadman switch thread.
-    start_daemon_thread(target=exit_on_stdin_close, 
+    start_daemon_thread(target=exit_on_stdin_close,
```
LeeKamentsky: :+1: note to self, name threads everywhere.

cellprofiler/gui/tests/test_cpfigure.py
```diff
@@ -1,295 +0,0 @@
-"""test_cpfigure- test cpfigure functionality
```
LeeKamentsky: :+1: these tests don't prove much and need the GUI

cellprofiler/modules/tests/test_identifyprimaryobjects.py
```diff
@@ -2279,32 +2279,32 @@ def test_09_01_small_images(self):
...
-        local_threshold,threshold = x.get_threshold(
-            cpi.Image(image), np.ones(image.shape,bool), workspace)
-        self.assertTrue(threshold > 0.040)
-        self.assertTrue(threshold < 0.050)
-        x.object_fraction.value = '0.50'
-        local_threshold,threshold = x.get_threshold(
-            cpi.Image(image), np.ones(image.shape,bool), workspace)
-        self.assertTrue(threshold > 0.0085)
-        self.assertTrue(threshold < 0.0090)
+    # def test_09_02_mog_fly(self):
```
LeeKamentsky: Fine for now as we talked about previously - would be solved if we added the fly image as a resource similarly to the test image in Centrosome (tests/resources/kirsch.tiff). But that is a lower priority than competing the setup.

setup.py
```diff
@@ -1,102 +1,188 @@
...
+setuptools.dist.Distribution({
+    "setup_requires": [
+        "clint",
+        "requests",
+        "pytest"
+    ]
+})
+
+
```
LeeKamentsky: My first guess about why `python setup.py build install` differs from `python setup.py bdist; pip install ..` is that neither the bdist command nor the pip install run the install class below. First step is to confirm that and if so, I will take a shot at making the download step happen as a subcommand of 'build' so that the jar is in place before bdist. Just a guess at this point and the change to the code may be pretty minimal.

LeeKamentsky:
@0x00B1 see https://github.com/CellProfiler/CellProfiler/pull/1719 for my fix and we can discuss if needed.

braymp:
I tried running 'python -m cellprofiler' but nothing happens; I just get the cmd prompt again after a few seconds. Am I missing something?

braymp:
Also, @LeeKamentsky, how do I run CP in Wing now?

LeeKamentsky:
If you do `python setup.py install`, it will generate "cellprofiler.exe" in the scripts directory of your Python. (if a virtualenv, <virtualenv-name>\Scripts\cellprofiler.exe, if Python, probably c:\Python27\Scripts\cellprofiler.exe). If you have the virtualenv activated for your command-line, you can just type "cellprofiler" to start it.

For Wing, my features/py2exe branch has a copy of CellProfiler.py whose only purpose is to call cellprofiler.__main__.main. So you could get that:

    git fetch origin
    git checkout features/py2exe -- CellProfiler.py

I think the above works, I didn't try it though. Wing has named entry points - I haven't got that to work though :disappointed:.

PR #1675, opened 2015-10-30
Implement composite mode in GrayToColor

LeeKamentsky:
This lets the user specify an arbitrary number of input images and colors for each. Each grayscale image is converted to its chosen color and then the set of color images are composited together by a simple addition.

### DIFF
--- a/cellprofiler/modules/graytocolor.py
+++ b/cellprofiler/modules/graytocolor.py
@@ -188,10 +188,26 @@ class GrayToColor(cpm.CPModule):
         default_color = DEFAULT_COLORS[
             len(self.stack_channels) % len(DEFAULT_COLORS)]
         group.append("image_name", cps.ImageNameSubscriber(
-            "Select the input image to add to the stacked image", cps.NONE))
+            "Image name", cps.NONE,
+            doc = 
+            """
+            <i>Used only if %(SCHEME_STACK)s or %(SCHEME_COMPOSITE)s is
+            chosen</i><br>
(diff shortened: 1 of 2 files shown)


braymp:
Here is an image from CPA which does a good job of color composition IMHO:
image

With the GrayToColor mods:
image

Pretty good. Might be good to add settings for relative weights as well, as the RGB option does.

LeeKamentsky:
If you are clever, you can just pick darker colors to adjust the relative weights (but the scale is limited because each color only has a 0-255 range). But I agree that having the relative weights is a more obvious solution and we are likely to have users who didn't figure out the work-around asking for the feature. I will mark the pull request "WIP" and add the weights unless someone objects.

LeeKamentsky:
@braymp added weights.

braymp:
:+1:

cellprofiler/modules/graytocolor.py
```diff
@@ -35,19 +35,30 @@
...
+OFF_STACK_CHANNEL_COUNT_V3 = 16
+OFF_STACK_CHANNEL_COUNT = 16
+
 SCHEME_RGB = "RGB"
 SCHEME_CMYK = "CMYK"
 SCHEME_STACK = "Stack"
+SCHEME_COMPOSITE = "Composite"
 LEAVE_THIS_BLACK = "Leave this black"
 
+DEFAULT_COLORS = [
```
0x00B1: Cute! :dancer:

PR #1679, opened 2015-11-03
fix typos for "Alcian"

dlogan:
https://en.wikipedia.org/wiki/Alcian_blue_stain

### DIFF
--- a/cellprofiler/modules/tests/test_unmixcolors.py
+++ b/cellprofiler/modules/tests/test_unmixcolors.py
@@ -84,8 +84,8 @@ UnmixColors:[module_num:1|svn_version:\'Unknown\'|variable_revision_number:2|sho
     Red absorbance\x3A:0.5
     Green absorbance\x3A:0.5
     Blue absorbance\x3A:0.5
-    Image name\x3A:AlicanBlue
-    Stain:Alican blue
+    Image name\x3A:AlcianBlue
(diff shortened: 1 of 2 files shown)


LeeKamentsky:
Unfortunately, you need to fix upgrade_settings to look for the old spelling and change it to the new one and ideally, add a test case for that as well.

dlogan:
Ah, ok.  Not that important, so I will add do this maybe on coding day.

0x00B1:
@dlogan What’s going on with this?

dlogan:
Oh, I didn't have time for this, and won't for awhile.   Admittedly, I don't think I've never modified upgrade_settings and my test case creation skills are lame. I'd be glad to have help :wink:

LeeKamentsky:
IMHO the fix is not urgently needed and either the branch or pull request can serve as documentation for what I need to do to resolve. Can we leave the pull request open and I'll get to it sooner or later?

dlogan:
Indeed!  No one else ever pointed this typo out, so it can wait.

0x00B1:
@LeeKamentsky and @dlogan bump

AnneCarpenter:
Sounds like it's not a priority to fix this under the circumstances. What is the procedure from here, so the partially functional pull request doesn't linger forever? Just "close"?

PR #1688, opened 2015-11-04
pip install cellprofiler on Windows

LeeKamentsky:
This works on Windows, including python setup.py test.

ToDo:

Build of numpy-vigra and Ilastik


LeeKamentsky:
@0x00b1 OK - don't get too excited about the green light. I disabled test_analysis.py and test_analysisworker.py. If you're running headless on a cluster, no big deal, but I have to debug why the builds mysteriously fail if you run these tests (may be segfault?)

0x00B1:
Should we drop the rpdb test altogether?

0x00B1:
Have you isolated the specific test_analysis and test_analysisworkers tests that are hanging? If not, I’ll try.

LeeKamentsky:
Regarding RPDB, it never feels great to take out a feature, but I think nobody's ever used it, so the test and the functionality should be removed - alternately, the hacked version of rpdb should be put back into the code base. Our version is derived from someone else's but ours has some handshaking between the GUI and the analysis worker that happens before the analysis worker descends into the debugger - it wouldn't be worthwhile to try to adapt existing packages in order to recover a facility that no-one uses.

LeeKamentsky:
It's been over a week - I forget exactly which test causes the problem. If I remember right, it's a segfault and if I remember right, I've seen this problem in other Unix builds. It might be a bug that's not present in a prior version of ZMQ, LIBHDF5 or H5Py.

If you look at this build: https://travis-ci.org/CellProfiler/CellProfiler/builds/89505966, you'll see at the end that the tests exit with a code of 255 (all the way at the bottom) without printing any information on which tests passed or failed. I interpreted that as a segfault that happened during a test, preventing the test framework from printing any information about test success or failure.

If you want to try enabling the tests and figuring out what's causing the segfault (or if there is one) in the travis environment, that would be great. Even if you can get a crash dump or a stack trace, that's a win and I might be able to diagnose things from there.

0x00B1:
I appreciate the information. I’m looking into this now.

0x00B1:
Would you mind squashing and sending a pull request to merge into #1483?

0x00B1:
Actually, would you mind if I did this? It’s blocking some of my setup.py work. :smile:

LeeKamentsky:
I can do the squash - probably best since then it will have my name on it.

0x00B1:
:+1:

LeeKamentsky:
@0x00B1 I squashed. Saved the branch's old state on my fork in case I flubbed it.

0x00B1:
LGTM! Thanks! :dancers: 

I'll merge into #1483 when Travis passes

PR #1709, opened 2015-11-20
removed CellProfiler_Developer_Tutorial.pptx

0x00B1:
We shouldn’t keep non-essential binaries in the repository.


AnneCarpenter:
Where should we keep it (assuming it’s still useful)? We do have a place to store Tutorials on imaging/analysis but it’s not public.

0x00B1:
Moved to S3.

(@LeeKamentsky bump)

Issue #1626, opened 2015-10-08
Errors in test suite


LeeKamentsky:
There's a class of errors showing up in the test suite on Travis:
https://travis-ci.org/CellProfiler/CellProfiler/builds/84153919#L5143

I think these are related to the numpy version. Code like:

    outlines |= centrosome.outline.outline(l)

should be

    outlines |= centrosome.outline.outline(l).astype(outlines.dtype)

to provide the explicit cast that was probably done implicitly in previous versions.

PR #1712, opened 2015-11-20
reformatting (WIP)

0x00B1:
- PEP 0008 compliance


LeeKamentsky:
There's a situation with numpy 1.9.1+ regarding augmented assignment (see #1626), so we should examine each of the changes.

0x00B1:
@LeeKamentsky What’s NumPy’s issue with augmented assignment? I successfully ran an example pipeline for each change. In addition to running Travis, what else would you recommend?

LeeKamentsky:
If you look here: https://travis-ci.org/CellProfiler/CellProfiler/builds/84153919#L5143 a test failed. Numpy 1.10 changed the rules for casting:
https://github.com/numpy/numpy/blob/97c35365beda55c6dead8c50df785eb857f843f0/doc/source/reference/ufuncs.rst

<quote>Provides a policy for what kind of casting is permitted. For compatibility with previous versions of NumPy, this defaults to 'unsafe' for numpy < 1.7. In numpy 1.7 a transition to 'same_kind' was begun where ufuncs produce a DeprecationWarning for calls which are allowed under the 'unsafe' rules, but not under the 'same_kind' rules. In numpy 1.10 the default will be 'same_kind</quote>

This caused some instances of +=, which I used infrequently, to break in Travis. The rules for a = a+b are a little more lenient. We may see some errors in the tests, but there may not be enough coverage and cases where people are adding boolean masks to floats or similar might cause breakage.

PR #1719, opened 2015-11-30
Create imagej/jars dependencies in build_lib

LeeKamentsky:
Moved artwork from data to package_data.

Took me 3 hours to figure out how to write 5 lines of code. The big change is to build the imagej/jar directory in the build_lib directory which if you are doing `python setup.py install` is build\lib and if you are doing `pip install ...` is relative to a temp directory. I think what was happening is that the dependencies were previously being written to their target directories and ignored.

I also moved the artwork so that, for Windows, it installs to the lib/site-packages/artwork directory rather than the artwork directory. There is a slight problem in that "artwork" isn't technically a package because it's missing an __init__.py file. The script gives you a warning, but it does create the directory and package.

On Windows, the setup creates a cellprofiler.exe file in the scripts directory and you can simply type "cellprofiler" after activating the virtual environment. First class!!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ class Install(setuptools.command.install.install):
 
         version = "1.0.3"
 
-        directory = os.path.join("imagej", "jars")
+        directory = os.path.join(self.build_lib, "imagej", "jars")
 
         if not os.path.exists(directory):
             os.makedirs(directory)
@@ -48,7 +48,9 @@ class Install(setuptools.command.install.install):
 
                         f.flush()
 
-        dependencies = os.path.abspath("imagej/jars/cellprofiler-java-dependencies-classpath.txt")
+        dependencies = os.path.abspath(os.path.join(
+            self.build_lib, 'imagej', 'jars', 
+            'cellprofiler-java-dependencies-classpath.txt'))
 
         if not os.path.isfile(dependencies):
             dependency = open(dependencies, "w")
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -25,7 +25,7 @@ def run(self):
 
         version = "1.0.3"
 
-        directory = os.path.join("imagej", "jars")
+        directory = os.path.join(self.build_lib, "imagej", "jars")
```
0x00B1: Nice

setup.py
```diff
@@ -134,16 +136,10 @@ def runtest(self):
...
-        (
-            os.path.join("artwork"),
-            glob.glob(os.path.join("artwork", '*'))
-        ),
-        (
-            os.path.join("imagej", "jars"),
-            glob.glob(os.path.join("imagej", "jars", "*.jar"))
-        )
-    ],
+    package_data={
```
0x00B1: I misread the [Packaging and Distributing Projects](https://packaging.python.org/en/latest/distributing/) documentation.

Woof

:cry:

0x00B1: Would you mind alphabetizing the keys (i.e. “package_data” and “include_package_data”) and using double quotes rather than single quotes?

setup.py
```diff
@@ -48,7 +48,9 @@ def run(self):
 
                         f.flush()
 
-        dependencies = os.path.abspath("imagej/jars/cellprofiler-java-dependencies-classpath.txt")
+        dependencies = os.path.abspath(os.path.join(
```
0x00B1: Was this part of the issue?

setup.py
LeeKamentsky: np - get right on it.

setup.py
LeeKamentsky: I needed to join self.build_lib to "imagej/jars/cellprofiler-java-dependencies-classpath.txt", so I broke it up into separate directory names. But os.join(self.build_lib, 'imagej...') probably works too.

LeeKamentsky:
Alphabetized setup dictionary + double quotes.

PR #1721, opened 2015-12-02
Py2exe command for windows binary

LeeKamentsky:
Questions / To Dos:
* Is it OK to have lots of files? It simplifies logic to operate differently if frozen.
* The distribution requires the MSVCRT redistributable. py2exe recommends this: http://www.py2exe.org/index.cgi/Tutorial#Step5 but I bundled the redistrutable in the MSI and did a separate install step previously. Advantages of their approach - the DLLs are local and you don't get DLL hell from whatever is installed (and don't create DLL hell either).
* MSI bundling. We used InnoSetup previously. It works, it does a nice job with file associations, uninstall, icons and other niceties. It's freeware. But that was 5 years ago and there may be an open-source solution.
* It's fat - weighs in at 540 mb. Meh, the old one, once installed was 530 mb.
* Java - the old install came with a local version of Java. Do we revisit that decision? (see http://stackoverflow.com/questions/6003816/how-should-i-bundle-java-with-my-windows-application)


braymp:
I've seen at least one case during a workshop (where users brought their own machines) in which Java wasn't installed. I'd like to think all computers have it but we can't guarantee it. An alternate is to inform the user if they don't have it and prompt them to get it.

CellProfiler.py
```diff
@@ -0,0 +1,3 @@
+from cellprofiler import __main__
+
+__main__.main()
```
0x00B1: We should use:

```Python
if __name__ == "__main__":
    cellprofiler.__main__.main()
```

per [docs.python.org](https://docs.python.org/2/library/__main__.html).

CellProfiler.py
```diff
@@ -0,0 +1,3 @@
+from cellprofiler import __main__
```
0x00B1: We should use:

```Python
import cellprofiler.__main__
```

per [docs.python-guide.org](http://docs.python-guide.org/en/latest/writing/structure/#modules).

0x00B1:
Awesome! :smile:

setup.py
```diff
@@ -1,19 +1,57 @@
...
 import glob
 import os
 import setuptools
 import setuptools.command.build_ext
+import setuptools.command.install
 import setuptools.dist
 import sys
-import setuptools.command.install
 
-setuptools.dist.Distribution({
```
0x00B1: Why was this removed?

setup.py
```diff
@@ -1,19 +1,57 @@
+import distutils
 import glob
 import os
 import setuptools
 import setuptools.command.build_ext
+import setuptools.command.install
 import setuptools.dist
 import sys
-import setuptools.command.install
```
0x00B1: Why was this removed?

setup.py
```diff
@@ -112,7 +150,39 @@ def runtest(self):
         cellprofiler.__main__.stop_cellprofiler()
 
         sys.exit(errno)
-
+        
+class CPPy2Exe(py2exe.build_exe.py2exe):
```
0x00B1: Can this be achieved with the [data_files](http://www.py2exe.org/index.cgi/data_files) option?

setup.py
```diff
@@ -1,19 +1,57 @@
...
-import setuptools.command.install
 
-setuptools.dist.Distribution({
-    "setup_requires": [
-        "clint",
-        "requests",
-        "pytest"
-    ]
-})
+try:
```
0x00B1: We should move these imports into the “setup_requires” dictionary.

setup.py
```diff
@@ -164,16 +248,42 @@ def runtest(self):
     license="BSD",
     long_description="",
     name="cellprofiler",
+    options = {
+        "py2exe": {
```
0x00B1: We should move py2exe options into a `[py2exe]` block in our setup.cfg and keep setup.py multi-platform.

setup.py
```diff
@@ -164,16 +248,42 @@ def runtest(self):
...
+                "lxml", "lxml.*",
+                "scipy.io.matlab.streams", "scipy.special", "scipy.special.*",
+                "scipy.sparse.csgraph._validation",
+                "skimage.draw", "skimage._shared.geometry", 
+                "skimage.filters.rank.*",
+                "sklearn.*", "sklearn.neighbors", "sklearn.neighbors.*",
+                "sklearn.utils.*", "sklearn.utils.sparsetools.*",
+                "zmq", "zmq.utils", "zmq.utils.*", "zmq.utils.strtypes"
+                ],
+            "packages": packages,
```
0x00B1: Shouldn’t we include “tutorial?”

setup.py
```diff
@@ -164,16 +248,42 @@ def runtest(self):
...
+                "nsi.dll",
+                "winnsi.dll"
+                ],
+            "excludes": [
+                "Cython",
+                "IPython",
+                "pylab",
+                "Tkinter"
+                ],
+            "includes": [
```
0x00B1: You’re a hero! :muscle: This saved me so much stress.

setup.py
LeeKamentsky: It looked like dead code to me (could have been wrong). I think the code makes an instance of Distribution and then doesn't assign it to a variable, so the instance is garbage collected?

setup.py
LeeKamentsky: :+1:

CellProfiler.py
LeeKamentsky: :+1:

setup.py
LeeKamentsky: Unfortunately, I think it can't be done with data_files in the setup dictionary. The `install` command would see the matplotlib data files, for instance, and install them (and I think it would install them in the root directory of the python installation rather than in site-packages. The install command is rerun as part of py2exe.run and it puts the data files in the dist directory. If the install command is run alone, the data files are not installed. If you run `python setup.py py2exe install`, it will probably do the wrong thing and process data files intended only for py2exe - technically, I should probably remove the data files I added to the distribution.

It might be worth trying package_data (and that would unify py2app and py2exe) - you may be able to specify package data for packages you aren't installing?

setup.py
LeeKamentsky: Good question. I think it's expedient to leave tutorial out (what if there's a Python package named "tutorial"?). IMHO there's not enough value at the moment to including it.

setup.py
LeeKamentsky: I started to do this, then ran into the problem that the packages dictionary entry is computed rather than hardcoded. I did try adding an options section for "foo" to see if setup.py would run with options to a non-existent command and it did, so it's not a blocker and we can talk about hardcoding the packages list.

LeeKamentsky:
@0x00B1 I think this is in good enough shape to try. Some commands:

    python setup.py install
    python setup.py develop install
    python setup.py py2exe # creates a runnable cellprofiler.exe in the dist directory
    python setup.py py2exe msi # creates the windows installer if you have inno setup

0x00B1:
This worked! I’m stoked. :grinning:

A few comments about the build process:

**`python setup.py install`**

It failed because clint wasn’t installed. It ran after I installed clint (`pip install clint`). I believe I broke this in an earlier patch (i.e. I suspect this isn’t a new regression).

**`pip install --e .`**

`pip install --e . failed` during lxml installation. It ran after I installed lxml using a wheel provided by Christoph Gohlke.

**`python setup.py py2exe`**

`python setup.py install` doesn’t install py2exe. While I don’t think it should, the py2exe installation was somewhat awkward since their PyPI distribution is for Python 3 rather than Python 2. I installed py2exe from source:

pip install http://sourceforge.net/projects/py2exe/files/latest/download?source=files 

We should document this somewhere.

0x00B1:
@braymp you should follow this thread until it’s merged.

0x00B1:
@LeeKamentsky I should note that I didn’t start with a fresh Windows 10 installation.

I installed Python 2.7.10 from the python.org distribution.

Here’s the post-installation output of `pip freeze`:

```sh
$ pip freeze
args==0.1.0
cellh5==1.3.0
-e git+https://github.com/CellProfiler/CellProfiler.git@747145186884353bbafc77af4936230d56d8f2b8#egg=CellProfiler-features_py2exe
centrosome==1.0.3
clint==0.5.1
colorama==0.3.3
Cython==0.23.4
decorator==4.0.4
h5py==2.5.0
hmmlearn==0.1.1
ilastik==0.5.12
javabridge==1.0.11
libtiff==0.4.0
lxml==3.4.4
matplotlib==1.4.3
MySQL-python==1.2.5
networkx==1.10
numpy==1.10.1
pandas==0.17.0
Pillow==3.0.0
py==1.4.31
py2exe==0.6.9
PyOpenGL==3.1.1b1
PyOpenGL-accelerate==3.1.1a1
pyparsing==2.0.6
PyQt4==4.11.4
pytest==2.8.4
python-bioformats==1.0.5
python-dateutil==2.4.2
pytz==2015.7
pyzmq==14.7.0
qimage2ndarray==1.3.1
scikit-image==0.11.3
scikit-learn==0.16.1
scipy==0.16.1
six==1.10.0
vigranumpy==1.10.0
virtualenv==13.1.2
wheel==0.26.0
wxPython==3.0.2.0
wxPython-common==3.0.2.0
```

0x00B1:
And here is the software I had installed:

```sh
Visual C++ MSBuild ARM Package                                                                    14.0.23107
Microsoft Build Tools 14.0 (amd64)                                                                14.0.23107
Microsoft Visual C++ 2008 Redistributable - x86 9.0.30729                                         9.0.30729
Microsoft Visual Studio Team Foundation Server 2015 Office Integration Language Pack (x64) - ENU  14.0.23102
Roslyn Language Services - x86                                                                    14.0.23107
Visual C++ Library PGO X86 Package                                                                14.0.23026
Visual Studio 2015 Prerequisites                                                                  14.0.23107
Microsoft Visual C++ 2015 x86 Additional Runtime - 14.0.23026                                     14.0.23026
Microsoft Visual Studio 2015 Diagnostic Tools - x86                                               14.0.23107
Visual C++ Compiler/Tools X86 Base Resource Package                                               14.0.23026
Windows Espc Package                                                                              14.0.23107
Microsoft Portable Library Multi-Targeting Pack                                                   14.0.23107.00
Microsoft Visual C++ 2008 Redistributable - x86 9.0.30729.4148                                    9.0.30729.4148
Visual C++ IDE Debugger Resource Package                                                          14.0.23107
Microsoft .NET Framework 4.5.1 SDK                                                                4.5.51641
Visual C++ Compiler/Tools X86 Base Resource Package                                               14.0.23026
Visual C++ Compiler/Tools X86 Base Package                                                        14.0.23026
WCF Data Services 5.6.4 Runtime                                                                   5.6.62175.4
PreEmptive Analytics Visual Studio Components                                                     1.2.5134.1
Visual C++ IDE Base Package                                                                       14.0.23107
Windows Software Development Kit DirectX x64 Remote                                               8.100.25984
Visual C++ IDE Common Resource Package                                                            14.0.23107
VMware Tools                                                                                      10.0.1.3160059
WCF Data Services Tools for Microsoft Visual Studio 2015                                          5.6.62175.4
Team Explorer for Microsoft Visual Studio 2015                                                    14.0.23102
Microsoft Blend for Visual Studio 2015                                                            14.0.23107
Microsoft Azure Shared Components for Visual Studio 2015 - v1.5                                   1.5.30619.1602
Microsoft Visual Studio 2015 Test Tools Language Pack - ENU                                       14.0.23107
Visual C++ MSBuild Base Resource Package                                                          14.0.23107
psqlODBC_x64                                                                                      09.03.0300
Microsoft Azure Mobile Services Connected Service                                                 14.0.23107
Visual C++ MSBuild Base Package                                                                   14.0.23107
Microsoft SQL Server 2014 Management Objects                                                      12.0.2000.8
Java Auto Updater                                                                                 2.8.65.17
Microsoft Visual Studio 2015 Diagnostic Tools - amd64                                             14.0.23107
Microsoft.VisualStudio.Office365                                                                  1.0.0.0
Visual C++ IDE Debugger Package                                                                   14.0.23107
Microsoft Azure Storage Connected Service                                                         14.0.23107
Microsoft Windows SDK for Visual Studio 2008 Headers and Libraries                                6.1.5288.17011
Microsoft Agents for Visual Studio 2015 Preview - ENU                                             14.0.23102
Visual C++ IDE Core Professional Plus Resource Package                                            14.0.23107
```

CellProfiler64.iss
```diff
@@ -5,7 +5,7 @@
 ; See the accompanying file LICENSE for details.
```
0x00B1: We don’t need a copyright header.

.travis.yml
```diff
@@ -43,9 +43,11 @@ install:
...
 #
   - pip install --no-deps javabridge
+  - pip install --no-deps pyzmq==13.1.0
   - pip install --requirement requirements.txt
   - pip install --editable git+https://github.com/h5py/h5py.git#egg=h5py
   - pip install --editable git+https://github.com/CellH5/cellh5.git#egg=cellh5
   - pip freeze
 script:
-  - python setup.py develop --no-deps install test
+  - python setup.py install
```
0x00B1: Awesome!

.travis.yml
```diff
@@ -43,9 +43,11 @@ install:
 # of numpy
 #
   - pip install --no-deps javabridge
+  - pip install --no-deps pyzmq==13.1.0
```
0x00B1: What was the problem with pyzmq? And do we still need to peg to 13.1.0?

CellProfiler64.iss
```diff
@@ -25,11 +25,11 @@ AppSupportURL=http://www.cellprofiler.org
 AppUpdatesURL=http://www.cellprofiler.org
 DefaultDirName={pf64}\CellProfiler
```
0x00B1: Is this `C:\Program Files\CellProfiler`?

CellProfiler64.iss
```diff
@@ -45,7 +45,6 @@ Name: "{userappdata}\CellProfiler\ijplugins"; Flags: uninsneveruninstall
 Source: ".\dist\CellProfiler.exe"; DestDir: "{app}"; Flags: ignoreversion
 Source: ".\dist\analysis_worker.exe"; DestDir: "{app}"; Flags: ignoreversion
```
0x00B1: Is this still needed? Couldn’t we mimic the Linux and OS X behavior and run `python CellProfiler.py`?

CellProfiler64.iss
0x00B1: (I apologize for the dumb questions. I’m far from a Windows expert.)

braymp: On Wed, Dec 9, 2015 at 1:40 PM, Allen Goodman <notifications@github.com>
wrote:

> > DefaultDirName={pf64}\CellProfiler
>
> Is this C:\Program Files\CellProfiler?
>
According to this (http://www.jrsoftware.org/ishelp/index.php?topic=consts),
yes.

CellProfiler64.iss
```diff
@@ -45,7 +45,6 @@ Name: "{userappdata}\CellProfiler\ijplugins"; Flags: uninsneveruninstall
...
 Source: ".\dist\*"; DestDir: "{app}"; Flags: ignoreversion recursesubdirs createallsubdirs
-Source: ".\windows\vcredist_x64.exe"; DestDir: "{tmp}"
```
0x00B1: Is the C++ redistributable included elsewhere?

.travis.yml
LeeKamentsky: The tests on travis were breaking so I tried this change in desperation. It worked.

CellProfiler64.iss
```diff
@@ -101,6 +99,7 @@ End;
 function InitializeSetup(): Boolean;
 Var
   Message: String;
+  JavaVer: String;
```
0x00B1: We’re including Java? If so, are including the JDK rather than the JRE?

CellProfiler64.iss
```diff
@@ -114,7 +113,20 @@ Message := 'This build can only run on a 64-bit operating system, but yours is 3
       Result := False;
       End
     else
-      Result := True;
+//
```
0x00B1: It’s probably worthwhile to breakout this into a separate function.

.travis.yml
LeeKamentsky: I think this was why the build was breaking on Travis:
https://travis-ci.org/CellProfiler/CellProfiler/builds/95658209

It failed in test_analysis which is using ZMQ, so I figured I'd revert. It worked, I do not know what I fixed. ZMQ 14+ works on windows, though.

CellProfiler64.iss
```diff
@@ -114,7 +113,20 @@ Message := 'This build can only run on a 64-bit operating system, but yours is 3
...
+// Check taken from the following stackoverflow post:
+//
+// http://stackoverflow.com/questions/1297773/check-java-is-present-before-installing
+//
+      RegQueryStringValue(HKLM, 'SOFTWARE\JavaSoft\Java Runtime Environment', 'CurrentVersion', JavaVer);
+      if Length( JavaVer ) > 0 then Begin
```
0x00B1: We should check for the appropriate version too.

0x00B1: (I say that because too many universities use unexpected versions of Java.)

CellProfiler64.iss
LeeKamentsky: Yup - it's innosetup's variable for the Windows x64 program directory.

CellProfiler64.iss
LeeKamentsky: Yes we could - it's a little bit of legacy at this point. I think it makes sense to standardize, but we have to take out some code in cellprofiler.analysis that expects analysis_worker.exe to be there in the frozen version of CellProfiler.

cellprofiler/utilities/cpjvm.py
```diff
@@ -67,7 +67,7 @@ def get_patcher_args(class_path):
 
     patchers = filter(
         (lambda x:os.path.split(x)[1].startswith("prokaryote")), class_path)
-    if len(patchers) > 0 and False:
```
0x00B1: What did `and False` check? It would’ve failed regardless.

Was this intended behavior to circumvent some other issue?

CellProfiler64.iss
LeeKamentsky: I changed the way we do the redistributable according to directions here:
http://www.py2exe.org/index.cgi/Tutorial#Step52

There's a step (https://github.com/CellProfiler/CellProfiler/blob/features/py2exe/setup.py#L236) in the custom py2exe command that goes looking for MSVC 9.0 (or MSVC 9.0 express) in the Windows registry and then finds the redistributable files there. This has fewer chances of conflict with the system copy of the .dll files and it means we don't have to unpack vcredist_x64.exe and run it on the user's machine with some remote possibility that we break some application on that machine.

CellProfiler64.iss
LeeKamentsky: Nope - we check to make sure Java is installed and if not, we tell the user to go get it.

cellprofiler/utilities/cpjvm.py
LeeKamentsky: This was in the code to prevent the patcher from running until Prokaryote-1.0.4 was released with the necessary javaagent stuff. I moved the version # to setup.cfg at about the same time as you released 1.0.4 and I figured that as long as I was moving the version, I'd update the version # and allow the patcher to run.

cellprofiler/utilities/version.py
```diff
@@ -92,10 +92,23 @@ def get_version():
         import cellprofiler.frozen_version
         return cellprofiler.frozen_version.version_string
 
+def get_dotted_version():
```
0x00B1: I don’t know if this is the appropriate place to discuss, but we should consider rethinking our existing versioning process. A few observations:

-	a version number (e.g. 2.2.0) determines CellProfiler’s most recent release
-	a version doesn’t describe the release channel (e.g. nightly, beta, or stable)
-	a version (i.e. frozen_version) includes the object information generated by Git (e.g. a commit’s SHA-1)

A few opinions about versioning from a debugging perspective:

-	a version number is useful for determining when issues arise or resolve
-	a version channel is useful for narrowing when an issues arise or resolve
-	object information is useful for pinpointing when an issue arose or resolved

However, object information is only useful if the debugger is capable of running arbitrary versions, channel distributions, and commits. While it’s certainly possible, it’s difficult and usually impractical. Moreover, I think testing every commit on Travis CI mostly addresses this.

I don’t know. On one hand, a SHA-1 is undoubtedly useful when debugging. On the other hand, it requires a lot of baggage.

CellProfiler64.iss
LeeKamentsky: OK. Version is 1.6+ at this point. There's some code at the link in the comments that we can cut and paste.

cellprofiler/utilities/version.py
```diff
@@ -92,10 +92,23 @@ def get_version():
...
+    if not hasattr(sys, 'frozen'):
+        try:
+            return subprocess.check_output(["git", "describe", "--tags"])\
```
0x00B1: It’s wholly functional, but it’s fragile. It assumes we’re using: 

-	Git for version control; and
-	Git tags for releases. 

While I wouldn’t expect this to change soon,  it’s a notable constraint nonetheless.

setup.cfg
```diff
@@ -37,6 +40,143 @@ packages =
...
+    cellprofiler.utilities,
+    centrosome,
+    contrib,
+    h5py,
+    imagej,
+    libtiff,
+    lxml
+skip_archive = 1
+
+[py2exe_ilastik]
```
0x00B1: We should wait to add this until we can successfully package ilastik.

0x00B1: And when that happens (I’m an optimist, LOL), is there any reason we’d want to distribute CellProfiler without Ilastik on Windows?

cellprofiler/utilities/version.py
LeeKamentsky: My workflow often is:

* Receive bug report with git hash
* Check out CellProfiler detached head at that git hash
* Confirm the bug at the git hash
  * if a bug, then check out master
    * if not a bug at master, tell the user that it's fixed
    * if a bug, fix

Mostly, people were using the trunk build as the distribution channel, no tagging, so I needed the git hash. We're releasing frequently enough that almost always, a bug report with a version # will be as good as a git hash.

The other cool thing about the SHA-1 has to do with reproducible research. If someone puts a CellProfiler pipeline in their paper or distributes the experiment measurements, the SHA-1 names the version and that could be a fork or a branch of the repo. To me, it has more of a real flavor: the thing itself rather than the name we gave to the thing. And, in BatchProfiler, we do stuff like pull out the git hash from the pipeline and run CellProfiler on that version.

setup.py
```diff
@@ -1,42 +1,151 @@
...
 import setuptools
 import setuptools.command.build_ext
+import setuptools.command.develop
+import setuptools.command.install
 import setuptools.dist
 import sys
 import setuptools.command.install
 import cellprofiler.utilities.version
 
+# Recipe needed to get real distutils if virtualenv.
```
0x00B1: Is this needed? I don’t think we’re using distutils anywhere anymore. If we are, we should update to setuptools since distutils was deprecated a few years ago.

cellprofiler/utilities/version.py
LeeKamentsky: It was a bit of a stopgap so that the title bar would say 2.2.0. I bet we can come up with some clever way of putting the version in one place. IMHO:

     python setup.py build_version

should do that and this code should go away.

setup.cfg
LeeKamentsky: I think moving `[py2exe_ilastik]` out of master is ok for now - I can work on it on a branch.

There's Ilastik the application and there's Ilastik the library that's needed for classify_pixels to run. In an ideal world, you'd have "ilastik==0.5" in the "install_requires" list and then we'd modify the `[py2exe]` section to look like `[py2exe_ilastik]`. Packaging the Ilastik dependency, Vigra so it can be pip installed would be very difficult - building Vigra requires CMake and installation of system packages like libhdf5. Until we get to that, I am hoping we can have a non-standard (but reproducible) build process for Ilastik and its dependencies, a command that packages a binary executable that can run classify_pixels and a setup script that allows CellProfiler to be pip installed on a bare Python without Ilastik support.

But yes - fine to have a CellProfiler distribution without the Ilastik application as long as we can provide a version of Ilastik that can produce classifiers compatible with classify_pixels.

setup.py
LeeKamentsky: The error messages show up when you run the distribution's "cellprofiler.exe" after running `python setup.py py2exe` within a virtual environment on the build machine.

h5py.version imports distutils.version and the virtualenv hack causes the "cannot import name dist", so it's not our use of distutils that requires the code. The second error happens because py2exe packages the virtualenv's site.py which won't work without a hand-crafted orig-prefix.txt file.

0x00B1:
@LeeKamentsky I’ve been thinking a lot about this in the past 48-hour—In my experience, maintaining a platform-independent build configuration (i.e. setup.py) reduces unexpected multi-platform problems. We should consider cherry-picking CellProfiler64.iss into a separate “windows” repository. We could also include a separate setup.py with a platform-specific configuration. What do you think?

setup.py
```diff
@@ -1,42 +1,151 @@
...
+
 
 class Install(setuptools.command.install.install):
     def run(self):
+        self.run_command("build_version")
+        self.run_command("build_java_dependencies")
+        setuptools.command.install.install.run(self)
+
+
+class Develop(setuptools.command.develop.develop):
```
0x00B1: I think we’re at risk of doing too much (i.e. having too many supplementary commands). Personally, I’m skeptical of builds that are anything more than “make,” “make test,” and “make install.”

setup.py
```diff
@@ -1,42 +1,151 @@
...
+
+    def run(self):
+        with open(os.path.join("cellprofiler", "frozen_version.py"),
+                  "w") as fd:
+            fd.write("version_string='%s'\n" %
+                     cellprofiler.utilities.version.version_string)
+            fd.write("dotted_version='%s'\n" % self.version)
+
+
+class BuildJavaDependencies(setuptools.Command):
```
0x00B1: While this solves the “CellProfiler looks for prokaryote in cellprofiler/imagej/jars” problem, I don’t know this is this is the best solution. I suggest we modify CellProfiler to only look for site-packages/imagej/jars. I admit that this is mostly my problem—I should’ve broke python-imagej into a published package when I made the prokaryote changes.

setup.py
```diff
@@ -1,42 +1,151 @@
...
+
+    def initialize_options(self):
+        self.version = None
+
+    def finalize_options(self):
+        if self.version is None:
+            self.version = self.distribution.metadata.version
+
+    def run(self):
+        with open(os.path.join("cellprofiler", "frozen_version.py"),
```
0x00B1: I’m thinking aloud, but I wonder if there’s a consensus best practice for including additional version information that avoids the namespace pollution of a frozen_version.py module. Moreover, missing version information shouldn’t necessarily crash CellProfiler or prevent CellProfiler from building.

But I suppose this is best left to a future commit! :stuck_out_tongue_closed_eyes:

setup.py
```diff
@@ -1,42 +1,151 @@
...
+    has_py2exe = False
+
+#
+# Recipe for ZMQ
+#
+if sys.platform.startswith("win"):
+    #
+    # See http://www.py2exe.org/index.cgi/Py2exeAndzmq
+    # Recipe needed for py2exe to package libzmq.dll
+    os.environ["PATH"] += os.path.pathsep + os.path.split(zmq.__file__)[0]
```
0x00B1: Is this still necessary?

0x00B1:
(I should add the idea of the separate platform ideas didn’t come from Windows. In fact, it came from Debian and an attempt to avoid polluting our source repository with the required “debian” directory and OS X to avoid DMG nonsense and an Info.plist.)

setup.py
LeeKamentsky: OK - let's talk about the different situations and what users and developers should do setupwise to sync their environments

setup.py
0x00B1: You’re much nicer than me.

I intentionally dropped anything required to build a CellProfiler executable on OS X from a virtual environment (e.g. OS X requires someone to crate symbolic links from the system site-packages  wxPython to the virtual environment site-packages and create an app bundle for your virtual environment’s Python—it’s horrible, LOL). I felt it’d be easier to keep documentation and our build system updated with possible py2app and py2exe fixes than our setup.py. I also wanted to maintain platform neutrality (see my other comments).

setup.py
LeeKamentsky: I think it is still needed. There is one sure way to find out - get rid of the code and see if the distribution still runs after you build it.

setup.py
0x00B1: I should correct myself, having a number of commands is common in Ruby and that didn’t bother me too much. Nonetheless, it’s true in Java:

```sh
mvn

mvn test
```

And many C and C++ builds:

```sh
make

make install

make test
```

Hrm, maybe what we’re missing is an equivalent to “make” (i.e. something to run before “python setup.py install”).

This simpler approach also lends itself nicely to our Travis CI configuration, deb and RPM packaging, and updating our Homebrew recipe. They all have built-in processes for the standard Python package installation process.

cellprofiler/utilities/version.py
0x00B1: Yeah, I’m convinced. :smile:

0x00B1: (I’m curious, have any authors included a commit’s SHA-1? I’d be impressed!)

setup.py
LeeKamentsky: Yup. I was thinking at one point of moving the code that's in `install` to `build`. The idea would be to stage things like prokaryote and frozen_version in build/lib as part of the build step. That would share between `install` and `py2exe`.

cellprofiler/utilities/version.py
LeeKamentsky: Never have seen a SHA-1 published. Mark knows the SHA-1 of every CellProfiler release though - if someone posts a pipeline on the forum, he figures out which version of CellProfiler from that.

setup.py
0x00B1: I sketched this out: #1732

LeeKamentsky:
I think the one remaining issue after 5c4d795 is to improve the install by informing the user if their Java version is < 1.6. We can turn that into an issue and handle it after the PR. So I'm marking this as not WIP.

PR #1726, opened 2015-12-07
`python -m cellprofiler` fix

0x00B1:
`if __name__ == "__main__"` is needed for expected `$ python -m cellprofiler` use


0x00B1:
@LeeKamentsky would you mind prioritizing this? It’s blocking PyPI submission.

PR #1729, opened 2015-12-08
Add a classifier mode to FilterObjects

jhung0:
classifier mode in FilterObjects allows classifier saved from CPA to be
use to classify objects (make sure measurement modules are correct)

It seems to work but things need to be added to the interface (detect # of classes, descriptors that distinguish it from rules)


cellprofiler/modules/filterobjects.py
```diff
@@ -367,6 +380,14 @@ def help_settings(self):
     def visible_settings(self):
         result =[self.target_name, self.object_name, 
                  self.spacer_2, self.mode]
+        if self.mode == MODE_CLASSIFIERS:
+            result += [self.rules_file_name, self.rules_directory,
+                       self.rules_class]
+            try:
```
LeeKamentsky: This shouldn't be necessary - the choices should be automatically populated if you make the changes to `get_rules_class_choices`

cellprofiler/modules/filterobjects.py
```diff
@@ -274,6 +280,13 @@ def get_rules_class_choices(self, pipeline):
         except:
             return [str(i) for i in range(1, 3)]
         
+    def get_classifier_class_choices(self, pipeline):
```
LeeKamentsky: I think your best bet is to move the logic in `get_classifier_class_choices` into `get_rules_class_choices` (and maybe rename `get_rules_class_choices` to `get_class_choices`. You can put an if statement to test whether the mode is `MODE_RULES` or `MODE_CLASSIFIERS`.

LeeKamentsky:
Example model is here: https://broadinstitute.org/~leek/pull1729/is_saturated.model

LeeKamentsky:
@braymp @daviddao @jhung0 
Hi guys, I think the code I have should work, but I need a little help before testing. The model above has a couple feature names that are a little wonky (or maybe you do train on the thumbnail info!!!!). I filed https://github.com/CellProfiler/CellProfiler-Analyst/issues/165 to ask you all to remove the thumbnails from the list of feature names and I am guessing x_loc and y_loc are missing too.

Let me know how to resolve - I am thinking of trying to filter out any feature names for which there are no features and see if that works, but it feels like a hack.

LeeKamentsky:
I worked around the "Thumbnail" issue by substituting "0" for any feature in the pickled classifier that's not being measured. So it runs, but I think it gave the wrong result.

jhung0:
I think what happens in CPA is the value is set to 0 if it's not a number (float, int), so this should work

LeeKamentsky:
I used the latest classifier from Mark and it classified correctly and displayed the result for FlagImage.

It remains to do unit tests for the "classify" mode. I will work on those next.

daviddao:
Yeah! Wuhu :)

LeeKamentsky:
Hi guys, time to review.

braymp:
Need a validation error if "Classifier" is chosen as the flag basis but a rules file is input. Making the wrong choice yields this error:
```
  File "C:\Trunk\pull_requests\1729\cellprofiler\gui\pipelinecontroller.py", line 2826, in do_step
    self.__pipeline.run_module(module, workspace)
  File "C:\Trunk\pull_requests\1729\cellprofiler\pipeline.py", line 2067, in run_module
    module.run(workspace)
  File "C:\Trunk\pull_requests\1729\cellprofiler\modules\flagimage.py", line 377, in run
    statistics += self.run_flag(workspace, flag)
  File "C:\Trunk\pull_requests\1729\cellprofiler\modules\flagimage.py", line 518, in run_flag
    flag.measurement_settings[0])
  File "C:\Trunk\pull_requests\1729\cellprofiler\modules\flagimage.py", line 603, in eval_measurement
    classifier = self.get_classifier(ms)
  File "C:\Trunk\pull_requests\1729\cellprofiler\modules\flagimage.py", line 508, in get_classifier
    return self.load_classifier(measurement_group)[0]
  File "C:\Trunk\pull_requests\1729\cellprofiler\modules\flagimage.py", line 504, in load_classifier
    d[path_] = joblib.load(path_)
  File "C:\Python27\lib\site-packages\sklearn\externals\joblib\numpy_pickle.py", line 462, in load
    obj = unpickler.load()
  File "C:\Python27\lib\pickle.py", line 864, in load
    dispatch[key](self)
  File "C:\Python27\lib\pickle.py", line 927, in load_int
    val = long(data)
ValueError: invalid literal for long() with base 10: 'F (Image_ImageQuality_StdIntensity_OrigDNA > 0.0099450900000000002, [-0.61852920660425292, 0.61852920660425292], [0.44249543031342037, -0.44249543031342037])\r\n'
```

braymp:
The display window for FlagImages doesn't show the correct image cycle number in test mode.

braymp:
In the display window, the Measurement column still shows "Rules" even if "Classifier" is selected.

braymp:
The help for the class selection panel says "The flag is set if the image falls into the selected class," which was written towards rules. But it seems that for the "Classifier" mode, the flag is set if the image is *not* in the selected class. Confirm?

LeeKamentsky:
I think so. There's a variable called "fail" that you're supposed to set with your prediction. I set it if the prediction is not in the user's choices. Let me know if I should change it.

braymp:
Yeah, I have to admit I always get this confused. 

I think it's best to swap it. Since the module says "The value of the flag is 1 if the image meets the selected criteria (for example, if it fails QC), and 0 if it does not meet the criteria (if it passes QC)." it should get flagged if it meets the prediction, e.g., for an image to get flagged as "Saturated", you should select "Saturated" as the class.

We might want to change the "Pass/Fail" wording though; sounds like a value judgement :smile:. Maybe have the display column header say "Flagged?" and have it say "Yes" or "No" (that is, "Yes" if it matches the selected class)

LeeKamentsky:
Will switch the polarity. Regarding pass/fail terminology, maybe vote on it on coding day.

LeeKamentsky:
Regarding the display window not updating the image cycle in test mode, I think that may be a general bug - won't address it here.

braymp:
Hmm, that's odd. The other module windows display it correctly, so I assume that it's only FlagImages....

LeeKamentsky:
OK everything should be fixed other than the window's cycle number

braymp:
Haven't checkout FilterObjects but FlagImages seems fine to me.

Re: Pass/Fail - Here, I'm just thinking of the labelling in the display window; the actual 1 or 0 assigned in the measurements remains unchanged.

braymp:
Giving the per-object classification a try, since I think it will be useful. Did the following:
* Ran the examplehuman pipeline with the addition of creating a db.
* In CPA, trained for the pH3+ cells and saved a classifier.

Now, the hitch here is that the classifier is made with the combo of nuclei/cell/cytoplasm features, which is typical. However, in FilterObjects, you can only filter on features from only one object at a time. So I selected 'Nuclei' and expected FilterObjects to fail, which it did:
```
  File "C:\Users\brayma2\Desktop\GitHub\pull_requests\1729\cellprofiler\gui\pipelinecontroller.py", line 2826, in do_step
    self.__pipeline.run_module(module, workspace)
  File "C:\Users\brayma2\Desktop\GitHub\pull_requests\1729\cellprofiler\pipeline.py", line 2067, in run_module
    module.run(workspace)
  File "C:\Users\brayma2\Desktop\GitHub\pull_requests\1729\cellprofiler\modules\filterobjects.py", line 484, in run
    indexes = self.keep_by_class(workspace, src_objects)
  File "C:\Users\brayma2\Desktop\GitHub\pull_requests\1729\cellprofiler\modules\filterobjects.py", line 911, in keep_by_class
    for feature_name in features])
  File "C:\Users\brayma2\Desktop\GitHub\pull_requests\1729\cellprofiler\measurements.py", line 364, in __getitem__
    return self.get_measurement(*key)
  File "C:\Users\brayma2\Desktop\GitHub\pull_requests\1729\cellprofiler\measurements.py", line 960, in get_measurement
    vals = self.hdf5_dict[object_name, feature_name, image_set_number]
  File "C:\Users\brayma2\Desktop\GitHub\pull_requests\1729\cellprofiler\utilities\hdf5_dict.py", line 360, in __getitem__
    result = self[object_name, feature_name, [num_idx]]
  File "C:\Users\brayma2\Desktop\GitHub\pull_requests\1729\cellprofiler\utilities\hdf5_dict.py", line 364, in __getitem__
    assert feature_exists
AssertionError
```
My only rec here would be a more informative statement as to which feature is missing (or show the first missing, if many).

braymp:
The classifier model works in FilterObjects if set <code>classifier_ignore_columns  =  .*Cells.*, .*Cytoplasm.*</code>, i.e. exclude anything but nuclei features. So :+1: from me for the merge.

braymp:
@LeeKamentsky: Any chance you might be able to resolve these conflicts and take the PR?

LeeKamentsky:
OK, hopefully editing the merge conflicts did not introduce problems. I am going to merge if it passes tests.

PR #1732, opened 2015-12-10
refactored `get_path_to_jars`

0x00B1:


LeeKamentsky:
This has a good story to tell: "sys.path is the search order for Python modules. You resolve module locations using sys.path to establish priorities. We search for prokaryote the same way."

cellprofiler/utilities/cpjvm.py
```diff
@@ -29,16 +29,18 @@
...
-
-    imagej_path = os.path.join(root_path, 'imagej','jars')
-    return imagej_path
+
+    pathnames = [
+        os.path.join(
+            directory, "imagej", "jars"
+        ) for directory in sys.path if os.path.exists(
+            os.path.join(
+                directory, "imagej", "jars", "prokaryote-1.0.4.jar"
```
LeeKamentsky: Figure out some way to specify the version # as stored in setup.cfg.

0x00B1:
Yeah! Exactly. :smile:

PR #1733, opened 2015-12-15
use the prokaryote python package

0x00B1:


0x00B1:
@LeeKamentsky this is an alternative to #1732

LeeKamentsky:
@0x00B1 let me know when you've updated the prokaryote version on PyPi so it works. Basically, this is all really positive, I have no corrections and I'll take the PR (or you can) after it passes the build.

PR #1742, opened 2015-12-21
patch for issue #1553 - error from classifyobjects

erexhepa:
The bit of code I added is between lines 295 and 302.
Elton


LeeKamentsky:
Alternate PR accepted

PR #1743, opened 2015-12-21
Fixes #1553 - handle situation with primary objects when measuring t…

### LINKED ISSUE #1553: error from classifyobjects module
Hi,

I was getting an error from the excecution of the classifyobject.py from our local installation of cellprofiler. I started to debug it and error is raised at  the level of 

def run_single_measurement(self, group, workspace):

at this line

labels = object_color[objects.segmented]

and it is an indexing error due to the objects.segmented having a few objects more then the entries in object_color. What happens is that the measurement upon which the classification is based is a two channel correlation measure. Before measuring the correlation the two channels are filtered using a binary mask and the labels of a few objects on the measurecorrelation.py are filtered out by the binary mask and measurecorrelation.py stores fewer values then the number of objects in the workspace. 

Since we are computing other metrics in others channels for all the objects and compare them with the correlation feature ideally the correlation should be a NaN for the masked objects during the correlation. I did a few changes in measurecorrelation.py 

line 292, changed the AND to an OR
mask   = ((labels > 0) | first_mask & second_mask)

seems to work but then I need to put to NaN the value for the masked objects and was wondering if it is safe to do it at the measurecorrelation.py level or classifyobjects.py ? If I set the NaN at the measurecorrelation.py I get errors in other modules but was wondering if it is because of this. 

Anyway trying to figure out the errors above but thought of posting this because if you filter the masked objects during the correlation and you get less correlation values than there are objects then the are the correlation values correctly indexes to the objects when added to the workspace?

LeeKamentsky:
…he correlation between two channels A and B and both channels are masket out"

"Fixes #1553 - handle situation with primary objects when measuring the correlation between two channels A and B and both channels are masket out"

Corrected rebasing against cellprofiler master branch

### DIFF
--- a/cellprofiler/modules/measurecorrelation.py
+++ b/cellprofiler/modules/measurecorrelation.py
@@ -281,8 +281,13 @@ class MeasureCorrelation(cpm.CPModule):
         second_pixels = second_pixels[mask]
         labels = labels[mask]
         n_objects = objects.count
-        if n_objects == 0:
+        # Handle case when both images for the correlation are completely masked out
+
+        if ((n_objects == 0)):
             corr = np.zeros((0,))
+        elif((np.where(mask)[0].__len__()==0)):
+            corr = np.zeros((n_objects,))
+            corr[:]=np.NaN
         else:
             #
             # The correlation is sum((x-mean(x))(y-mean(y)) /


LeeKamentsky:
@erexhepa I think I've captured it with you still as the author. If it looks OK, let me know and I'll merge it in.

erexhepa:
Hi, yeah it looks OK to me.

PR #1770, opened 2016-01-26
RunScript module

LeeKamentsky:
This is my first attempt at Mark Bickle's request for a module that can run a Python script. RunScript will initialize the script with images, objects and measurements of your choosing, stored in variables that are accessible from within the script and it will collect values out of variables after the script is run and store them in images, objects and measurements.

No unit tests yet so you should assume it doesn't work, but I used it and it worked for me.


LeeKamentsky:
Added unit tests. @shntnu if you want to try... ready for review too..

shntnu:
@LeeKamentsky I'll get to it right after we've figured out #1775. RunScript sounds awesome!

AnneCarpenter:
I think the concern here is security risk and we have decided not to add to this to CP. The code could nonetheless serve as a model for those who want to use it locally. What is the procedure from here?

braymp:
I have at least one colleague here who is interested in a module with this capability; I knew there was work but couldn't find it until now!

mcquin:
@AnneCarpenter Agree this will not be added to CellProfiler due to security concerns. Closing the pull request will not delete the branch so that @LeeKamentsky @shntnu and @braymp can link to this as a reference (though such a module is highly discouraged!!).

PR #1771, opened 2016-02-03
WIP: Fixes #1761

### LINKED ISSUE #1761: ExportToDatabase: Add props fields for image classification
From https://github.com/CellProfiler/CellProfiler-Analyst/pull/124, there are a few items that should be added to the props file to make things work:
* image_classification: Add as yes/no setting to ExportToDatabase
* image_width, image_height: Add and assign values automatically. 

If the question posed in https://github.com/CellProfiler/CellProfiler-Analyst/issues/127 comes back as 'no', then complete this issue, otherwise close.

LeeKamentsky:


LeeKamentsky:
@braymp I think I got all the pieces. If you want to review and try it out, that would be great - if it's ok, merge it, if not, get back to me.

braymp:
According to @daviddao, @jhung0, turns out that figuring out the image width and height is hard from the CPA standpoint, unless it's specified in the props file. They'd prefer CP provide that info if at all possible.

How tough would it be to have CP query a single image for its dimensions in prepare_run?

LeeKamentsky:
If I need to, I can do it. It's not so much work, just architecturally very bad. It would be similar to them reading the first record out of the image table and using the image width and height from there, pretty much just as architecturally bad.

braymp:
Actually, scratch that. I think I'll just have classifier/imagegallery
query the db on startup to find the needed params.

braymp:
Done with https://github.com/CellProfiler/CellProfiler-Analyst/commit/9bf5fc6f6975e51c80d76d821da28b6852f040b3

PR #1774, opened 2016-02-05
WIP: pixel classifier

LeeKamentsky:
The idea here is to use CellProfiler to train a classifier by feeding it objects representing ground truth and gray-scale images representing the channels. Underlying, the data structure is 5D, but only channel, I and J are used.

A sampling window is taken around each pixel to create a feature vector. Dimensionality reduction is done via PCA to get a set of eigentextures. The output of these eigentexture filters creates a reduced feature vector for a classifier, e.g. random forests. Currently, things are trained using randomly selected pixels and then two rounds of training are done with random pixels augmented with misclassified pixels.

Classification applies the kernel to the image (in 128x128 chunks), runs the eigentexture filters to reduce the dimensionality and then classifies using the trained random forest.


LeeKamentsky:
As of 2/8, I think it's operating correctly, but slowly, especially the training phase. I am creating the per-pixel feature vectors and writing them to disk and it seems that disk transfer is the source of almost all of the time used to train. For any reasonable training set, the number of samples drawn is far less (10% or fewer) than the number of examples in the ground truth, so it may be cheaper to store the images and create the feature vectors for the training set on the fly (compute time << disk access time).

I am in the middle of a run using 6 images from the Moy C. elegans screen and in the initial pass, 97% of the background pixels were classified correctly. There are two further rounds, each of which augment the training set from the misclassified pixels, so the final will be much better.

LeeKamentsky:
My first classified image. No illumination correction, training done on another plate.

first_light

0x00B1:
@holgerhennig

0x00B1:
@LeeKamentsky can I help with this? Let me know! :)

LeeKamentsky:
You can now use label .tif files exported by Ilastik in addition to objects. I took the liberty of not doing backwards compatibility because this isn't on the master branch yet, so old pipelines using the classify module will not load correctly.

0x00B1:
> I took the liberty of not doing backwards compatibility because this isn't on the master branch yet, so old pipelines using the classify module will not load correctly.

Could you use an alternative name to mitigate this issue?

LeeKamentsky:
It does use an alternative name - the old one that uses Ilastik is called ClassifyPixels and this one is called Classify. But if David checked out the branch this morning and made a pipeline using the Classify module, it would break. Almost certainly, David is the only person who will suffer from backwards compatibility - just wanted to warn him.

0x00B1:
Ah, thanks for the clarification!

holgerhennig:
Great, I'll test the pixelbased classifier on a different project, will let you know how it goes

0x00B1:
It’d be nice if we could depreciate ClassifyPixels entirely. It’d help cleanup `cellprofiler.__main__.py`.

holgerhennig:
Good idea to depreciate ClassifyPixels from a users perspective, it's confusing when there's an old classifypixels and the new classify

0x00B1:
@LeeKamentsky Will you upload your example pipeline (i.e. `c:/temp/output/classify/classify.cpproj`)?

cellprofiler/modules/classify.py
```diff
@@ -0,0 +1,952 @@
...
+
+        :param filter_bank_name:
+        :param sample:
+
+        :return:
+        """
+        dataset = self.g_filters[filter_bank_name]
+
+        if dataset.attrs[A_CLASS] == CLASS_FILTER:
+            if USE_DOT:
```
0x00B1: `USE_DOT` is always `True`

cellprofiler/modules/classify.py
```diff
@@ -0,0 +1,952 @@
...
+                result = numpy.dot(sample, dataset[:].T)
+            else:
+                #
+                # A dot product... but cluster's np.dot is so XXXXed
```
0x00B1: LOL, is this true?

cellprofiler/modules/classify.py
```diff
@@ -0,0 +1,952 @@
...
+
+        self.path = cellprofiler.settings.DirectoryPath("Classifier folder")
+
+        self.filename = cellprofiler.settings.FilenameText("Classifier file", "Classifier.cpclassifier", get_directory_fn=get_directory_fn,
+                                                           set_directory_fn=set_directory_fn,
+                                                           exts=[("Pixel classifier (*.cpclassifier)", "*.cpclassifier"), ("All files (*.*)", "*.*")])
+
+        self.gt_source = cellprofiler.settings.Choice("Ground truth source", [SRC_OBJECTS, SRC_ILASTIK],
```
0x00B1: `SRC_ILASTIK` is always `"Ilastik"`

cellprofiler/modules/classify.py
```diff
@@ -0,0 +1,952 @@
...
+        Set the kernel used to sample pixels from a neighborhood
+
+        :param kernel: an N x 5 matrix of N offsets by 5 dimensions (C, T, Z, Y, X)
+
+        :return:
+        """
+        if DATASET_KERNEL in self.root.keys():
+            del self.root[DATASET_KERNEL]
+
+        self.root.create_dataset(DATASET_KERNEL, data=kernel).attrs[A_CLASS] = CLASS_KERNEL
```
0x00B1: `CLASS_KERNEL` is always `"Kernel"`

cellprofiler/modules/classify.py
```diff
@@ -0,0 +1,952 @@
...
+
+        :param classifier_name: save using this name
+        :param sample: S samples x N features
+        :param classes: S class labels indexing into the class names
+        :param algorithm: algorithm to use to train
+
+        :return:
+        """
+        if algorithm is None:
+            algorithm = sklearn.ensemble.ExtraTreesClassifier(n_estimators=N_ESTIMATORS, min_samples_leaf=MIN_SAMPLES_PER_LEAF)
```
0x00B1: `N_ESTIMATORS` is undefined

0x00B1: `MIN_SAMPLES_PER_LEAF` is undefined

cellprofiler/modules/classify.py
```diff
@@ -0,0 +1,952 @@
...
+
+        self.min_samples_per_leaf = cellprofiler.settings.Integer("Minimum samples per leaf", DEFAULT_MIN_SAMPLES_PER_LEAF, 1,
+                                                                  doc="This setting determines the minimum number of ground truth pixels that the classifier will use to split a decision tree. There must be at least this number of example pixels in each branch for the classifier to have confidence that the split is real and not just an artifact of an irrelevant measurement.\n Lower this setting if the classifier does a good job on most of the pixels but does not draw sharp distinctions between one class and another at the border between the classes (e.g. at the edges of cells). Raise this setting if the classifier misclassifies pixels that are clearly not the right class - this is overtraining.")
+
+        self.path = cellprofiler.settings.DirectoryPath("Classifier folder")
+
+        self.filename = cellprofiler.settings.FilenameText("Classifier file", "Classifier.cpclassifier", get_directory_fn=get_directory_fn,
```
0x00B1: Can you use lambda rather `get_directory_fn`? What about `set_directory_fn`?

0x00B1: Can you use lambda rather `get_directory_fn`? What about `set_directory_fn`?

cellprofiler/modules/classify.py
0x00B1: Shouldn’t this be the default argument for `algorithm` (rather than “None”)?

cellprofiler/modules/classify.py
```diff
@@ -0,0 +1,949 @@
...
+                result += group.visible_settings()
+
+            result.append(self.add_output_button)
+
+        return result
+
+    def prepare_settings(self, settings):
+        for count, sequence, add_fn in zip([int(_) for _ in settings[:4]], (self.object_classes, self.label_classes, self.images, self.outputs),
+                                           (self.add_objects, self.add_labels, self.add_image, self.add_output)):
+            del sequence[:]
```
0x00B1: Why the explicit memory de-allocation?

cellprofiler/modules/classify.py
```diff
@@ -0,0 +1,949 @@
...
+        return self.g_training_set[class_name]
+
+    @property
+    def ground_truth_chunk_size(self):
+        """The size of a chunk of ground truth that fits in memory"""
+        return int(50 * 1000 * 1000 / self.get_kernel().shape[0])
+
+    @property
+    def pix_chunk_size(self):
+        return 1000 * 1000
```
0x00B1: ???

cellprofiler/modules/classify.py
LeeKamentsky: feel free to change USE_DOT and find out

cellprofiler/modules/classify.py
LeeKamentsky: It's not a deallocation, it's deleting the elements of the list.

cellprofiler/modules/classify.py
LeeKamentsky: For get_directory_fn, a lambda function is possible, for set_directory_fn, it's a stretch and only possible because of the order of outputs from self.path.get_parts_from_path.

    set_directory_fn=lambda path: self.path.join_parts(*self.path.get_parts_from_path(path))

cellprofiler/modules/classify.py
LeeKamentsky: Good catch on N_ESTIMATORS and MIN_SAMPLES_PER_LEAF. That's a legacy of me fooling around with the classifier when those were static variables.

LeeKamentsky:
Projects that train and classify (with not so good ground truth) are at /imaging/cold/People/Lee/wormclassifier

holgerhennig:
Testing the module Classify (pixel-based classifier) with Myelin sheaths images. This is more of a tough example to test with (not necessarily expecting great results w/ a pixel-based classifier). Biological aim is to differentiate between complete (marked green) and incomplete tubes (marked red/purple) in image

General comments/recommendations
- module classify works great! 2 basic steps: train, classify
- improve performance: currently several minutes (each for training and classifying)
- training mode: Classifier.cpclassifier file is huge (~1G) compared to input image file (7KB)
- so far: PCA features are calculated by the classify module. Add other features (texture etc)? 
- As a guide for users to know when the cpclassifier file is either being created (train mode) or read (classify mode), I'd recommend: In train mode, replace the text "Classifier file" by "Output classifier file". In classify mode, replace the text "Classifier file" by "Input classifier file"
- class label names need to be inserted manually (not read from file), but is no big deal when having only few classes

General workflow involving labeling in Ilastik
1) label the classes (here we have 4 classes: complete, incomplete, background, junk) in Ilastik and export labels as tiff, e.g. labels.tif
2) in CP's module classify in train mode, choose Ilastik as ground truth input and point to labels.tif. Train the classifier
Note: A separate CP pipeline for training and classifying needs to be built
3) Build pipeline using classify module in classify mode, this generates the probability maps. Then identify primary objects (complete, incomplete myelin sheaths) based on the probability maps

Results of CP pixel-based classifier:
overlay008

For comparison: labeling of a biologist (not used for training the classifier)
max_021616expt series008 -mbp example annotated

LeeKamentsky:
@holgerhennig 
Performance: there are two factors that determine performance during training. The first is the radius and a 2x increase in radius increases the processing time and memory by 4x. As you get further away from the pixel of interest, there is less correlation between a pixel and the  one of interest, so you could use a strategy that produces fewer pixel features at larger radii for input to PCA. For instance, you could use bicubic spline interpolation to sample the annulus from radius / 2 to radius at 1/2 the resolution.
The second factor that influences performance during training (but not classification) is the size and number of rounds of training. I subsample from the ground-truth data currently and run three rounds to try to zero in on the pixels that are at the error margins of the classifier. So you could reduce the training time linearly with the # of rounds or the # of pixels per round.

Classifier file size: I use the classifier file as temp storage during training. There's no way to shrink the size of an HDF file, so all that temp storage remains after the file has been used. It would be pretty straightforward to copy only the datasets that are needed for classification - these are pretty small.

Adding features: I wish I had a tool that would let you visualize the eigentextures - when you have ~100 of them, they are very comprehensive. Perhaps something like eigenvalues of the hessian might be useful because it's not a convolution. Perhaps just using SIFT or SURF instead of PCA would work too. But my gut is that adding features will not improve performance noticeably.

Class labels: these are not available from the .tif files output from Ilastik (yes, I looked through the TIF metadata). Possibly they are from the .HDF file and they definitely are available if you use CellProfiler .tif object files.

bethac07:
Has anyone successfully gotten this working using the "Object" based classification?  I've been trying to use it and after training it in four slightly different ways (using a full image set, using a subset of the images, using all of the channels, using only one of the channels), each one produced a probability map that was just pure white.  :-/

dlogan:
Not me.  @holgerhennig did though.  Have you tried looking at the raw data, e.g. in Matlab?  Maybe the values (erroneously) range from [1,N] and thus look white?

bethac07:
Thanks @dlogan , I saw @holgerhennig had done it using ilastik as an input but didn't know he'd gotten it working with objects as well.  

I haven't tried that but I'll try it next; I did mouse over it and it was a sea of 1.00000 but I'll export it and take a look.

mcquin:
Is anyone committed to finishing this work? @LeeKamentsky @0x00b1

AnneCarpenter:
There IS a plan to implement pixel classification via a better connection to ilastik #2393 - that work might make this less necessary.

PR #1777, opened 2016-02-13
Updates by VS to include co-localization measures including Overlap, …

VRSingan:
…Manders, Costes and Rank Weighted co-localization

### DIFF
--- a/cellprofiler/modules/measurecorrelation.py
+++ b/cellprofiler/modules/measurecorrelation.py
@@ -1,13 +1,11 @@
-'''<b>Measure Correlation</b> measures the correlation between intensities in different images (e.g.,
-different color channels) on a pixel-by-pixel basis, within identified
+'''<b>Measure Correlation</b> measures the colocalization and correlation between intensities in different images (e.g., different color channels) on a pixel-by-pixel basis, within identified
 objects or across an entire image.
 <hr>
-Given two or more images, this module calculates the correlation between the
-pixel intensities. The correlation can be measured for entire
-images, or a correlation measurement can be made within each
-individual object.
+Given two or more images, this module calculates the correlation & colocalization (Overlap, Mander's, Costes' Automated Threshold & Rank Weighted Colocalization) between the
+pixel intensities. The correlation / colocalization can be measured for entire
+images, or a correlation measurement can be made within each individual object.
 
-Correlations will be calculated between all pairs of images that are selected in 
+Correlations / Colocalizations will be calculated between all pairs of images that are selected in
 the module, as well as between selected objects. For example, if correlations 
 are to be measured for a set of red, green, and blue images containing identified nuclei, 
 measurements will be made between the following:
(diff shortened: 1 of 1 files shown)


cellprofiler/modules/measurecorrelation.py
```diff
@@ -83,6 +116,13 @@ def create_settings(self):
         self.spacer_2 = cps.Divider(line=True)
         
         self.add_object_button = cps.DoSomething("", 'Add another object', self.add_object)
+        
+        self.thr = cps.Float(
+            "Set threshold as percentage of maximum intensity for the images",
+            15, minval=0, maxval=99, doc='''\
+            Select the threshold as a percentage of the maximum intensity of the above image [0-99].''')
```
LeeKamentsky: If a setting is added to a module, the following changes need to be made to ensure backward compatibility with old pipelines:

- The `variable_revision_number` needs to be incremented to distinguish the old settings from the new.
- The module's `upgrade_settings` method needs to be modified to convert the settings format of the old variable revision into the new.

At the same time, we like to add a test to the module's test suite to check that a pipeline with the new settings format can be loaded.

cellprofiler/modules/measurecorrelation.py
```diff
@@ -235,20 +277,155 @@ def run_image_pair_images(self, workspace, first_image_name,
...
+                        
+            # Costes' thershold calculation
+            combined_thresh_c = (fi > Thr_fi_c) & (si > Thr_si_c)
+            fi_thresh_c = fi[combined_thresh_c]
+            si_thresh_c = si[combined_thresh_c]
+            tot_fi_thr_c = fi[(fi > Thr_fi_c)].sum()
+            tot_si_thr_c = si[(si > Thr_si_c)].sum()
+            
+            #Threshold as percentage of maximum intensity in each channel
+            thr_fi = self.thr.value * scistat.scoreatpercentile(fi,100.0) / 100
```
LeeKamentsky: A couple things about scoreatpercentile:

The docs for scoreatpercentile indicate that it will become obsolete in the future and that users should use np.percentile for Numpy 1.9+. So one approach would be to do something like:

    try:
        from numpy import percentile
    except ImportError:
        from scipy.stats import scoreatpercentile as percentile

Is scistat.scoreatpercentile(x, 100.0) the same as np.max(x)? If so, that should be the choice.

cellprofiler/modules/measurecorrelation.py
```diff
@@ -305,9 +517,163 @@ def run_image_pair_objects(self, workspace, first_image_name,
...
             corr[scind.sum(1, labels, lrange) == 0] = np.NaN
+            result += [[first_image_name, second_image_name, object_name,"Mean Correlation coeff","%.3f"%np.mean(corr)],
+                       [first_image_name, second_image_name, object_name,"Median Correlation coeff","%.3f"%np.median(corr)],
+                       [first_image_name, second_image_name, object_name,"Min Correlation coeff","%.3f"%np.min(corr)],
+                       [first_image_name, second_image_name, object_name,"Max Correlation coeff","%.3f"%np.max(corr)]]
+
+            # Threshold as percentage of maximum intensity of objects in each channel
+            tff = (self.thr.value/100) * fix(scind.maximum(first_pixels,labels, lrange))
+            tss = (self.thr.value/100) * fix(scind.maximum(second_pixels,labels, lrange))
+
```
LeeKamentsky: There's a problem in scipy.ndimage.sum if there are no values for a label:

    > scind.sum(np.zeros(0), np.zeros(0, int), np.arange(1, 2))
      Python Shell, prompt 8, line 1
      File "C:\Python27\Lib\site-packages\scipy\ndimage\measurements.py", line 580, in sum
        count, sum = _stats(input, labels, index)
      File "C:\Python27\Lib\site-packages\scipy\ndimage\measurements.py", line 499, in _stats
         labels.min() < 0 or labels.max() > labels.size):
      File "C:\Python27\Lib\site-packages\numpy\core\_methods.py", line 21, in _amin
        out=out, keepdims=keepdims)
    ValueError: zero-size array to reduction operation minimum which has no identity

I am going to fix this so that you sum over the values * the truth array. That will give you values of zero for masked pixels which is the same as leaving them out.

VRSingan:
Hi Lee, I've made two changes. Can you please review for merge ?

LeeKamentsky:
great - thx for the last fix. All done

PR #1792, opened 2016-03-01
Fixes #1714 - kirsch filter + tests

### LINKED ISSUE #1714: Add Kirsch filter to EnhanceEdges
A CP workshop attendee suggested the the [Kirsch filter](https://en.wikipedia.org/wiki/Kirsch_operator) works well for pre-processing his brightfield images for segmentation. Might be worth adding.

LeeKamentsky:


0x00B1:
2.2.0-rc3

cellprofiler/modules/enhanceedges.py
```diff
@@ -26,6 +27,7 @@
 M_ROBERTS = "Roberts"
 M_LOG = "LoG"
 M_CANNY = "Canny"
+M_KIRSCH = "Kirsch"
```
0x00B1: In the future, we should consider using [enumerations](https://www.python.org/dev/peps/pep-0435/) rather than constants (they’ve been [back-ported to 2.7](https://pypi.python.org/pypi/enum34)).

0x00B1:
:+1:

PR #1804, opened 2016-03-02
removed splash screen

0x00B1:
:hamster: :smile:


braymp:
This still makes me cry a bit on the inside :sob: But do what you must.

PR #1813, opened 2016-03-03
replaced MySQLdb with PyMySQL

0x00B1:


LeeKamentsky:
I'm thinking of doing #1814 so that this can be tested. Sounds good? Worth holding off merge until it passes after #1814 is fixed?

0x00B1:
@LeeKamentsky Sounds good!

LeeKamentsky:
@0x00B1 needs merge confilcts resolved and then I will take PR.

PR #1835, opened 2016-03-10
Issues/1834

LeeKamentsky:
The median filter operates using an octagon and you need a reasonable kernel size in order to draw the octagon over the image (e.g. a 3x3 kernel is too small). Also, there's no point speeding it up since white_tophat runs just as fast or faster.


LeeKamentsky:
Needed to take this to fix the "contrib" problem, sorry.

PR #1837, opened 2016-03-10
Fixes #1834 limit object diameter to 3 for speckles only

### LINKED ISSUE #1834: EnhanceOrSuppressObjects -> Speckles can produce values < 0
Run zipped pipeline with the image provided (I ran Mac 2.2.0rc3).
[lysosomes_structures_DLogan.cppipe.zip](https://github.com/CellProfiler/CellProfiler/files/167535/lysosomes_structures_DLogan.cppipe.zip)
8b8af277acc313079835cefd8f2cb0c6f5232ab1

EnhanceOrSuppressFeatures looked odd in Image Contrast = Normalized mode, but it turns out that it is because (I think) the lowest intensity values are < 0. I think they should be [0,1] valued.

image

Zoomed in, you can see some darker, < 0 value pixels:
image

LeeKamentsky:


cellprofiler/settings.py
```diff
@@ -722,6 +722,25 @@ def set_value_text(self, value_text):
             self.__default = self.str_to_value(value_text)
         except:
             logger.debug("Number set to illegal value: %s" % value_text)
+            
+    def set_min_value(self, minval):
```
0x00B1: Would it be possible to use a field?

```Python
class Example:
    self.maxval = None

    self.minval = None
```

If not, would you mind using the property decorator?

```Python
class Example:
    @property
    def min_value(self):
        return self._minval

    @min_value.setter
    def min_value(self, value)
        self._minval = value
```

0x00B1: `maximum` and `minimum` are more descriptive names than `max_value` and `min_value`

cellprofiler/settings.py
```diff
@@ -722,6 +722,25 @@ def set_value_text(self, value_text):
...
+        '''Programatically set the minimum value allowed'''
```
0x00B1: A docstring for a property is overkill. Especially when the property has a descriptive name.

cellprofiler/settings.py
```diff
@@ -722,6 +722,25 @@ def set_value_text(self, value_text):
...
+        self.__maxval = maxval
+        
+    def get_min_value(self):
+        '''The minimum value (inclusive) that can legally be entered'''
+        return self.__minval
+    
+    def get_max_value(self):
+        '''The maximum value (inclusive) that can legally be entered'''
+        return self.__maxval
+
```
LeeKamentsky: The accessors are added with the property() function below. You can see above how `self.object_size.min_value` is referenced. If you want the separate decorators for stylistic reasons, feel free to change.

0x00B1: Why can’t you use a field?

LeeKamentsky: Meh, legacy. I had hidden them previously, not seeing any reason why I'd ever want to expose them... and when I did, I figured there would be less chance of code breakage to add accessors than to search and replace all occurrences of the original field. Was my personal choice, going for the path that would risk the least amount of breakage at the expense of style, feel free to change it.

PR #1853, opened 2016-03-15
Fix a problem with a binary skimage wheel on pypi

LeeKamentsky:
Trying to see why build is breaking first, then will fix.


LeeKamentsky:
@0x00B1 if you want to get your builds to work, you can cherry-pick cab1adc. It looks like the scikit-image developers uploaded a bogus binary wheel which broke our build.

0x00B1:
Thanks, @LeeKamentsky!

PR #1957, opened 2016-04-08
Fix #1789, should import from cpfigure_tools

### LINKED ISSUE #1789: ReassignObjectNumbers: Per-parent unification performing inappropriate masking
In the attached pipeline (to be run with the two sample images), the three objects are made children of an overlapping parent object; these are unified per-parent in ReassignObjectNumbers. 

RelateObjects shows the desired outcome: the 3 objects are unified, with original boundaries intact.

However, in ReassignObjectNumbers, the 3 objects are basically masked, then unified.

untitled-2
untitled-1
[issue.cppipe.txt](https://github.com/CellProfiler/CellProfiler/files/151634/issue.cppipe.txt)

dlogan:


dlogan:
Merging myself, since Beth and I need this for a build, and it seems a pretty straightforward change.

0x00B1:
:+1:

PR #2099, opened 2016-06-17
moduleview: added missing self. to refresh_pending

emmenlau:
To avoid repeated refresh, and to avoid auto-selection of text boxes.


0x00B1:
Slick!

0x00B1:
Thanks, @emmenlau!

PR #2122, opened 2016-06-27
parallelized scikit-image operation with apply_parallel

0x00b1:


braymp:
Aren't there a few other modules where median is performed, e.g., Smooth?

mcquin:
We are not yet ready to begin `apply_parallel` work.

Issue #876, opened 2013-10-28
Add functionality for object consolidation


braymp:
It would be nice to have a module which permits the combination of multiple object sets into one object set. For a disjoint collection of objects, this module wold simply be the relabeled union of them.

If the object sets spatially intersect, some decisions will need to be made. Perhaps an option to handle such regions by either (a) creating a new object from the intersection or union of the intersecting objects, or (b) perhaps removing the intersection.

There definitely needs to be "Add another object" buttons for this one.

Perhaps this could be included in ReassignObjectNumbers or MaskObjects?

AnneCarpenter:
Work is in progress on this, name is TBD: https://github.com/CellProfiler/CellProfiler/pull/2151

0x00b1:
Supplanted by https://github.com/CellProfiler/CellProfiler/pull/2151

Issue #632, opened 2013-05-09
Ponder whether ReassignObjectNumbers needs a name change


braymp:
The module that is now ReassignObjectNumbers was formerly RelateObjects in CP 1.0, but the name doesn't make clear what it does. We're wondering whether another name change is in order.

@AnneCarpenter thinks SplitOrMergeObjects is a good name but is suspicious that we must've thought about that and decided against it for some reason.

AnneCarpenter:
Note that one option within the module is to split or merge based on distance (neighbors) but the other option is per-parent so SplitOrMergeObjects is probably better than SplitOrMergeNeighbors.

bethac07:
It's definitely a confusing name from an end-user point of view, I'm +1 for changing it (and agnostic between Anne's two preferences).

braymp:
:+1: for changing as well. I think I prefer SplitOrMergeObjects; it's generic enough to capture what's there, while also encompassing other similar operations we might add later.

AnneCarpenter:
I'd say with two of the world's best assay developers, plus a CP co-founder, agreeing on this, we should re-open this issue. Doing so now. SplitOrMergeObjects it is.

We should change the wording in the module - we carefully use "Unify" instead of "Merge" inside the module to reflect the possibility that we might be assigning the same object number to non-spatially-contiguous objects. But I think that it's ok to use Merge to describe that. Do others disagree?

We should also add some default text in the module window for this module (and also in the overview section of the help for the module) to state that the IdentifyPrimaryObjects module has options for more complex splitting and merging behaviors.

mcquin:
A decision has been made. Closing in favor of https://github.com/CellProfiler/CellProfiler/issues/2697

PR #2151, opened 2016-07-18
[WIP] MergeObjects module merges two objects.

mcquin:
Merge two objects into one object using the MergeObjects module. TODOs:
1. Decide what to do when two distinct objects overlap. One suggestion is to merge them together. Another suggestion is to follow the MaskObjects approachand split the overlapping segments across their intersection.
2. Set object labels and IVJ values. Thankfully, these support overlapping objects. That is, we should be able to give one point multiple labels. We should only do this if a) this is the correct interpretation of this functionality and b) we decide to not join overlapping segments.
3. Decide if there is a more appropriate name for this module. Current suggestions include: UnionObjects, CombineObjects, ______.
4. Other TODOs may be listed in the code. Do them. Also, fix failing tests.

/cc @0x00B1


bethac07:
Mark and Lee were discussing "MergeIdentifiedObjects" in #876 as a name; I think that using "merge" even with the "Identified" in there might give the wrong impression, perhaps CombineObjectTypes or CombineDifferentObjects ?

braymp:
I think "CombineObjects" would get the point across.

One question, though: Was placing this as an option in ReassignObjectNumbers given consideration? (granted, it's one of the worst named modules ever, but still...)

AnneCarpenter:
My votes:
1. Definitely offer users all possible options for what to do with overlaps because they all make sense for various applications: delete overlapping pixels altogether, include overlapping pixels in both objects, assign overlapping pixels to class A, assign overlapping pixels to class B, merge the two objects and give the resulting object a single label, and (bonus points), split the overlapping pixels evenly (i.e., using Voronoi). 

3. Name: definitely not MergeObjects because it implies objects will be physically merged with nearby neighbors, which will be only one of the options. UnionObjects gets my downvote because users won't typically know that word and it's not a verb; UniteObjects isn't much better for the same reason as Merge. So far CombineObjects seems the best candidate. Maybe even CombineObjectTypes? Adding "Types" may make things a lot clearer whatever name we choose; suddenly MergeObjectTypes doesn't seem so bad. @karhohs and @bethac07 and @braymp should ponder the name thing deeply.

5. Make a new issue offering the user to choose a percentage overlap above which objects are merged and below which they are split (I think it will be relatively common that you want objects to remain separate if they are only overlapping a teeny bit, but merged if they are overlapping a lot).

braymp:
I'm not keen on the "Types" suffix since it's not clear what "type" refers to or means. E.g., MaskObjects and RelateObjects are fine as names, and both can manipulate any objects the user choses.

I like the percentage idea; MaskObjects has the same option. It also need to be made clear in the setting text which object the percentage refers to. Since I'm imagining the settings are arranged much like MaskObjects, there could be two settings:
* Perform operation depending on overlap?: yes/no (using "operation" to keep it generic)
* (Expose if "yes" to above) Fraction of allowed overlap with first object: Float between 0,1 (to be consistent with MaskObjects)

AnneCarpenter:
Another todo, per https://github.com/CellProfiler/CellProfiler/issues/876 

"Add another object" button, in case the user has three or more object types that they want to combine.

I don't know how valuable this is, it's just that Mark requested it.

AnneCarpenter:
FYI, we've settled on renaming ReassignObjectNumbers to be SplitOrMergeObjects in this other issue: https://github.com/CellProfiler/CellProfiler/issues/632  
 
For this new module you're working on, we could do things step-wise: initially, just make this standalone module called whatever you like, maybe CombineObjects. 

But eventually, it would embed into SplitOrMergeObjects. You may wish to go straight to this route right away so we don't need to build any backwards compatibility for a module that will soon be outdated. 

It would keep the current first Q: "Select the input objects" and the next Q would be "Do you want to merge with another object type?" If yes, the current options in SplitOrMergeObjects go away and instead you get the WIP module's options. I think it would be too messy to allow users, in a single module, to combine two different object types AND ALSO merge objects within a certain distance etc. The user can always just run two copies of the SplitOrMergeObjects module if they want to do that.

(Now, note our overall strategy in module complexity in CellProfiler: our user base appreciates being able to scan all the modules and selecting among them. Instead of having a separate module for every little function in CellProfiler, we prefer to have modules with fairly broad functionality so the user isn't overwhelmed with tons of options. We have ~15 object modules and ~25 image processing modules currently. We can argue about exactly where the line is drawn, though, so it's up for debate whether the WIP module should be a separate one. To me it seems to confusing to have a module called SplitOrMergeObjects and a separate one called CombineObjects, hence my recommendation to combine the functionality in a single module, SplitOrMergeObjects.)

0x00B1:
@AnneCarpenter and @braymp Has a module been renamed in the past? I don’t understand how this would work without breaking backwards-compatability outside of retaining both names inside of CellProfiler since CellProfiler identifies modules by a string value (e.g. “crop”) rather than a unique and immutable identifier (e.g. a random integer).

AnneCarpenter:
Indeed, we renamed modules a lot, and Lee patiently backwards compatibilized them for us. But I don't know how.

0x00B1:
> Indeed, we renamed modules a lot, and Lee patiently backwards compatibilized them for us. But I don't know how.

I just looked. He kept both identifiers. This is annoying. We should stop using names (i.e. strings) for keys.

AnneCarpenter:
Fine by me  :D

0x00B1:
I don’t think this should include the “by percentage” feature. It’s too annoying to implement when nobody has an immediate use.

> (Now, note our overall strategy in module complexity in CellProfiler: our user base appreciates being able to scan all the modules and selecting among them. Instead of having a separate module for every little function in CellProfiler, we prefer to have modules with fairly broad functionality so the user isn't overwhelmed with tons of options. We have ~15 object modules and ~25 image processing modules currently. We can argue about exactly where the line is drawn, though, so it's up for debate whether the WIP module should be a separate one. To me it seems to confusing to have a module called SplitOrMergeObjects and a separate one called CombineObjects, hence my recommendation to combine the functionality in a single module, SplitOrMergeObjects.)

I disagree. 😝

0x00B1:
My proposed name: “Join Segmentation.”

AnneCarpenter:
Personally, I suspect the functionality in the module is nearly useless without the percentage function; else, it will only work for very sparse, nicely spaced objects which are a rarity for us. Others should chime in on that, though.

Yeah, we could devote a whole coding day to debating, as Lee once put it, "unfortunate tension between having a giant jumble of very clean, single-purpose modules and having a small number of nicely categorized modules, each of which is a giant jumble of sometimes loosely-connected functions."

I think both options are sane choices. But we've chosen the latter (largely because users value being able to skim a small number of nicely categorized modules).

braymp:
@0x00B1: Similar case in point re: percentage: I often use MaskObjects over RelateObjects for precisely this reason. RelateObjs will parent an object if there is only pixel of overlap, which is almost always not what you want. 

I can easily imagine an instance where this new module merges two barely touching objects, and that result being the wrong answer.

karhohs:
Merge objects could add additional options to the unify operation of merging objects. In addition to merging based upon distance, there would be the merging based on size. Objects below a size threshold would be merged together until these new objects were above the threshold. There could also be a "gravity" effect, where if a small object is near two larger objects, the larger objects would exert a kind of "pull" to decide which object the smaller object should merge to. The pulling force would be a function of size and distance.

mcquin:
Finishing this is not currently a priority.

braymp:
I'm slightly sad about this. :frowning:

mcquin:
@braymp the issue is still open :)

leeleavitt:
I want this attribute so bad.  Even a simple add objects would be nice.

braymp:
@leeleavitt: Could you use IdentifyTeritiary for what you describe? I believe it works even if the larger/smaller objects are not related via identifySecondary.
And yes, I'd love to see this attribute as well :smile:

PR #2160, opened 2016-07-21
exit code never assigned

Swarchal:
As mentioned on the forums [here](http://forum.cellprofiler.org/t/converting-example-pipeline-to-run-headless/3575/6?u=sjwarch), if `options.done_file` **is** `None`, then the exit code is never assigned.

Not sure why this is happening in our case, but it means the processes are not exiting on our cluster.


0x00B1:
Awesome! Thanks, @Swarchal!

Issue #1939, opened 2016-03-30
Remove “Temporary folder” option


0x00b1:
CellProfiler should use a platform’s default temporary directory:

**Linux and OS X**

`$TMPDIR`

**Windows**

`%TEMP%`

LeeKamentsky:
It uses the OS mechanisms for finding the temporary file location as the default, but we've had many requests to override this behavior, hence the configuration option.

0x00B1:
> It uses the OS mechanisms for finding the temporary file location as the default, but we've had many requests to override this behavior, hence the configuration option.

If you’re using the appropriate environment variables (i.e. $TMPDIR on Linux and OS X and %TMP% or %TEMP% on Windows), then it’s already configurable. My motivation is not breaking existing platform conventions.

braymp:
A bit worried about this, since the system temp folder may not be equipped to handle the storage needs for a CP run. 
For instance: on the VMs, the temp folder was limited due to low HD space, and CP would regularly crash as a result. The solution was to have the CP temp folder point to shared high-performance space. I would not want my OS temp folder to point to that same location, though.

0x00B1:
@braymp:

> A bit worried about this, since the system temp folder may not be equipped to handle the storage needs for a CP run. 

It should still configurable, but it should be configurable by the platform (i.e. an operating system) rather than the application. An example on a POSIX-platform (e.g. Linux or OS X):

```sh
TMPDIR=./foo/bar cellprofiler
```

LeeKamentsky:
I looked for the associated issue for this, could not find it. Someone asked for it, not my idea.

dlogan:
This seems a valid reason for keeping this option, at least as a command-line switch: Cluster users do not always have permissions to change this location: 
http://forum.cellprofiler.org/t/unable-to-change-temporary-directory-on-cluster/2544

PR #2192, opened 2016-08-10
put back -t/--temporary directory option.

thouis:
reverts 059357f92c8bffb79f.


thouis:
I'm not sure if this should also revert #1939.

PR #2262, opened 2016-08-26
Template matching (resolves #2261)

### LINKED ISSUE #2261: Template matching

0x00b1:


0x00b1:
templatematching__5__image_cycle__1

braymp:
May I suggest the name "MatchTemplate" to match the current module nomenclature? :smile:

0x00b1:
[q]May I suggest the name "MatchTemplateImage" to match the current module nomenclature? 😄[/q]

Noted! However, I plan on keeping “Template matching.”

0x00b1:
The module works with:
 
- single-channel two-dimensional images
- multi-channel two-dimensional images
- single-channel three-dimensional images

i.e. it works for images with an `(x, y, z)` shape.

cellprofiler/modules/templatematching.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+
+        self.template_name = cellprofiler.setting.Pathname(
+            "Template",
+            ""
+        )
+
+        self.output_image_name = cellprofiler.setting.ImageNameProvider(
+            "Output",
+            "response",
```
mcquin: Should this be left blank?

mcquin:
Looks good! Squash commits, please.

0x00b1:
@mcquin Thanks!

cellprofiler/modules/templatematching.py
0x00b1: I didn’t have any ideas about a sensible default.

0x00b1:
@mcquin Squashed!

cellprofiler/modules/templatematching.py
0x00b1: Fixed!

mcquin:
Some documentation for this module would be nice. :)

tests/modules/test_templatematching.py
```diff
@@ -0,0 +1,57 @@
...
+
+        image_set.add("image", image)
+
+        module.input_image_name.value = "image"
+
+        module.output_image_name.value = "output"
+
+        module.template_name.value = os.path.join(os.path.dirname(__file__), "../resources/template.png")
+
+        module.run(workspace)
```
mcquin: Can we test the result of this module? For example, does the template match?

What is the result when there is no match?

AnneCarpenter:
Sorry I'm not sure how to reply to individual chunks of this thread, so here goes:
- I bet @bethac07 would be happy to do the user-facing documentation, given she requested this module?
- Do all new modules have a standard battery of tests like what happens when you try to run a blank image through, a color image if it's expecting grayscale, etc. or do you have to think about those manually for each new one?
- I'm with Mark on voting for the name to be MatchTemplate. It would be inconsistent to have all the other modules start with verbs and not this one, and skimming a list of actions is easier than a mixture of tenses.

0x00b1:
> I bet @bethac07 would be happy to do the user-facing documentation, given she requested this module?

Too late!

> Do all new modules have a standard battery of tests like what happens when you try to run a blank image through, a color image if it's expecting grayscale, etc. or do you have to think about those manually for each new one?

If a test is repeated from module to module, it should be elsewhere in a generic module. A module test should test functionality specific to the module.

> I'm with Mark on voting for the name to be MatchTemplate. It would be inconsistent to have all the other modules start with verbs and not this one, and skimming a list of actions is easier than a mixture of tenses.

Noted! We’re keeping “TemplateMatching!” Nonetheless, thanks for your feedback.

tests/modules/test_templatematching.py
0x00b1: 😞

Of course. I apologize for my laziness. LOL

0x00b1: > What is the result when there is no match?

The pixel value is 0.

braymp:
> Noted! We’re keeping “TemplateMatching!” Nonetheless, thanks for your feedback.

Pretty please?

image

AnneCarpenter:
Hmmm... who is this "we" in "We're keeping it"? Does this mean @mcquin signed off on ignoring internal consistency *and* the user audience just because some random person chose a noun form of the phrase to start a wikipedia page? Would it help if I brought cookies & milk to the office? So many questions.

0x00b1:
> Hmmm... who is this "we" in "We're keeping it"? Does this mean @mcquin signed off on ignoring internal consistency and the user audience just because some random person chose a noun form of the phrase to start a wikipedia page? Would it help if I brought cookies & milk to the office? So many questions.

Is this worth a flame war?

0x00b1:
This is temporary directory flag part two.

AnneCarpenter:
😢

0x00b1:
super annoyed

bethac07:
Playing with this now; one thing that jumped out at me quickly is that for rotationally-asymmetrical templates (like a line) we're going to run into issues.  I googled a bit and it seems to SIGNIFICANTLY complicate life to try to do a true rotationally invariant template match.  

As a proof of concept though, I took the line and created 4 rotations of my template (horizontal, diagonal up, diagonal down, vertical), ran MatchTemplate 4 times, then took a max projection of the 4 outputs in ImageMath.  After applying a reasonable threshold, it did a not-terrible job at pulling out the features of interest with only 4 angles- see below.  Given that MatchTemplate seems to be very fast, I wonder if this is something we could do under the hood while still keeping performance time reasonable- allow the user to call it up to some arbitrary number of times (say 12), rotate the pixel data of the template 180/n degrees, and then call match_template and numpy.maximum() the output array vs the previous maximum at each iteration.  

Would you be terribly opposed to me trying to make those changes?  I think it's straightforward enough I could probably do it on my own.  Given that our major use cases for this function were all linear, I think it may be necessary in order to make it actually useful.

rotation_image

0x00b1:
@bethac07 Interesting idea! But you should use the common method of a local feature detector and descriptor, e.g. [scale-invariant feature transform](https://en.wikipedia.org/wiki/Scale-invariant_feature_transform) (SIFT) or [speeded up robust features](https://en.wikipedia.org/wiki/Speeded_up_robust_features) (SURF).

0x00b1:
@bethac07 And they are

bethac07:
> Interesting idea! But you should use the common method of a local feature detector and descriptor, e.g. scale-invariant feature transform (SIFT) or speeded up robust features (SURF).

There's no implementation for these in scikit (it can read results files from other SIFT or SURF applications but not create the mappings AFAICT).  There is one in OpenCV apparently.

cellprofiler/modules/matchtemplate.py
```diff
@@ -0,0 +1,97 @@
...
+"""
+
+import cellprofiler.image
+import cellprofiler.module
+import cellprofiler.setting
+import skimage.feature
+import skimage.io
+
+
+class MatchTemplate(cellprofiler.module.Module):
```
mcquin: This class can be updated to inherit from `cellprofiler.module.ImageProcessing`.

cellprofiler/modules/matchtemplate.py
```diff
@@ -0,0 +1,97 @@
...
+
+        template = skimage.io.imread(template_name)
+
+        output_pixels = skimage.feature.match_template(
+            image=input_pixels,
+            template=template,
+            pad_input=True
+        )
+
+        output_image = cellprofiler.image.Image(output_pixels, parent_image=input_image)
```
mcquin: Does it work in 3D?

bethac07:
Note from 20161118- use RANSAC (I'm trying to add it on my fork but something very odd is happening)

Issue #2197, opened 2016-08-12
thread safe cellprofiler.*


0x00b1:

mcquin:
#2271

mcquin:
@0x00b1 is this related to #2204 ?

AnneCarpenter:
Once this happens Allen says we can "just annotate each skimage call with http://scikit-image.org/docs/dev/api/skimage.util.html#apply-parallel".

0x00b1:
way, way better

PR #2271, opened 2016-08-31
Thread safety, etc.

0x00b1:
@shntnu was speaking at a conference and during the Q&A, a person stood up and began their question, “As you know, CellProfiler is slow …” 

Yes, we know.

CellProfiler’s poor performance is beyond debate. 

CellProfiler’s poor performance has entered “known problem” territory.

Why is CellProfiler slow? 

Most assume CellProfiler is slow because image processing is computationally expensive. It’s true. But rest assured, few write faster implementations than Lee Kamentsky. If an  operation is `O(MN)`, our implementation is predictably `O(MN)`. 

(If you find an exception, it’s a bug! Please file an issue on the CellProfiler repository or the scikit-image repository!) 

So what’s the problem?

Well … there are many—way too many to enumerate. 

But three performance issues are especially frustrating to CellProfiler users: distributed computing, thread safety, and parallel computing.

While CellProfiler has a mechanism for distributing work (an image and pipeline) to other CellProfiler instances, it makes two odd decisions: it’s only available from the CellProfiler application (i.e. it’s not available from the command line or by calling CellProfiler from your Python application with `import cellprofiler`) and it distributes work locally rather than over the network.

Ideally, distributed computing mechanisms should be generic (e.g. capable of using a variety of message brokers). Likewise, a distributed computing use shouldn’t be assumed. Presently, when you use the CellProfiler application, running an analysis will create CellProfiler workers and distribute work. Practically, this significantly degrades performance for pipelines that process less than a few hundred images.

When inter-operation is important, lack of thread safety wrecks havoc on performance while creating numerous obscure issues that are difficult to debug.

Ideally, applications are single threaded and parallelization is achieved with processes (or processors) rather than threads (physical and software). However, some features require multi-threading. For CellProfiler, multi-threading was introduced with the CellProfiler application since the GUI runs inside an event loop. When this is considered alongside CellProfiler’s other threading needs (e.g. distributing and processing work), the multi-threading situation becomes tricky and performance suffers. This is especially befuddling for CellProfiler contributors due to the lack of a clear separation between CellProfiler’s logic (e.g. it’s core data structures or modules) and its GUI. For example, thread safety isn’t guaranteed when running CellProfiler from the command line.

Finally, CellProfiler’s logic suggests there are a number of areas that are embarrassingly parallel. For example, CellProfiler can be simplified into the following loop that’s embarrassingly parallel:

``` Python
for image in images:
    image = process(image)

    measurement = analyze(image)
```

Unfortunately, the previous two issues make this loop difficult to parallelize in practice.

However, I’m uncertain if this type of loop is a good candidate for parallelization. It seems like it’d make far more sense to use a parallel implementation of CellProfiler’s processing and analysis algorithms (e.g. thresholding or watershed) and leave the bigger problems (e.g. the aforementioned loop) for a distributed computing solution. It’d also buy us the ability to perform out-of-core computation (i.e. processing images that exceed available memory).


mcquin:
Does this make CP single-threaded?

0x00b1:
@shntnu was speaking at a conference and during the Q&A, a person stood up and began their question, “As you know, CellProfiler is slow …” 

Yes, we know.

CellProfiler’s poor performance is beyond debate. 

CellProfiler’s poor performance has entered “known problem” territory.

Why is CellProfiler slow? 

Most assume CellProfiler is slow because image processing is computationally expensive. It’s true. But rest assured, few write faster implementations than Lee Kamentsky. If an  operation is `O(MN)`, our implementation is predictably `O(MN)`. 

(If you find an exception, it’s a bug! Please file an issue on the CellProfiler repository or the scikit-image repository!) 

So what’s the problem?

Well … there are many—way too many to enumerate. 

But three performance issues are especially frustrating to CellProfiler users: distributed computing, thread safety, and parallel computing.

While CellProfiler has a mechanism for distributing work (an image and pipeline) to other CellProfiler instances, it makes two odd decisions: it’s only available from the CellProfiler application (i.e. it’s not available from the command line or by calling CellProfiler from your Python application with `import cellprofiler`) and it distributes work locally rather than over the network.

Ideally, distributed computing mechanisms should be generic (e.g. capable of using a variety of message brokers). Likewise, a distributed computing use shouldn’t be assumed. Presently, when you use the CellProfiler application, running an analysis will create CellProfiler workers and distribute work. Practically, this significantly degrades performance for pipelines that process less than a few hundred images.

When inter-operation is important, lack of thread safety wrecks havoc on performance while creating numerous obscure issues that are difficult to debug.

Ideally, applications are single threaded and parallelization is achieved with processes (or processors) rather than threads (physical and software). However, some features require multi-threading. For CellProfiler, multi-threading was introduced with the CellProfiler application since the GUI runs inside an event loop. When this is considered alongside CellProfiler’s other threading needs (e.g. distributing and processing work), the multi-threading situation becomes tricky and performance suffers. This is especially befuddling for CellProfiler contributors due to the lack of a clear separation between CellProfiler’s logic (e.g. it’s core data structures or modules) and its GUI. For example, thread safety isn’t guaranteed when running CellProfiler from the command line.

Finally, CellProfiler’s logic suggests there are a number of areas that are embarrassingly parallel. For example, CellProfiler can be simplified into the following loop that’s embarrassingly parallel:

```Python
for image in images:
    image = process(image)

    measurement = analyze(image)
```

Unfortunately, the previous two issues make this loop difficult to parallelize in practice.

However, I’m uncertain if this type of loop is a good candidate for parallelization. It seems like it’d make far more sense to use a parallel implementation of CellProfiler’s processing and analysis algorithms (e.g. thresholding or watershed) and leave the bigger problems (e.g. the aforementioned loop) for a distributed computing solution. It’d also buy us the ability to perform out-of-core computation (i.e. processing images that exceed available memory).

0x00b1:
@mcquin I updated the issue with some context.

0x00b1:
> Does this make CP single-threaded?

No, but it solves part of the problem. I still need to resolve GUI problems.

0x00b1:
@mcquin note the passing tests. 😎

0x00b1:
@AnneCarpenter @bethac07 @braymp @karhohs It’d be especially helpful if you could spend some time with this branch since it’ll likely have some issues that wouldn’t be caught by our tests.

0x00b1:
There’s an issue where the progress bar doesn’t stop when an analysis has completed. Likewise, the progress bar doesn’t increment properly. 

banners_and_alerts_and_cellprofiler_2_2_0__rev_bf6aac9__and_thread_safety__etc__by_0x00b1_ _pull_request__2271_ _cellprofiler_cellprofiler

braymp:
Cool! @0x00b1, I've just cloned this branch and I plan on kicking the tires a bit....

braymp:
OK, I ran into the same progress bar issue.

Also, I should ask: What sort of problems/things should I be on the lookout for (if any)?

AnneCarpenter:
Naive question - this isn't the nightly build, right? I'd be happy to help if there is a compiled version handy.

AnneCarpenter:
Not sure if comments go here or if I should file a new issue... 

So this branch is pretty cool, seems very fast, and it's the first time I've gotten to see all the clean new UI improvements. SOOOO nice. Breath of fresh air!

Things seemed to work well running the example pipeline for me. The only additional issue I found was funny checkbox behavior (the ones that activate/deactivate a particular module). I checked them all kinds of ways: They work fine on master and they work fine on features/joblib when first launched. But after running an analysis, the checkboxes do not check/uncheck anymore.  By contrast, the eyeballs to show/hide windows behave properly before and after running an analysis.

But that's all I found. I even checked a pipeline with TrackObjects because that has a funky run-after-all-image-sets behavior and it seemed to do its thing properly.

I did a speed test whereby I removed the Crop modules (to make images bigger/slower) and I removed SaveImages so it wouldn't pause to ask me about overwriting. I timed it using a stopwatch app, stopping when the last of the .csv's is written (since it's hard to tell when the new branch is finished given the progress bar wonkiness). Here are the results:
master: 1m15sec
features/joblib: 41sec

Nice!

0x00b1:
Travis is failing because of a Dask issue.

mcquin:
https://github.com/CellProfiler/CellProfiler/issues/2197

0x00b1:
Outdated :(

braymp:
@0x00b1: Curious as to what you mean by outdated: "won't fix" or "can't fix"?

0x00b1:
Moved into different pull requests and commits. This was an exploratory PR.

Issue #2193, opened 2016-08-10
MeasureObjectIntensityDistribution calculates intensity zernikes even when not asked to.


bethac07:
Running an analysis pipeline on AWS on the stable 2.2.0 I got the following error:
```
Wed Aug 10 18:48:16 2016: Image # 19, module MeasureObjectNeighbors # 34: 3.55 sec
Error detected during run of module MeasureObjectIntensityDistribution
Traceback (most recent call last):
  File "/usr/local/src/CellProfiler/cellprofiler/pipeline.py", line 1819, in run_with_yield
    self.run_module(module, workspace)
  File "/usr/local/src/CellProfiler/cellprofiler/pipeline.py", line 2067, in run_module
    module.run(workspace)
  File "/usr/local/src/CellProfiler/cellprofiler/modules/measureobjectintensitydistribution.py", line 438, in run
    self.calculate_zernikes(workspace)
  File "/usr/local/src/CellProfiler/cellprofiler/modules/measureobjectintensitydistribution.py", line 786, in calculate_zernikes
    z_ = z[mask, :]
MemoryError
Wed Aug 10 18:48:19 2016: Image # 19, module MeasureObjectIntensityDistribution # 35: 63.51 sec
```

I can't post the full pipeline in this case but the relevant module settings from the cppipe are below:
```
MeasureObjectIntensityDistribution:[module_num:35|svn_version:\'Unknown\'|variable_revision_number:5|show_window:False|notes:\x5B\x5D|batch_state:array(\x5B\x5D, dtype=uint8)|enabled:True|wants_pause:False]
    Hidden:2
    Hidden:3
    Hidden:1
    Hidden:0
    Calculate intensity Zernikes?:None
    Maximum zernike moment:9
    Select an image to measure:Channel2
    Select an image to measure:Channel3
    Select objects to measure:Cells
    Object to use as center?:These objects
    Select objects to use as centers:None
    Select objects to measure:Cytoplasm
    Object to use as center?:These objects
    Select objects to use as centers:None
    Select objects to measure:Nuclei
    Object to use as center?:These objects
    Select objects to use as centers:None
    Scale the bins?:Yes
    Number of bins:4
    Maximum radius:100
```

My pipeline is crashing due to memory errors on seemingly 75+% of the images so I'm highly suspicious this is why.

tomgreen66:
This looks very similar to a problem we are experiencing but we got a bit further possibly:
```
[...]

PR #2296, opened 2016-09-08
Make sure we actually check the value of the option.

tomgreen66:
#2193 highlighted a problem with calculating Zernikes even when not asked to.  It seems the check if Zernikes are wanted before calculating them does not check the value of the option but falls into Pythons default `True` if a class has no `__nonzero__` or `__len__` defined.

Reran some jobs which were failing and it now works when you have the Zernike option turned off.


mcquin:
Nice! :+1:

PR #2299, opened 2016-09-09
Update __init__.py

AnneCarpenter:


0x00b1:
@AnneCarpenter Thanks!

PR #2329, opened 2016-09-16
Thresholding

### LINKED ISSUE #2177: Thresholding

mcquin:
Resolves #2177


cellprofiler/modules/thresholding.py
```diff
@@ -0,0 +1,231 @@
...
+import cellprofiler.module
+import cellprofiler.setting
+import numpy
+import skimage.exposure
+import skimage.filters
+import skimage.morphology
+
+
+class Thresholding(cellprofiler.module.Module):
+    category = "Volumetric"
```
0x00b1: `category` should be “Image processing.”

cellprofiler/modules/thresholding.py
```diff
@@ -0,0 +1,231 @@
...
+    variable_revision_number = 1
+
+    def create_settings(self):
+        self.x_name = cellprofiler.setting.ImageNameSubscriber(
+            u"Input"
+        )
+
+        self.y_name = cellprofiler.setting.ImageNameProvider(
+            u"Output",
+            u"OutputImage"
```
0x00b1: “Thresholding?”

cellprofiler/modules/thresholding.py
```diff
@@ -0,0 +1,231 @@
...
+        )
+
+        self.method = cellprofiler.setting.Choice(
+            u"Method",
```
0x00b1: Should we use “operation” instead of “method?”

Either way, we should standardize our terminology.

cellprofiler/modules/thresholding.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+class Thresholding(cellprofiler.module.Module):
+    category = "Volumetric"
+    module_name = "Thresholding"
+    variable_revision_number = 1
+
+    def create_settings(self):
+        self.x_name = cellprofiler.setting.ImageNameSubscriber(
+            u"Input"
```
0x00b1: I appreciate your use of Unicode (i.e. the prefixed `u`).

I should do this too to simplify upcoming Python 3 work.

cellprofiler/modules/thresholding.py
```diff
@@ -0,0 +1,231 @@
...
+            ],
+            u"Gaussian"
+        )
+
+        self.adaptive_offset = cellprofiler.setting.Float(
+            u"Offset",
+            value=0.0
+        )
+
+        self.adaptive_mode = cellprofiler.setting.Choice(
```
0x00b1: You might consider dropping this. I’ve never modified it.

cellprofiler/modules/thresholding.py
```diff
@@ -0,0 +1,231 @@
...
+                u"Constant",
+                u"Mirror",
+                u"Nearest",
+                u"Reflect",
+                u"Wrap"
+            ],
+            u"Reflect"
+        )
+
+        self.isodata_bins = cellprofiler.setting.Integer(
```
0x00b1: You should consolidate `isodata_bins`, `otsu_bins`, and `yen_bins` into one `bins` setting.

cellprofiler/modules/thresholding.py
```diff
@@ -0,0 +1,231 @@
+# coding=utf-8
+
+"""
+
+Thresholding
```
0x00b1: Documentation?

0x00b1: Test?

0x00b1: Test?

cellprofiler/modules/thresholding.py
```diff
@@ -0,0 +1,231 @@
...
+
+"""
+
+import cellprofiler.image
+import cellprofiler.module
+import cellprofiler.setting
+import numpy
+import skimage.exposure
+import skimage.filters
+import skimage.morphology
```
0x00b1: Unused

cellprofiler/modules/thresholding.py
```diff
@@ -0,0 +1,231 @@
...
+
+Thresholding
+
+"""
+
+import cellprofiler.image
+import cellprofiler.module
+import cellprofiler.setting
+import numpy
+import skimage.exposure
```
0x00b1: Unused

cellprofiler/modules/thresholding.py
```diff
@@ -0,0 +1,231 @@
...
+            [
+                u"Gaussian",
+                u"Generic",
+                u"Mean",
+                u"Median"
+            ],
+            u"Gaussian"
+        )
+
+        self.adaptive_offset = cellprofiler.setting.Float(
```
0x00b1: You should use the generic name `offset` rather than `adaptive_offset`.

cellprofiler/modules/thresholding.py
```diff
@@ -0,0 +1,231 @@
...
+                u"Adaptive",
+                u"Iterative selection thresholding",
+                u"Manual",
+                u"Minimum cross entropy thresholding",
+                u"Otsu’s method",
+                u"Yen’s method"
+            ]
+        )
+
+        self.adaptive_block_size = cellprofiler.setting.Integer(
```
0x00b1: You should use the generic name `block_size` rather than `adaptive_block_size`.

cellprofiler/modules/thresholding.py
```diff
@@ -0,0 +1,231 @@
...
+
+        y_name = self.y_name.value
+
+        images = workspace.image_set
+
+        x = images.get_image(x_name)
+
+        x_data = x.pixel_data
+
+        x_data = skimage.img_as_uint(x_data)
```
0x00b1: Is this necessary?

cellprofiler/modules/thresholding.py
mcquin: How about "Binary"?

cellprofiler/modules/thresholding.py
mcquin: I like "operation".

cellprofiler/modules/thresholding.py
mcquin: It looks like it's used: https://github.com/CellProfiler/CellProfiler/blob/aae5d0873b7b7c0b607ae9643da63cbcc2fd94f8/cellprofiler/modules/thresholding.py#L174-L180

cellprofiler/modules/thresholding.py
0x00b1: I like “Thresholding” because it’s consistent across image processing modules (i.e. the default output image name can be the name of the module).

cellprofiler/modules/thresholding.py
0x00b1: Me too 😎

cellprofiler/modules/thresholding.py
0x00b1: Yeah, remove the argument from the call and use the default (“reflect”).

0x00b1:
We also need to depreciate ApplyThreshold.

(And I suppose the depreciated patch should be included in this pull request.)

0x00b1:
It’d also mean abandoning some operations, e.g. maximum entropy (“Kapur”).

But It’s 2016 and we understand threshold applications.

Likewise, we’re not ImageJ and don’t make guarantees about completeness of operations.

We’re limiting options to operations with strong biology use and I think that’s a win.

AnneCarpenter:
Regarding this comment: 
`It’d also mean abandoning some operations, e.g. maximum entropy (“Kapur”). But It’s 2016 and we understand threshold applications. Likewise, we’re not ImageJ and don’t make guarantees about completeness of operations. We’re limiting options to operations with strong biology use and I think that’s a win.`

Totally agreed about not promising completeness and it being okay to lose a few methods. 

Is there any sensible way to assess whether people use a particular method often, other than to remove it and see whether anyone complains? This is one of those moments I wish stealth-CellProfiler told us all the pipelines people run. 

I suppose at the least we could ask at the forum whether people want to passionately defend their thresholding method of choice. I suspect it's unlikely all existing methods would get strong support but at least we'd hear what people use often and/or care about to help inform the decision. If you think it makes sense to do that, I will post.

0x00b1:
> Is there any sensible way to assess whether people use a particular method often, other than to remove it and see whether anyone complains? This is one of those moments I wish stealth-CellProfiler told us all the pipelines people run.

I wish!

However, I don’t know if existing pipelines would be useful since I bet most people modify the default (i.e. they modify the method used by `IdentifyPrimaryObjects`).

It’s difficult to benchmark since the effectiveness of thresholding is dependent on the application (i.e. “optimal threshold” is an empiric rather than analytic claim) so any benchmark needs to be fitted for a specific problem.

We could review papers. We could also ask around.

Regardless, thresholding is a (necessary) nightmare. It destroys useful information. Thankfully, it’ll be abandoned for non-threshold boundaries in our deep learning future. 😜

AnneCarpenter:
All right, I've asked around, here. At least it is some source of data even if only Beth and Kyle reply: http://forum.cellprofiler.org/t/whats-your-favorite-thresholding-method-in-cellprofiler/3982

bethac07:
Provided some data.

AnneCarpenter:
Some good discussion has accumulated on the forum, and furthermore it's heartwarming to see that people notice and care enough to give input. So far sounds like Allen's instinct is correct that a few methods can be dropped without causing horrendous pain to anyone.

mcquin:
@0x00b1 this module is missing documentation, and tests. Also, should deprecation of `ApplyThreshold` been addressed, as suggested?

PR #2332, opened 2016-09-16
Noise reduction

### LINKED ISSUE #2172: Noise reduction
- [ ] Non-local means

mcquin:
Resolves #2172


cellprofiler/modules/noisereduction.py
```diff
@@ -0,0 +1,107 @@
...
+        x_data = x.pixel_data
+
+        x_data = skimage.img_as_uint(x_data)
+
+        y_data = skimage.restoration.denoise_nl_means(
+            image=x_data,
+            patch_size=size,
+            patch_distance=distance,
+            h=cutoff_distance,
+            multichannel=False,
```
0x00b1: You’ll need to check whether the image is multi-channel and use the appropriate value.

0x00b1: Likewise, the module should return an error if the user selects “non-local means” and doesn’t provide either a single-or-multi-channel two-dimensional or single-channel three-dimensional image.

cellprofiler/modules/noisereduction.py
```diff
@@ -0,0 +1,107 @@
...
+"""
+
+import cellprofiler.image
+import cellprofiler.module
+import cellprofiler.setting
+import skimage.restoration
+import skimage.util
+
+
+class NoiseReduction(cellprofiler.module.Module):
```
0x00b1: Should we add support for [total variation denoising](https://en.wikipedia.org/wiki/Total_variation_denoising)?

It’s not non-local, but it’s faster.

cellprofiler/modules/noisereduction.py
```diff
@@ -0,0 +1,107 @@
+"""
+
+Noise reduction
```
0x00b1: Documentation?

0x00b1: Test?

cellprofiler/modules/noisereduction.py
```diff
@@ -0,0 +1,107 @@
...
+
+        cutoff_distance = self.cutoff_distance.value
+
+        images = workspace.image_set
+
+        x = images.get_image(x_name)
+
+        x_data = x.pixel_data
+
+        x_data = skimage.img_as_uint(x_data)
```
0x00b1: Is this necessary?

PR #2344, opened 2016-09-19
Medial axis

0x00b1:


cellprofiler/modules/medialaxis.py
```diff
@@ -0,0 +1,89 @@
+"""
+
+<strong>Medial axis<strong>
```
mcquin: Documentation.

0x00b1:
@mcquin This is ready for review! 😎

PR #2368, opened 2016-09-26
Re-expose option for alternate colormaps

### LINKED ISSUE #2367: No ability to select color maps with 8 bit tif images
As of 2.1.1 if you were writing an 8 bit tif you could select the color map to use when writing- the default was (and is) greyscale but it was sometimes nice to be able to write to other color maps for display purposes.  The code looks to me like it still supports this (at least, it checks for the prescribed color map), but the ability to pull it up in visible_settings seems to have been a casualty of the refactoring from bit depth=8 to bit depth=8-bit integer.   Can someone confirm to me whether this was intentionally deprecated or whether this was a mistake?  If it's a mistake I'll put in a PR to fix it.

bethac07:
Resolves #2367


0x00b1:
Weird bug.

PR #2381, opened 2016-10-03
Overlay labels on image in segmentation display

### LINKED ISSUE #2380: display segmentations in color

mcquin:
Resolves #2380 

Not a great segmentation, but this gives an idea of the output:

screen shot 2016-10-03 at 9 54 14 am

### DIFF
--- a/cellprofiler/module.py
+++ b/cellprofiler/module.py
@@ -10,6 +10,7 @@ import cellprofiler.measurement
 import cellprofiler.object
 import cellprofiler.setting as cps
 import pipeline as cpp
+import skimage.color
 
 
 class Module(object):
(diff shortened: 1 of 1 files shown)


mcquin:
@bethac07 has spoken!

AnneCarpenter:
Happy to see colored segmentations (or as my littlest recently called them -"oooh!  Jelly beans!")  

A question: is the colormap selectable for the label matrices (in general but also for the 3D case)? I can't recall if our recent discussion related to colormaps for saving grayscale images or if it also applied to label matrices. Just curious.

mcquin:
LOL! @bethac07 thought these looked like Easter eggs. But I like jelly beans, too.

The color map is not selectable for label matrices on display. I have not looked into saving, yet.

AnneCarpenter:
Definitely not a request, I doubt anyone NEEDS it, it's just something that had the potential to bring joy. 😊

PR #2384, opened 2016-10-04
Volumetric support for ConvertObjectsToImage

### LINKED ISSUE #2382: ConvertObjectsToImage should support volumes.

mcquin:
Resolves #2382

### DIFF
--- a/cellprofiler/image.py
+++ b/cellprofiler/image.py
@@ -433,7 +433,6 @@ class RGBImage(object):
 
 def check_consistency(image, mask):
     """Check that the image, mask and labels arrays have the same shape and that the arrays are of the right dtype"""
-    assert (image is None) or (len(image.shape) in (2, 3)), "Image must have 2 or 3 dimensions"
     assert (mask is None) or (len(mask.shape) == 2), "Mask must have 2 dimensions"
     assert (image is None) or (mask is None) or (image.shape[:2] == mask.shape), "Image and mask sizes don't match"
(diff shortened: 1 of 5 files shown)


mcquin:
@0x00b1 something to review while you're waiting for travis ;)

cellprofiler/image.py
```diff
@@ -433,7 +433,6 @@ def pixel_data(self):
 
 def check_consistency(image, mask):
     """Check that the image, mask and labels arrays have the same shape and that the arrays are of the right dtype"""
-    assert (image is None) or (len(image.shape) in (2, 3)), "Image must have 2 or 3 dimensions"
```
mcquin: We could add an equivalent check based on `dimensionality` and `image.shape`.

tests/modules/test_convertobjectstoimage.py
```diff
@@ -1,180 +1,322 @@
...
+
+    objects = workspace.get_objects("inputobjects")
+
+    assert image.dimensions == objects.segmented.ndim
+
+    pixel_data = image.pixel_data
+
+    assert pixel_data.shape == objects.shape
+
+    if objects.segmented.ndim is 2:
```
mcquin: Casing, like this, in tests makes me uncomfortable. We could potentially skip testing, for example, volumes while making it look like we are. Or, in the future, someone could make a change that would skip test from being run on some subset of inputs. 

I am wondering if we can avoid this casing in the tests by testing images and volumes separately. Test inputs could be 2D `segmented` and `ijv`, and 3D `segmented` (update with `ijv` when supported).

PR #2391, opened 2016-10-12
Random walker algorithm (resolves #2390)

### LINKED ISSUE #2390: Random walker algorithm

0x00b1:


paulmenzel:
Should the commits be squashed? The second one looks like a fix-up commit.

Also some notes about the implementation would be nice.

0x00b1:
@paulmenzel GitHub added a `Squash and merge` button! It rules! I also added some documentation. What do you think?

@mcquin This is ready for review!

paulmenzel:
@0x00b1, I don’t like the *Squash and merge*, as there are often good reasons, for having a branch consist of more commits, to make it better digestible.

I also don’t know, if it will work, when you have merge the master branch back into your branch.

Also merging the branches, makes it hard to review now. At least in the file changes, I see `scikit-image` added to the depends. Or is that from one of your commits?

0x00b1:
random_walker_algorithm__6__image_cycle__1

0x00b1:
> @paulmenzel Also merging the branches, makes it hard to review now. At least in the file changes, I see scikit-image added to the depends. Or is that from one of your commits?

This was a mistake. 😳

0x00b1:
> @paulmenzel I don’t like the Squash and merge, as there are often good reasons, for having a branch consist of more commits, to make it better digestible.

I agree. However, I think in this case one commit makes sense.

paulmenzel:
Agreed. In this case it probably does.

cellprofiler/modules/randomwalkeralgorithm.py
```diff
@@ -0,0 +1,105 @@
...
+import cellprofiler.module
+import cellprofiler.object
+import cellprofiler.setting
+import numpy
+import skimage.measure
+import skimage.segmentation
+
+
+class RandomWalkerAlgorithm(cellprofiler.module.ImageSegmentation):
+    module_name = "Random walker algorithm"
```
mcquin: For visual consistency, this should be "RandomWalkerAlgorithm".

cellprofiler/modules/randomwalkeralgorithm.py
```diff
@@ -0,0 +1,105 @@
+# coding=utf-8
+
+"""
+
+Random walker algorithm
+
+Single-channel images can be two-or-three-dimensional and multichannel images can be three-or-four-dimensional.
```
mcquin: I think it's nice to provide a description of what this module does.

PR #2398, opened 2016-10-24
BlobDetection

### LINKED ISSUE #2306: Detect blobs in a volume

mcquin:
Resolves #2306


mcquin:
Will rebase before merge.

0x00b1:
Would you mind posting a screenshot?

cellprofiler/modules/blobdetection.py
```diff
@@ -0,0 +1,379 @@
+# coding=utf-8
+
+"""
+
+Find blobs in a grayscale (single-channel) image or volume. Blobs are bright on dark or dark on bright regions in an
```
0x00b1: Are you saying that it’s limited to single-channel two-dimensional and single-channel multi-dimensional images? Is there a reason this cannot be extended to multichannel two-dimensional and multichannel three-dimensional images? Is it possible to do so by pre-processing with `rgb2gray`?

mcquin:
@0x00b1 here's one https://github.com/CellProfiler/CellProfiler/issues/2306

cellprofiler/modules/blobdetection.py
```diff
@@ -0,0 +1,379 @@
...
+image. The output of this module is a binary image of white circles or spheres centered around found blobs on a black
```
0x00b1: Nice. 😎

I assumed it’d return boundaries, but filled circles are probably more useful.

cellprofiler/modules/blobdetection.py
```diff
@@ -0,0 +1,379 @@
...
+        ]
+
+        if self.operation.value in [u"Determinant of the Hessian (DoH)", u"Laplacian of Gaussian (LoG)"]:
+            settings = settings + [
+                self.scale
+            ]
+
+        return settings
+
+    def on_setting_changed(self, setting, pipeline):
```
0x00b1: What is this?

0x00b1:
Awesome work

cellprofiler/modules/blobdetection.py
mcquin: I thought about using `rgb2gray`, but that assumes all multi-channel images are RGB images.

cellprofiler/modules/blobdetection.py
mcquin: Each operation's `skimage.feature` equivalent has slightly different default parameters. This has the benefit of changing the settings to match those default parameters when the operation is changed. The downside is that it forgets/wipes out user defined settings when the operation is changed.

mcquin:
Squashed.

PR #2409, opened 2016-10-28
Improve watershed segmentation.

### LINKED ISSUE #2405: Watershed over-segmentation problems
Presently, the Watershed module suffers from over-segmentation issues when segmenting typical CellProfiler images (e.g. DAPI). I think this can be improved by removing the `footprint` argument from `peak_local_max` and adding a `min_distance` argument to `peak_local_max`. The `min_distance` should be a configurable setting. While it won’t fix the issue entirely, it’ll be a major improvement over the existing implementation.

mcquin:
Add option "Size" specifying minimum size of objects to segment in image.

Resolves #2405.


AnneCarpenter:
If I read the code correctly, the user just sees "size" as the prompt for the setting, is that right?

If so, does it make sense to change "size" to be something more specific? I could imagine someone thinking it means area instead of diameter (at least, I assume it's the latter?) So I'm thinking this should be something like "Typical minimum diameter of objects"? 

I see the documentation IS more detailed but I'm thinking we want the prompt to be itself relatively complete for the majority of readers, with the documentation providing further detail for a small subset of nerds/experts/confused people.

0x00b1:
@AnneCarpenter Good question. It isn’t size, right? It’s the minimum number of pixels between peaks. 

Ideally, each peak corresponds to an object (e.g. a cell nucleus), so it’d be “minimum distance between objects.” And even that isn’t exactly correct since the formula was normalized to be “minimum half-way distance between objects.” And there’s the “+ 1” so it’s greater than or equal to rather than …

Nonetheless, I agree, there’s a better description for this setting. Unfortunately, I don’t have any ideas. 😝

Issue #1156, opened 2014-07-01
IdentifyPrimaryObjects Laplacian of Gaussian artifacts


dlogan:
Using IdentifyPrimaryObjects with the Laplacian of Gaussian (LoG) can cause odd artifacts in some segmentations.  See the attached screenshot for a couple examples.

To replicate, load the project here (I believe it is self-contained).  The artifacts always seem to be diagonal invaginations which don't necessarily seem to match any intensity dip in the input image.

Attachments:
http://cellprofiler.org/issues/uploaded/8GnTRx/LoG_issue.cpproj
http://cellprofiler.org/issues/uploaded/i-0vm1/ScreenShot2014-07-01at9.38.24AM.png

AnneCarpenter:
Given we have many functional methods for identifying objects, I don't think we need to go to heroics to save LoG or work around its limitations. I was going to go adjust the help for the module but can't find anything about LoG in IdentifyPrimaryObjects' help... where was this setting, @dlogan ? The project file attachments are no longer available so I couldn't check.

braymp:
Similar artifacts occur with the Shape declumping method at times; I think the reasons are related. If I can find an example, I'll open a new issue.

dlogan:
Sorry, just saw this @AnneCarpenter...
To see Laplacian of Gaussian (LoG), load up the Welcome Screen pipeline, and look under IdentifyPrimaryObjects, in the setting "Method to distinguish clumped objects".  
![image](https://cloud.githubusercontent.com/assets/1225595/19322481/40711d18-9087-11e6-8e16-4d385fc53cc1.png)
If you click the "?" next to that setting, there is some Help existing there.

PR #2440, opened 2016-11-04
Update identifyprimaryobjects.py

AnneCarpenter:
Holy smokes, did I do this entire process properly?


AnneCarpenter:
This relates to #1156 which is essentially a "Won't fix"; here, I just warn the user in the help about peculiar behavior.

0x00b1:
Nice

PR #2466, opened 2016-11-15
Volumetric support for MeasureImageQuality

### LINKED ISSUE #2455: MeasureImageQuality should support three-dimensional images

mcquin:
- [ ] calculate radially averaged power spectrum

Resolves #2455


0x00b1:
We can revisit RAPS. 🎤

PR #2477, opened 2016-11-16
Add method to return regionprops-like features from cellprofiler.obje…

### LINKED ISSUE #2450: Add method to return regionprops-like features from cellprofiler.objects.Objects
Adding a method to return `regionprops`-like (e.g. area) features from `cellprofiler.objects.Objects` will simplify feature extraction. While `Objects` has an `areas` method, it would be nice if this was generalized into a method that returned other commonly used features.

0x00b1:
…cts.Objects (resolves #2450)

### DIFF
--- a/cellprofiler/object.py
+++ b/cellprofiler/object.py
@@ -5,6 +5,7 @@ import centrosome.index
 import centrosome.outline
 import numpy
 import scipy.sparse
+import skimage.measure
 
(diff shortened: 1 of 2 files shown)


mcquin:
We're thinking that `regionprops` should be an internal method and the Object class should expose methods like `area` and `perimeter` which may use result of `regionprops`.

PR #2478, opened 2016-11-16
Add dimensions property to cellprofiler.object.Objects (resolves #2449)

### LINKED ISSUE #2449: Add dimensions property to cellprofiler.object.Objects
Adding a dimensions property to `cellprofiler.object.Objects` will simplify choosing whether to extract two-or-three dimensional features.

0x00b1:


cellprofiler/object.py
```diff
@@ -40,6 +40,10 @@ def __init__(self):
         self.__parent_image = None
 
     @property
+    def dimensions(self):
+        return self.__parent_image.dimensions
```
mcquin: What if `self.__parent_image` is `None`?

0x00b1: Fixed. :)

cellprofiler/object.py
```diff
@@ -40,6 +40,15 @@ def __init__(self):
         self.__parent_image = None
 
     @property
+    def dimensions(self):
+        if self.__parent_image:
+            return self.__parent_image.dimensions
+        else:
```
mcquin: `else` isn't super necessary here

mcquin:
@0x00b1 needs rebase

PR #2479, opened 2016-11-16
Add boolean volumetric property to Image class. (resolves #2476)

### LINKED ISSUE #2476: Add boolean volumetric property to Image class.
```python
if image.dimensions is 3:
    ...
else:
    ...
```

becomes

```python
if image.volumetric:
    ...
else:
    ...
```

0x00b1:


cellprofiler/image.py
```diff
@@ -106,6 +106,13 @@ def __init__(self,
     def multichannel(self):
         return True if self.dimensions is 2 and self.pixel_data.ndim is 3 else False
 
+    @property
+    def volumetric(self):
+        if self.dimensions is 3:
+            return True
+        else:
```
mcquin: `else` is unnecessary here.

mcquin:
I think there is also a case for adding `volumetric` to `cellprofiler.object.Objects`.

PR #2481, opened 2016-11-16
Add masked property to cellprofiler.object.Objects (resolves #2480)

### LINKED ISSUE #2480: Add masked property to cellprofiler.object.Objects
A masked property would return the segmented objects masked by the parent image’s mask.

0x00b1:


cellprofiler/object.py
```diff
@@ -40,6 +40,10 @@ def __init__(self):
         self.__parent_image = None
 
     @property
+    def masked(self):
+        return self.segmented & self.parent_image.mask
```
mcquin: Using the `&` operation I get:

```python
data = numpy.ones((5, 5))
mask = numpy.zeros((5, 5), dtype='bool')
data & mask

TypeError: ufunc 'bitwise_and' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

Perhaps `numpy.logical_and` is a better option.

mcquin:
This would be real nice to have on `cellprofiler.image.Image`, too.

PR #2490, opened 2016-11-17
Move cellprofiler.workspace.Workspace methods that import cellprofile…

### LINKED ISSUE #2488: Move cellprofiler.workspace.Workspace methods that import cellprofiler.gui into subclass
- [x] get_module_figure
- [x] create_or_find_figure

0x00b1:
…r.gui into subclass (resolves #2488)

### DIFF
--- a/cellprofiler/workspace.py
+++ b/cellprofiler/workspace.py
@@ -237,12 +237,6 @@ class Workspace(object):
 
     in_background = property(get_in_background, set_in_background)
 
-    def get_module_figure(self, module, image_set_number, parent=None):
-        pass
-
-    def create_or_find_figure(self, title=None, subplots=None, window_name=None):
-        pass
-
     def get_outline_names(self):
         """The names of outlines of objects"""
         return self.__outlines.keys()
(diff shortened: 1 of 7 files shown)


cellprofiler/workspace.py
```diff
@@ -238,66 +238,10 @@ def set_in_background(self, val):
     in_background = property(get_in_background, set_in_background)
 
     def get_module_figure(self, module, image_set_number, parent=None):
```
mcquin: Could this method be removed?

cellprofiler/workspace.py
```diff
@@ -238,66 +238,10 @@ def set_in_background(self, val):
...
-                                        name=window_name,
-                                        title=title)
-
-        if not figure in self.__windows_used:
-            self.__windows_used.append(figure)
-
-        return figure
+        pass
 
     def create_or_find_figure(self, title=None, subplots=None, window_name=None):
```
mcquin: Could this method be removed?

cellprofiler/workspace.py
0x00b1: Good call

PR #2508, opened 2016-12-02
Thresholding should support local thresholding operations (resolves #…

0x00b1:
…2468)


cellprofiler/gui/cpframe.py
```diff
@@ -25,7 +25,6 @@
 import cellprofiler.pipeline
 import cellprofiler.preferences
 import cellprofiler.workspace
-import cellprofiler.utilities.version
```
mcquin: Unnecessary change?

PR #2554, opened 2017-01-10
Add missing measurements for Image, MeasureObjectSizeShape

### LINKED ISSUE #2553: MeasureObjectSizeShape doesn't measure everything it says it measures
The ExportToSpreadSheet module fails to export measurements for discovered objects. This could relate to the expectation of 3D measurements that do not yet exist.

mcquin:
Resolves #2553 

* `Image_Count` is an expected measurement when objects are present.
* MeasureObjectSizeShape does not measure the same features in 2D as in 3D. Defining separate measurements based on whether or not an image is 2D or 3D requires significant refactoring. Set unmeasured features to `NaN`.

### DIFF
--- a/cellprofiler/module.py
+++ b/cellprofiler/module.py
@@ -965,6 +965,15 @@ class ImageProcessing(Module):
 class ImageSegmentation(Module):
     category = "Image Segmentation"
 
+    def add_measurements(self, measurements, labels):
+        n_objects = len(np.unique(labels)[1:]) if 0 in labels else len(np.unique(labels))
+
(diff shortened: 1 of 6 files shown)


AnneCarpenter:
@mcquin I've never heard of Image_Count before (only things like "Count_Cells", named after the object and present in the exported Image.csv file)... is Image_Count an end user-readable metric? Or is it some internal thing?

mcquin:
@AnneCarpenter `Image_Count` is misleading. On further investigation, it's an Image measurement named `Count_OBJECTNAME`.

PR #2560, opened 2017-01-13
Clean IdentifyPrimaryObjects formatting

mcquin:
* documentation formatting
  * cleaner HTML
  * use formatting and explicit variables
* optimize imports and remove import aliases
* PEP08 compliance

Next: IPO should support volumes :)


AnneCarpenter:
Glad you got the tough stuff out of the way, now onto 3D, easy-peasy!

Issue #57, opened 2011-11-23
JIRA [IMG-1532] IdentifyPrimaryObjects:"Merge small objects"non-existent?


thouis:
See this post: http://cellprofiler.org/forum/viewtopic.php?f=16&t=1480 <br/><br/>AFAICT, the poster is right; this setting isn&#39;t connected to anything in IdentifyPrimaryObjects. If so, the functionality should be added or the setting should be removed altogether if it&#39;s not useful.

Original reporter: mbray
Jira link: https://jira.broadinstitute.org/browse/IMG-1532
Fix version: next
Assignee: leek

PR #2576, opened 2017-02-01
Remove unhelpful and unused settings from ApplyThreshold and Identify modules

### LINKED ISSUE #2575: Remove unhelpful and unused settings from ApplyThreshold and Identify modules.

mcquin:
* Remove "Per object" option from "Thresholding strategy" setting. Replace value with "None". Users must update their pipelines to pick another strategy.
  *  Remove associated setting "Masking objects".
* Remove "Mixture of Gaussian" option from "Thresholding method" setting. Replace value with "None". Users must update their pipelines to pick another method.
  * Remove associated setting "Approximate fraction of image covered by objects?".
* Remove setting "Method to calculate adaptive window size". Always specify adaptive window size.
  * Increase default "Size of adaptive window" from 10 to 50.
* Remove "Laplacian of Gaussian" option from "Method to distinguish clumped objects" setting. Replace value with "Intensity".
  * Remove associated settings:
    * "Automatically calculate the threshold using the Otsu method?"
    * "Enter Laplacian of Gaussian threshold"
    * "Automatically calculate the size of objects for the Laplacian of Gaussian filter?"
    * "Enter LoG filter diameter"
* Remove settings "Retain outlines of the identified objects?" and "Name the outline image". Use OverlayOutlines instead.
* Remove setting "Handling of objects if excessive number of objects identified".
  * Remove associated setting "Maximum number of objects"
* Remove setting  "Try to merge too small objects with nearby larger objects?"
* Remove "Binary image" option from "Thresholding strategy" setting. Replace value with "None". Users must update their pipelines to pick another strategy.
  * Remove associated setting "Select binary image".
* Remove "Truncate" option from setting "Handling of objects if excessive number of objects identified". Replace value with "None". Users must update their pipelines to pick another method.
* Remove "Automatic" option from setting "Threshold strategy".
  * Set "Thresholding method" value to "MCT"
  * Set "Select the smoothing method for thresholding" to "Manual"
  * Set "Threshold smoothing scale" to 1.3488 (`sigma = 1.0`)
* ApplyThreshold only generates binary images.
* Remove setting "Select the smoothing method for thresholding". Always specify "Threshold smoothing scale". Use 0 to apply no smoothing.
* Remove "Minimize the weighted variance or the entropy?". Always minimize weighted variance.
* Remove setting "Use default parameters?" when thresholding with robust background. If default was selected, set parameters to default values.
* Remove "Background" thresholding method in favor of "Robust background". Update parameters to use "Background" thresholding defaults.

Resolves #2575


braymp:
Re: "Select the smoothing method for thresholding" - Is the default now no smoothing, or is it some automatic value?

mcquin:
The default is that there is no smoothing.

braymp:
I think "Binary image" might also be removed as a thresholding strategy; I imagine that using MaskImage/MaskObjects effectively does the same thing?

mcquin:
@braymp I'm not confident in how masks are used in CP. I think "Binary image" will be more robust.  :)

bethac07:
I don't think it's necessary to keep 'Binary', but yeah, otherwise looks good!

mcquin:
TODOs:
- []  Remove `reported_log_filter_diameter`, `reported_log_threshold`

AnneCarpenter:
`Remove settings "Retain outlines of the identified objects?" and "Name the outline image". Use OverlayOutlines instead.`
Heads up, this one might require modification in OverlayOutlines to handle an incoming binary image as opposed to objects (which I think, but am not certain, that it expects):

AnneCarpenter:
This is SO exciting and refreshing! I have some questions...

`Remove setting "Try to merge too small objects with nearby larger objects?"
`
This one is very useful for piecing back together improperly divided nuclei (that is, small chunks); is there a workable workaround?

AnneCarpenter:
`Remove "Per object" option from "Thresholding strategy" setting. Replace value with "None". Users must update their pipelines to pick another strategy.`

This is useful when each object has a background amount of fluorescence, which varies from one object to the next (as in a case of nuclear speckles, or a stain that marks the nucleus strongly and the cytoplasm less strongly). I'm not sure whether there are any helpful workarounds if this is removed, but could we pop up a suggestion when people have this option chosen? Something like: "You might be able to recreate a similar effect by using per-image illumination correction (to suppress the background fluorescence in each object) or by using EnhanceOrSuppress Features (with the speckles option)."

AnneCarpenter:
`Remove setting "Method to calculate adaptive window size". Always specify adaptive window size.
Increase default "Size of adaptive window" from 10 to 50.`

Love this.

AnneCarpenter:
`Remove settings "Select the smoothing method for thresholding" and "Threshold smoothing scale". These operations can be done in other modules.`

Are we talking about removing the smoothing that happens as part of IdentifyPrimary? Or is this something unique to Threshold/ApplyThreshold? If the latter I'm 100% in agreement, if the former I'd like to learn more (because all of IdPrimary can be reconstituted from other modules, the magic is in the mixture).

AnneCarpenter:
`Remove setting "Handling of objects if excessive number of objects identified".`

This is needed for some projects when segmentation fails and thousands of objects are erroneously identified, which then causes CP to crash (though I'm not sure if it's in IdentifyPrim or further down the pipeline). Is there a way to delete the setting but prevent that kind of problem? Like, if it's further down the pipeline that problems are caused, one could put in a module that skips the rest of the pipeline if there are more than NNN objects detected. But I suspect that doesn't save you all any time/energy.

AnneCarpenter:
Overall, I'm really excited about the things you've targeted for simplifying. Some of these I wouldn't have seen and most are not likely to have much impact on users, so I'm happy you went through this exercise!

braymp:
Re: <code>"Try to merge too small objects with nearby larger objects?"</code>- This setting was actually removed a while back since it didn't do anything. https://github.com/CellProfiler/CellProfiler/issues/57

bethac07:
@AnneCarpenter :

> Heads up, this one might require modification in OverlayOutlines to handle an incoming binary image as opposed to objects (which I think, but am not certain, that it expects). I assume this setting is only in Threshold/ApplyThreshold (where, yeah - no idea how this setting snuck in there!) and not in IdentifyPrimary.

OverlayOutlines can take either an image of the outlines or a label matrix.  This is in fact an IDPrimary setting.

> "Remove setting "Try to merge too small objects with nearby larger objects?"
This one is very useful for piecing back together improperly divided nuclei (that is, small chunks); is there a workable workaround?

This has never actually worked- see https://github.com/CellProfiler/CellProfiler/issues/57

>Remove "Per object" option from "Thresholding strategy" setting. Replace value with "None". Users must update their pipelines to pick another strategy.

This is useful when each object has a background amount of fluorescence, which varies from one object to the next (as in a case of nuclear speckles, or a stain that marks the nucleus strongly and the cytoplasm less strongly). I'm not sure whether there are any helpful workarounds if this is removed, but could we pop up a suggestion when people have this option chosen? Something like: "You might be able to recreate a similar effect by using per-image illumination correction (to suppress the background fluorescence in each object) or by using EnhanceOrSuppress Features (with the speckles option)."

Maybe someone else has had a different experience, but this setting does not actually work IME.

>Are we talking about removing the smoothing that happens as part of IdentifyPrimary

In IDPrimary.

braymp:
Re: <code>"Retain outlines of the identified objects?" and "Name the outline image"</code> - Nope, it's in the Identify modules as well. Though I've long since abandoned it in favor of the OverlayOutlines approach.

mcquin:
Wow, you folks are fast on the keyboard!!!

@AnneCarpenter https://github.com/CellProfiler/CellProfiler/pull/2576#issuecomment-276978726: This is smoothing that happens before a threshold is applied. Smoothing the threshold before declumping still exists.

@AnneCarpenter https://github.com/CellProfiler/CellProfiler/pull/2576#issuecomment-276979919: Ah, I understand the use case. I think the "Truncate" option is weird. Should we only support "Continue" and "Erase"?

mcquin:
@braymp Do you use/have you been successful with "Per object" thresholding?

0x00b1:
@mcquin we need a hostile fork of CellProfiler.

braymp:
@mcquin: I can't say that I've looked at it super-closely. I've used in a few cases where it seemed the appropriate method to use, but I don't know if the results would have been different if I masked instead, for example. Would looking at the tests for this method be helpful?

AnneCarpenter:
`@AnneCarpenter #2576 (comment): Ah, I understand the use case. I think the "Truncate" option is weird. Should we only support "Continue" and "Erase"?`

100% on board w deleting Truncate!

AnneCarpenter:
> @AnneCarpenter https://github.com/CellProfiler/CellProfiler/pull/2576#issuecomment-276978726: This is smoothing that happens before a threshold is applied. Smoothing the threshold before declumping still exists.

Oh, wow. Didn't realize this was a step! Looks like it's a teeny amount of smoothing. It's going to be really weird if everyone needs to add a Smooth module with 1 pixel gaussian prior to identifying anything so it's really just empirical whether we can get away without this typically, or not.

AnneCarpenter:
> @AnneCarpenter https://github.com/CellProfiler/CellProfiler/pull/2576#issuecomment-276978726: This is smoothing that happens before a threshold is applied. Smoothing the threshold before declumping still exists.

Wow, didn't realize this was a step! But I do recall there being issues with jaggedy edges as described in the help for this setting. I just don't know how commonly this is a problem. It would definitely be weird if nearly every time someone runs IdentifyPrimary they need to add the Smooth module beforehand to smooth w a 1 pixel gaussian! Most people wouldn't realize they need to and then come ask on the forum why their edges look weird. So I propose we see how results from a few example pipelines look if we just remove it. The fact that this setting is here makes me suspect that it's generally useful but obviously that logic fails sometimes :D

mcquin:
@AnneCarpenter I'm happy to add Gaussian smoothing back in, with configurable parameters as an advanced setting.

Similar for too many objects, sans truncate.

mcquin:
@bethac07 Can you please re-review changes https://github.com/CellProfiler/CellProfiler/pull/2576#issue-204672253 and 👍 , 👎 ? Thanks!!

cellprofiler/modules/applythreshold.py
```diff
@@ -33,70 +27,94 @@
...
+            doc="Choose the image to be thresholded."
+        )
+
+        self.thresholded_image_name = cellprofiler.setting.ImageNameProvider(
+            "Name the output image",
+            "ThreshBlue",
+            doc="Enter a name for the thresholded image."
+        )
+
+        self.binary = cellprofiler.setting.Choice(
```
mcquin: Remove this option.

cellprofiler/modules/applythreshold.py
```diff
@@ -33,70 +27,94 @@
...
+                "BINARY": BINARY
+            })
+        )
 
         # if not binary:
-        self.low_or_high = cps.Choice(
-                "Set pixels below or above the threshold to zero?",
-                [TH_BELOW_THRESHOLD, TH_ABOVE_THRESHOLD], doc="""
-            <i>(Used only when "%(GRAYSCALE)s" thresholding is selected)</i><br>
+        self.low_or_high = cellprofiler.setting.Choice(
```
mcquin: Remove this option.

cellprofiler/modules/applythreshold.py
```diff
@@ -33,70 +27,94 @@
...
+            })
+        )
 
         # if not binary and below threshold
-
-        self.shift = cps.Binary(
-                "Subtract the threshold value from the remaining pixel intensities?", False, doc='''
-            <i>(Used only if the output image is %(GRAYSCALE)s and pixels below a given intensity are to be set to zero)</i><br>
-            Select <i>%(YES)s</i> to shift the value of the dim pixels by the threshold value.''' % globals())
+        self.shift = cellprofiler.setting.Binary(
```
mcquin: Remove this option.

cellprofiler/modules/applythreshold.py
```diff
@@ -33,70 +27,94 @@
...
         # if not binary and above threshold
-
-        self.dilation = cps.Float(
-                "Number of pixels by which to expand the thresholding around those excluded bright pixels",
-                0.0, doc='''
-            <i>(Used only if the output image is grayscale and pixels above a given intensity are to be set to zero)</i><br>
-            This setting is useful when attempting to exclude bright artifactual objects:
-            first, set the threshold to exclude these bright objects; it may also be desirable to expand the
-            thresholded region around those bright objects by a certain distance so as to avoid a "halo" effect.''')
+        self.dilation = cellprofiler.setting.Float(
```
mcquin: Remove this option.

cellprofiler/modules/identify.py
```diff
@@ -156,549 +138,513 @@
...
-                'Threshold strategy',
-                TS_ALL, doc="""
-            The thresholding strategy determines the type of input that is used
-            to calculate the threshold. The image thresholds can be based on:
+    threshold_setting_version = 3
+
+    def create_threshold_settings(
+            self,
+            methods=[centrosome.threshold.TM_BACKGROUND,
+                     centrosome.threshold.TM_KAPUR,
```
mcquin: Remove this option.

cellprofiler/modules/identify.py
```diff
@@ -156,549 +138,513 @@
...
+                     centrosome.threshold.TM_MCT,
+                     centrosome.threshold.TM_OTSU,
+                     centrosome.threshold.TM_RIDLER_CALVARD,
```
mcquin: Remove this option.

cellprofiler/modules/identify.py
```diff
@@ -156,549 +138,513 @@
...
-            one of your images, you might check whether the automatically calculated
-            threshold was unusually high or low compared to the other images. See the
-            <b>FlagImage</b> module if you would like to flag an image based on the threshold
-            value.</p>
-
-            <p>There are a number of methods for finding thresholds automatically:
+                <li>The pixel intensities of the input image (this is the most common).</li>
+                <li>A single value manually provided by the user.</li>
+                <li>A single value produced by a prior module measurement.</li>
+                <li>A binary image (called a <i>mask</i>) where some of the pixel intensity values are set to
```
mcquin: Remove line.

cellprofiler/modules/identify.py
```diff
@@ -156,549 +138,513 @@
...
+                0, and others are set to 1.</li>
+            </ul>These options allow you to calculate a threshold based on the whole image or based on image
+            sub-regions such as user-defined masks or objects supplied by a prior module.<br>
```
mcquin: Remove line.

cellprofiler/modules/identify.py
```diff
@@ -156,549 +138,513 @@
...
+                "TM_MCT": centrosome.threshold.TM_MCT,
+                "TM_OTSU": centrosome.threshold.TM_OTSU,
+                "TM_RIDLER_CALVARD": centrosome.threshold.TM_RIDLER_CALVARD,
+                "TM_ROBUST_BACKGROUND": centrosome.threshold.TM_ROBUST_BACKGROUND
+            })
+        )
+
+        self.threshold_smoothing_choice = cellprofiler.setting.Choice(
+            "Select the smoothing method for thresholding",
+            [TSM_MANUAL, TSM_NONE],
```
mcquin: Consider removing "None" option in favor of 0-value for scale.

cellprofiler/modules/identify.py
```diff
@@ -156,549 +138,513 @@
...
+            </dl>
+            """.format(**{
+                "O_THREE_CLASS": O_THREE_CLASS,
+                "O_TWO_CLASS": O_TWO_CLASS,
+                "PROTIP_AVOID_ICON": PROTIP_AVOID_ICON,
+                "PROTIP_RECOMEND_ICON": PROTIP_RECOMEND_ICON
+            })
+        )
+
+        self.use_weighted_variance = cellprofiler.setting.Choice(
```
mcquin: Remove this setting.

cellprofiler/modules/identify.py
```diff
@@ -156,549 +138,513 @@
...
-        self.threshold_scope = cps.Choice(
-                'Threshold strategy',
-                TS_ALL, doc="""
-            The thresholding strategy determines the type of input that is used
-            to calculate the threshold. The image thresholds can be based on:
+    threshold_setting_version = 3
+
+    def create_threshold_settings(
+            self,
+            methods=[centrosome.threshold.TM_BACKGROUND,
```
mcquin: Consider removing in favor of robust background.

cellprofiler/modules/identifyprimaryobjects.py
```diff
@@ -516,7 +476,7 @@ def create_settings(self):
         )
 
         self.automatic_suppression = cellprofiler.setting.Binary(
-            AUTOMATIC_MAXIMA_SUPPRESSION_SETTING_TEXT,
+             AUTOMATIC_MAXIMA_SUPPRESSION_SETTING_TEXT,
```
mcquin: Remove leading whitespace.

dlogan:
Wow, I can't take all this in -- is this a new Executive Order?  :)

mcquin:
@dlogan We're attempting to identify and remove the settings in ApplyThreshold and IXO modules that don't work and typically are unused. Our goal is to make IPO and ISO easier to use without compromising effectiveness. I'm keeping a summary of changes up to date in the PR description. :)

cellprofiler/modules/applythreshold.py
```diff
@@ -1,25 +1,19 @@
-"""<b>Apply Threshold</b> sets pixel intensities below or above a certain threshold to zero
+"""
+<b>Apply Threshold</b> sets pixel intensities below or above a certain threshold to zero
 <hr>
-<b>ApplyThreshold</b> produces either a grayscale or binary image
-based on a threshold which can be pre-selected or calculated automatically using one of many methods.
+<b>ApplyThreshold</b> produces either a grayscale image based on a threshold which can be
```
mcquin: Remove "either"

cellprofiler/modules/applythreshold.py
```diff
@@ -30,134 +24,66 @@
...
-        # if not binary and above threshold
-
-        self.dilation = cps.Float(
-                "Number of pixels by which to expand the thresholding around those excluded bright pixels",
-                0.0, doc='''
-            <i>(Used only if the output image is grayscale and pixels above a given intensity are to be set to zero)</i><br>
-            This setting is useful when attempting to exclude bright artifactual objects:
-            first, set the threshold to exclude these bright objects; it may also be desirable to expand the
-            thresholded region around those bright objects by a certain distance so as to avoid a "halo" effect.''')
+        threshold_methods = [
```
mcquin: Remove. Use Identify default.

PR #2597, opened 2017-02-13
Use binary operations where appropriate (resolves #2595)

### LINKED ISSUE #2595: Use binary morphology operations when appropriate
For example, when given binary data use `skimage.morphology.binary_opening` instead of `skimage.morphology.opening`. Improves running time.

`binary_data` has shape `(92, 512, 512)`:

screen shot 2017-02-13 at 10 56 07 am

mcquin:


0x00b1:
👍

PR #2620, opened 2017-02-24
Anisotropic Gaussian filter

karhohs:
Previously, the GaussianFilter module filtered uniformly across all dimensions. However, given the spatial differences that occur when acquiring z-stacks, i.e. the z-spacing is often greater than the pixel size, it is good practice to adjust the filter to account for this difference.

### DIFF
--- a/cellprofiler/modules/gaussianfilter.py
+++ b/cellprofiler/modules/gaussianfilter.py
@@ -8,6 +8,7 @@ Gaussian filter
 
 import cellprofiler.module
 import cellprofiler.setting
+import cellprofiler.image
 import skimage.filters
 import numpy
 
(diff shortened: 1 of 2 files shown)


cellprofiler/modules/gaussianfilter.py
```diff
@@ -28,6 +30,29 @@ def run(self, workspace):
...
+
+        images = workspace.image_set
+
+        x = images.get_image(x_name)
+
+        dimensions = x.dimensions
+
+        x_data = x.pixel_data
+
+        spacing = numpy.divide(x.spacing[1], x.spacing)
```
0x00b1: `spacing = x.spacing[1] / x.spacing`

cellprofiler/modules/gaussianfilter.py
```diff
@@ -28,6 +30,29 @@ def run(self, workspace):
...
+        sigma = numpy.multiply(self.sigma.value, spacing)
```
0x00b1: `sigma = self.sigma.value * spacing`

PR #2626, opened 2017-02-27
Document loading volumes (resolves #2625)

### LINKED ISSUE #2625: Update documentation on loading volumes.

mcquin:


cellprofiler/gui/help.py
```diff
@@ -2118,8 +2118,13 @@
         process the images in all movies together as if they were constituents of only one movie.
     </li>
 </ul>
+If your images represent a volume, you can follow the above example to process your volume. It is important to note,
+however, that CellProfiler will analyze each image individually. If you would like to analyze your data as a whole
+volume, we suggest building a .TIF stack from the individual image files using another software application, like FIJI.
+Additionally, whole volume processing is only supported for single-channel volumes. If your images are multi-channel
+you will need to separate the channels before combining images into a single-image stack.
```
mcquin: @bethac07 @karhohs @AnneCarpenter Can you please verify the above paragraph makes sense?

cellprofiler/gui/help.py
```diff
@@ -2129,15 +2134,48 @@
...
+            <li>Provide a descriptive name for the channel <i>GFP</i>.</li>
+            <li>Click the "Update" button below the divider to confirm that the proper images are listed and
+            matched across the channels. All file names ending in CH01.TIF should be matched together.</li>
+        </ul>
+    </li>
+    <li>In the <b>Groups</b> module, select "Yes" to automatically group your image sets by image name. All images
+    beginning with IMG01 will be processed in a single group, while images beginning with IMG02 will be processed in
+    another group, and so on. This way, CellProfiler can process all channels from each image.
+    </li>
+</ul>
```
mcquin: @bethac07 @karhohs @AnneCarpenter Can you please verify the above example makes sense? Lines 2137-2171.

AnneCarpenter:
FLEX example: it’s confusing bc I expected the example to be movies or time-lapse but it’s not currently described as such, can we either provide some context before this example explaining that we are switching gears from 3d/time? Or, add a description that says it's 8 fields of view, 3 channels, and N timepoints?

cellprofiler/gui/help.py
```diff
@@ -2027,287 +2027,323 @@
...
-</li>
-</ul>
-</p>
-
+If your images represent a volume, you can follow the above example to process your volume. It is important to note,
+however, that CellProfiler will analyze each Z-slice image individually and sequentially. If you would like to analyze 
+each Z-stack as a whole
+volume, we suggest building a .TIF stack from the individual image files using another software application, like FIJI
+and then using the instructions below for single-file sequences.
+Additionally, whole volume processing is only supported for single-channel volumes. If your images are multi-channel
```
AnneCarpenter: About this phrase "whole volume processing is only supported for single-channel volumes.” Don’t we currently find nuclei then expand outwards to find cell edges (using membrane stain)?

cellprofiler/gui/help.py
```diff
@@ -2027,287 +2027,323 @@
...
+            proper metadata values are being collected from each image.</li>
+        </ul>
+    </li>
+    <li>In the <b>NamesAndTypes</b> module, assign the channel(s) to a name of your choice. If there are multiple
+    channels, you will need to do this for each channel.<br>
+        For this example, you could do the following:
+        <ul>
+            <li>Select "Assign images matching rules".</li>
+            <li>Make a new rule <code>[Metadata][Does][Have Stain matching][fluor]</code> and name it
+            <i>OrigFluor</i>.</li>
```
AnneCarpenter: @karhohs 
Could you suggest alternate names to replace OrigFluor and OrigFluo2? The current example is weird because renaming is done but it’s not substantive. Would be better if “fluor” is renamed to be “DNA” and “fluo2” becomes “actin”. But I don’t know if those are realistic, maybe you or Beth know the stain names better than I do.

bethac07: DNA and Actin seem fine to me

cellprofiler/gui/help.py
```diff
@@ -2027,287 +2027,323 @@
...
-<li>The tables below this setting will update themselves, and you should be able to visually confirm that
-each well is defined as a group, each with 287 frames' worth of images.</li>
-</ul>
-Without this step, CellProfiler would not know where one movie ends and the next one begins, and would
-process the images in all movies together as if they were constituents of only one movie.
-</li>
-</ul>
-</p>
-
+If your images represent a volume, you can follow the above example to process your volume. It is important to note,
```
bethac07: You're going to hate me for this, but I think 'volume' is going to be a problem here (and elsewhere).  You'd be better off using '3D stack' or 'series of Z-stacks' or something like that.  Unless @karhohs or @AnneCarpenter have a better way of putting it, but 'volume' is not commonly used in microscopy and I think will confuse people.

cellprofiler/gui/help.py
bethac07: Yes, but I think each channel is loaded as a separate file.

cellprofiler/gui/help.py
```diff
@@ -2027,287 +2027,323 @@
...
+however, that CellProfiler will analyze each Z-slice image individually and sequentially. If you would like to analyze 
+each Z-stack as a whole
+volume, we suggest building a .TIF stack from the individual image files using another software application, like FIJI
+and then using the instructions below for single-file sequences.
+Additionally, whole volume processing is only supported for single-channel volumes. If your images are multi-channel
+you will need to separate the channels before combining images into a single-image stack.
```
bethac07: Suggested rewrite of that paragraph:

"If your image series represents a 3D image, while you _can_ essentially follow the instructions for the grouping of time series, it will mean that your individual z slices will be processed separately rather than as one 3D whole.  If you want to analyze the series as a whole, we suggest building a .TIF stack from the individual image files using another software application, like FIJI and then using the instructions below for single-file sequences.  Please note that 3D images should be separated into their individual channels, ie one 3D .TIF stack for the DNA channel, one 3D .TIF stack for the actin channel, etc."

We're going to want to either write or point them to such a tutorial for making such a series as well.

cellprofiler/gui/help.py
bethac07: To be super clear then- if I have a 12 frame tif series from my microscope that represents 2 channels, 6 z planes, I need to split it into two 6-frame tif series in ImageJ or elsewhere, right?

cellprofiler/gui/help.py
```diff
@@ -2027,287 +2027,323 @@
...
-information is contained in the <i>C</i> and <i>Series</i> columns. In the figure shown, the <i>C</i> column
-shows three unique values for the channels represented, numbered from 0 to 2. The <i>Series</i> column shows
-8 values for the slices collected in each stack, numbered from 0 to 7, followed by the slices for other stacks. </li>
+<p>In this case, the procedure to set up the input modules to handle these files is as follows:</p>
+<ul>
+    <li>In the <b>Images</b> module, drag-and-drop your folders of images into the File list panel. If necessary, set
+    your rules accordingly in order to filter out any files that are not images to be processed.<br>
+    In the above example, you would drag-and-drop the .TIF files into the File list panel.</li>
+    <li>In the <b>NamesAndTypes</b> module, select "Yes" for "Volumetric" to tell CellProfiler your images should be
+    processed as whole volumes. You can provide the voxel spacing in each dimension. CellProfiler will use this
```
bethac07: Instead of 'Volumetric', I think the visible text for the setting should be something like 'Is this image a 3D volume?'

cellprofiler/gui/help.py
karhohs: I agree with Beth that dna and actin are good placeholders, especially if there is an example image. I think it is worth considering using colors such as "green" or "red" or some kind of amalgam of wavelength and stain, such as DAPI_405 or green_488 or actin_563.

cellprofiler/gui/help.py
AnneCarpenter: Awaiting @mcquin to confirm that's accurate, but if so and if IIUC, this means the text should reads something more like: "Additionally, 3D data must be loaded into CellProfiler as separate, single-channel files. If your images are multi-channel..."

cellprofiler/gui/help.py
karhohs: I like @bethac07 idea to point them to a tutorial. I think it is okay to keep this short: "Whole volume processing is only supported for single-channel .TIF stacks. Converting image sets into .TIF stacks can be done using another software application, like FIJI. For detailed instructions see X tutorial at cellprofiler.org"

cellprofiler/gui/help.py
AnneCarpenter: Well, that is indeed a good point. One reason 'volume' is a little confusing to a biologist is that it sounds like the stack has a single object in it. I could go either way on this - I don't think we need to completely avoid using the term volume, but we can help clarify by defining it first (like "volume (that is, a 3D stack)") and avoid using the term volume in isolation, as in a description for a setting where we don't have a lot of text to elaborate.

cellprofiler/gui/help.py
AnneCarpenter: Definitely agree that "Volumetric" here would be opaque to most biologists. Alternate suggestions: "3D image stack" or "3D volume", if you want to avoid using question forms in the settings.

cellprofiler/gui/help.py
```diff
@@ -2027,287 +2027,323 @@
...
+            the individual stack, and selecting this parameter ensures that the channels are first matched within each
+            stack, rather than across stacks.</li>
+            <li>Click the <img src="memory:{MODULE_ADD_BUTTON}">&nbsp; button to the right to add another row, and
+            select <i>Series</i> for each channel.</li>
+            <li>Click the "Update" button below the divider to confirm that the proper image slices are listed and
+            matched across the channels. The corresponding <i>FileLocation</i> and <i>Series</i> for each channel
+            should now be matched to each other.</li>
+        </ul>
+    </li>
+    <li>In the <b>Groups</b> module, select the metadata that defines a distinct image stack. For the example above, do
```
bethac07: I don't get the rationale for this.  Are they 8 sites, or 8 slices?  Line 2182 seems to suggest that they're individual sites, but this seems more like what you would do if they were slices.

cellprofiler/gui/help.py
```diff
@@ -2027,287 +2027,323 @@
...
+    </li>
 </ul>
-Without this step, CellProfiler would not know where one stack ends and the next one begins, and would
-process the slices in all stacks together as if they were constituents of only one stack.
-</li>
+<p><i>Example:</i> You have two Z-stacks in the following format:</p>
+<ul>
+    <li>The stacks are in Zeiss' CZI format.</li>
+    <li>Each stack consists of a number of slices with 4 channels (DAPI, GFP, Texas Red and Cy3) at each slice.</li>
+    <li>One stack has 9 slices, two stacks have 7 slices and the fourth has 12 slices. Even though the stacks were
```
bethac07: Do you have 2, or do you have 4?

cellprofiler/gui/help.py
```diff
@@ -2027,287 +2027,323 @@
...
-<li>Each stack consists of a number of slices with 4 channels (DAPI, GFP, Texas Red and Cy3) at each slice. </li>
-<li>One stack has 9 slices, two stacks have 7 slices and the fourth has 12 slices. Even though the stacks
-were collected with differing numbers of slices, the pipeline to be constructed is intended to analyze all stacks in
-the same manner.</li>
-<li>Each slice is in grayscale format.</li>
+<p>In this case, the procedure to set up the input modules to handle these this file is as follows:</p>
+<ul>
+    <li>In the <b>Images</b> module, drag-and-drop your folders of images into the File list panel. If necessary, set
+    your rules accordingly in order to filter out any files that are not images to be processed.<br>
+    In the above example, you would drag-and-drop the LSM files into the File list panel. In this case, the default
```
bethac07: Are these LSM or CZI files?

cellprofiler/gui/help.py
```diff
@@ -2027,287 +2027,323 @@
...
+    "Images only" filter is sufficient to capture the necessary files.</li>
+    <li>In the <b>Metadata</b> module, enable metadata extraction in order to obtain metadata from these files. The key
+    step here is to obtain the metadata tags necessary to do two things:
+        <ul>
+            <li>Distinguish the stacks from each other. This information is contained as the file itself, that is, each
+            file represents a different stack.</li>
+            <li>For each stack, distinguish the timepoints from each other, ensuring proper ordering. This information
```
bethac07: These are Z planes, not time point

cellprofiler/gui/help.py
mcquin: @bethac07 is correct. I can add an example outlining what should be done here.

cellprofiler/gui/help.py
mcquin: I understand the confusion. :)

cellprofiler/gui/help.py
mcquin: Nice suggestions!

cellprofiler/gui/help.py
mcquin: Nice!!

cellprofiler/gui/help.py
```diff
@@ -2027,287 +2027,323 @@
...
+            <li>Click the "Add another image" button to define a second image with a set of rules.</li>
+            <li>Make a new rule <code>[File][Does][Contain][CH02]</code></li>
+            <li>Provide a descriptive name for the channel <i>GFP</i>.</li>
+            <li>Click the "Update" button below the divider to confirm that the proper images are listed and
+            matched across the channels. All file names ending in CH01.TIF should be matched together.</li>
+        </ul>
+    </li>
+    <li>In the <b>Groups</b> module, select "Yes" to automatically group your image sets by image name. All images
+    beginning with IMG01 will be processed in a single group, while images beginning with IMG02 will be processed in
+    another group, and so on. This way, CellProfiler can process all channels from each image.
```
mcquin: You're probably right. I'll test it out.

cellprofiler/gui/help.py
mcquin: LOL! It looks like I'll have to proof these prior examples.

cellprofiler/gui/help.py
mcquin: I'm not sure. I'm unfamiliar with this file format, and am not having much luck finding information about it on the internet.

cellprofiler/modules/namesandtypes.py
```diff
@@ -314,7 +314,7 @@ def create_settings(self):
                 doc=LOAD_AS_CHOICE_HELP_TEXT)
 
         self.volumetric = cps.Binary(
-            text="Volumetric",
+            text="Data is 3D",
```
0x00b1: Really? 3D could mean grayscale time-lapse.

mcquin: What do you suggest?

AnneCarpenter: @bethac07 and @karhohs could chime in. Requires thinking about the context in which this is being asked.

braymp: Perhaps `Data is 3-D (xyz)?`

AnneCarpenter: IIUC, one might have 3d (xyz) data but want to process it slice by slice, in such a case the answer here would be "no". Now, we can just assume that in this case the person will know they are over-riding 'normal' 3D processing behavior and so maybe they will feel ok checking 'no'.

0x00b1: @AnneCarpenter IIRC, this is why I went with “Volumetric.” Could you make a final decision? I’d like to merge.

PR #2638, opened 2017-03-14
RemoveObjects should operate on objects (resolves #2631)

### LINKED ISSUE #2631: The input list are images for Remove objects module not segmented objects (nightly 2-16-17)
The Remove objects module only displays images as inputs not the expected objects that have been segmented in preceding modules.  For example, the segmented nuclei objects which are output from the watershed module do not show up in Remove objects.

mcquin:
RemoveObjects module should remove small objects from objects not images.

Additionally implements base class `cellprofiler.module.ObjectProcessing` with ability to record basic object measurements.


cellprofiler/measurement.py
```diff
@@ -1983,6 +1983,8 @@ def __init__(self, module_number, relationship,
 M_LOCATION_CENTER_X = '%s_%s' % (C_LOCATION, FTR_CENTER_X)
 FTR_CENTER_Y = "Center_Y"
 M_LOCATION_CENTER_Y = '%s_%s' % (C_LOCATION, FTR_CENTER_Y)
+FTR_CENTER_Z = "Center_Z"
```
0x00b1: Nice

cellprofiler/module.py
```diff
@@ -1034,3 +1035,226 @@ def visible_settings(self):
             self.x_name,
             self.y_name
         ]
+
+
+class ObjectProcessing(Module):
```
0x00b1: Nice

cellprofiler/module.py
```diff
@@ -1034,3 +1035,226 @@ def visible_settings(self):
...
+
+    def add_measurements(self, workspace):
+        objects = workspace.object_set.get_objects(self.y_name.value)
+
+        labels = objects.segmented
+
+        if not objects.volumetric:
+            labels = numpy.asarray([labels])
+
+        centers = mahotas.center_of_mass(numpy.ones_like(labels), labels=labels)
```
0x00b1: If we use `scipy.ndimage.measurements.center_of_mass`, you can save the Mahotas import.

cellprofiler/modules/removeobjects.py
```diff
@@ -42,6 +45,21 @@ def visible_settings(self):
         ]
 
     def run(self, workspace):
-        self.function = skimage.morphology.remove_small_objects
+        self.function = lambda labels, diameter: remove_objects(labels, diameter)
 
         super(RemoveObjects, self).run(workspace)
+
+
+def remove_objects(labels, diameter):
```
0x00b1: We should consider pushing this upstream to scikit-image.

mcquin: I think the parameter makes sense for scikit-image. I think the documentation could be improved.

Issue #2647, opened 2017-03-17
IdentifySecondaryObjects fails in master


bethac07:
Running this pipeline
[BBBC022_Analysis_Final.zip](https://github.com/CellProfiler/CellProfiler/files/851763/BBBC022_Analysis_Final.zip)
 (made in 2.2).  
```
Traceback (most recent call last):
  File "I:\GitHub\CellProfiler\cellprofiler\pipeline.py", line 1890, in run_image_set
    self.run_module(module, workspace)
  File "I:\GitHub\CellProfiler\cellprofiler\pipeline.py", line 1997, in run_module
    module.run(workspace)
  File "I:\GitHub\CellProfiler\cellprofiler\modules\identifysecondaryobjects.py", line 479, in run
    (labels_in[0, :], labels_in[-1, :], labels_in[:, 0], labels_in[:, -1]))
  File "C:\Users\bcimini\AppData\Local\Continuum\Anaconda2\lib\site-packages\numpy\core\shape_base.py", line 280, in hstack
    return _nx.concatenate(arrs, 1)
ValueError: all the input array dimensions except for the concatenation axis must match exactly
```

braymp:
Duplicate of #2617 (though that issue goes off on a tangent). Try the dependency update suggestion and see if it helps.

mcquin:
This is caused by comparisons like `value is 3`. Windows stores values as longs and `3L is not 3`.

mcquin:
Resolved #2650 2650

PR #2650, opened 2017-03-20
compare numeric values with '==', not 'is'

mcquin:


mcquin:
@bethac07 @braymp Here's the solution to one of your problems. :) (#2647)

PR #2693, opened 2017-04-11
Add error message to dialog. Print stack trace to logger. (Resolves #2691)

### LINKED ISSUE #2691: Error dialogue should include error message and stacktrace.
screen shot 2017-04-11 at 10 59 03 am

mcquin:
screen shot 2017-04-11 at 2 23 29 pm


braymp:
Is the stack trace easy to access for the average user (say, for reporting an error on the forum)?

mcquin:
It's available in the console, so it might not be the easiest to find. I'd
like to have a "View stack trace" button, though it requires additional
work.

PR #2714, opened 2017-04-24
Indicate if a module supports 3D processing.

### LINKED ISSUE #2710: Indicate 3D compatible modules in GUI.
Perhaps an error could appear next to the module when "Process as 3D" is true and the module does not support 3D processing.

mcquin:
Indicate if a module supports 3D processing by raising an error when

1) Process as 3D  is "Yes" and a 2D-only module is added to the pipeline:
screen shot 2017-04-24 at 8 22 14 am

2) Running the pipeline and any module is 2D-only and Process as 3D is "Yes":
screen shot 2017-04-24 at 8 22 08 am

Resolves #2710


braymp:
For the former, does it have to be a full-on dialog window? I'd imagine that the usual red warning flag might be sufficient.

mcquin:
The reg warning flags are difficult to debug. I don't know there's a way to see the error until you try to run the pipeline. The error on adding a non-3D module to pipeline (when configured for 3D) is a fail fast solution that gives users more transparency.

0x00b1:
Our current validation implementation is also responsible for numerous redrawing bugs. Especially on Windows.

PR #2722, opened 2017-04-26
File > Save Project As... default directory

jhung0:
File > Save Project As... dialog to default to the directory where the
project is from


0x00b1:
Awesome! Thanks, @jhung0!

PR #2729, opened 2017-04-29
TrAM module

DanRuderman:
This is a pull request for the initial check in of the TrAM module.

You can add this module after tracking with TrackObjects. You will be able
to select the objects to perform the computation on and their corresponding
measurements to locate jumps in. There is documentation in the help window.

Let me know if you have any questions or difficulties testing. I'm happy to
provide a test project and images.


0x00b1:
Hi, @DanRuderman, a test pipeline (and test images) would be helpful!

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
+from cellprofiler.modules import trackobjects
+import numpy as np
+import itertools as it
+import cellprofiler.module as cpm
+import cellprofiler.setting as cps
+import cellprofiler.measurement as cpmeas
```
mcquin: Please refer to our [style guide](https://github.com/CellProfiler/CellProfiler/blob/master/CONTRIBUTING.md#style-guide) for style guidelines.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+                "Tracked objects", cps.NONE, doc="""
+            Select the tracked objects for computing TrAM.""")
+
+        # which measurements will go into the TrAM computation
+        self.tram_measurements = MeasurementMultiChoiceForCategory(
+            "TrAM measurements", category_chooser=self.object_name, doc="""
+            These are measurements for the selected tracked objects which
+            will be used in the TrAM computation. At least one must be selected.
+            Note there may be a delay of a few seconds between the selection of
+            <i>Tracked objects</i> and the update of this selection component.""")
```
mcquin: This last sentence is unnecessary.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+
+        # Treat X-Y value pairs as Euclidian in the TrAM measure?
+        self.wants_XY_Euclidian = cps.Binary(
```
mcquin: Methods and properties should be lowercase.

mcquin: I would consider renaming this property to `self.isotropic`. It'll make more sense to users.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+            If selected (the default) then measurements that are available
+            as X-Y pairs (e.g. location) will be have a Euclidian (isotropic)
+            metric applied in TrAM. Note that the X-Y-Z extension of this feature
+            is not currently available.
+            """)
+
+        # number of spline knots
+        self.num_knots = cps.Integer(
+            "Number of spline knots", 4, minval=MIN_NUM_KNOTS, doc="""
+            Number of spline knots</i>The number of knots (indpendent values) used
```
mcquin: `Number of spline knots</i>` can be removed. The setting text automatically appears in the help documentation.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+            when computing smoothing splines. This should be around 1/5th the number
+            of frames for reasonably oversampled time lapse sequences, and must be 3
+            or greater. It is approximately the maximum number of wiggles expected in
+            well-tracked trajectories
+            """)
+
+        # TrAM exponent
+        self.p = cps.Float(
```
mcquin: Avoid single letter variables.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+            "TrAM exponent", 0.5, minval=0.01, maxval=1, doc="""
```
mcquin: `maxval` should be `1.0` to avoid an implicit cast.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+        # check on available tracking columns for the selected object
+        obj_name = self.object_name.get_value()
+        mc = pipeline.get_measurement_columns()
+        num_tracking_cols = len([entry for entry in mc if entry[0] == obj_name and entry[1].startswith(trackobjects.F_PREFIX)])
+        if num_tracking_cols == 0:
+            msg = "No %s data available for %s. Please select an object with tracking data."\
+                  % (trackobjects.F_PREFIX, obj_name)
+            raise cps.ValidationError(msg, self.object_name)
+
+    def visible_settings(self):
```
mcquin: You can omit implementing the `visible_settings` method when it returns `self.settings()`.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+            display_data - the run() module should store anything to be
+                           displayed in this attribute, which will be used in
+                           display()
+
+        run() should not attempt to display any data, but should communicate it
+        to display() via the workspace.
+        """
+        pass
+
+
```
mcquin: Unnecessary new lines. Only one line is required between the end of one method and the beginning of the next.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+            measurements - the measurements for this run
+            frame        - the parent frame to whatever frame is created.
+                           None means don't draw.
+            display_data - the run() module should store anything to be
+                           displayed in this attribute, which will be used in
+                           display()
+
+        run() should not attempt to display any data, but should communicate it
+        to display() via the workspace.
+        """
```
mcquin: This documentation is unnecessary and can be deleted.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+
+        This method is only called if self.show_window is True
+
+        workspace - the current workspace. workspace.display_data should have
+                    whatever information is needed for the display. Numpy arrays
+                    lists, tuples, dictionaries and Python builtin objects are
+                    allowed.
+
+        figure - the figure to use for the display.
+        """
```
mcquin: This documentation is unnecessary and can be deleted.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+    def post_group(self, workspace, grouping):
+        self.show_window = True
+
+        measurements = workspace.measurements
+        obj_name = self.object_name.get_value() # the object the user has selected
+        img_numbers = measurements.get_image_numbers()
+        num_images = len(img_numbers)
+
+        def flatten_list_of_lists(lol):
+            return list(it.chain.from_iterable(lol))
```
mcquin: You should flatten a list of lists inline.

mcquin: And use a list comprehension rather than `itertools.chain`, a quick trick for single-nested lists:

```python
list = [[1, 2], [3, 4, 5]]
sum(list, [])
> [1, 2, 3, 4, 5]
```

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+        """If true, the module uses data from other imagesets in a group
+
+        Aggregation modules perform operations that require access to
+        all image sets in a group, generally resulting in an aggregation
+        operation during the last image set or in post_group. Examples are
+        TrackObjects, MakeProjection and CorrectIllumination_Calculate.
+        """
+        return True
+
+class MeasurementMultiChoiceForCategory(cps.MeasurementMultiChoice):
```
0x00b1: Move this into `cellprofiler.setting`.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+        all_values_dict = dict(get_feature_values_tuple(sel) for sel in selections)
+        # determine if there are any euclidian (XY) pairs
+        if self.wants_XY_Euclidian.value:
+            euclidian_pairs = TrAM.Determine_Euclidian_pairs(all_values_dict.keys())
+        else:
+            euclidian_pairs = []
+
+        # sanity check: make sure all vectors have the same length
+        vec_lengths = set([len(value) for value in all_values_dict.values()])
+        assert len(vec_lengths) == 1
```
mcquin: It is good practice to include a message when the assertion fails.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+        """
+        Compute the median absolute temporal difference in each of the features across all tracks
+        
+        :param values_dict: keys are feature names, values are lists of data values across images and tracks
+        :param labels_vec: A list of track labels corresponding to data values in their arrays
+        :param image_vec: A list of image numbers corresponding to data values in their arrays
+        :return: dictionary whose keys are feature names and values are median absolute differences
+        """
+        # input is a list of time series lists
+        def compute_median_abs_deviation(values_lists):
```
0x00b1: Use `mad` from `statsmodels`.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+                    result_dict.update({LABELS_KEY: str(int(tram_label))})
+                else: # if there is already a TRAM_KEY then we are a parent and don't have a valid TrAM
+                    result_dict.update({PARENT_KEY: 1})
+                    result_dict.update({TRAM_KEY: FLOAT_NAN})
+                    result_dict.update({LABELS_KEY: "%s|%s" % (result_dict[LABELS_KEY], str(int(tram_label)))}) # append
+
+                result_dict.update({SPLIT_KEY: split_flag})
+
+        # Loop over all images and save out
+        tram_values_to_save = list()
```
0x00b1: Initialize with `[]` rather than `list()`.

0x00b1:
@DanRuderman You’ll need to write units too before this can be properly reviewed. You should add a test module for both your module and for the “MeasurementMultiChoiceForCategory” setting.

0x00b1:
@DanRuderman Nevertheless, this is a strong contribution and it shouldn’t require too much work before it can be merged into master. 👍

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+        # get what was selected by the user
+        selections = self.tram_measurements.get_selections()
+
+        # get the object set to work on
+        object_name = self.object_name.get_value()
+
+        return [sel for sel in selections if sel.startswith(object_name)]
+
+    def get_measurement_columns(self, pipeline):
+        return [(self.object_name.get_value(), FULL_TRAM_MEAS_NAME, cpmeas.COLTYPE_FLOAT),
```
mcquin: You can use `self.object_name.value`.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+        if object_name == self.object_name.get_value():
+            return [CAT_TRAM]
+        return []
+
+    def get_measurements(self, pipeline, object_name, category):
+        if object_name == self.object_name.get_value() and category == CAT_TRAM:
+            return [MEAS_TRAM, MEAS_PARENT, MEAS_SPLIT, MEAS_LABELS]
+        return []
+
+    def get_measurement_scales(self, pipeline, object_name, category, feature, image_name):
```
mcquin: This isn't necessary for the measurements this module makes.

cellprofiler/modules/tram.py
0x00b1: (or `numpy.median(numpy.abs(x - numpy.median(x)))`)

DanRuderman:
@0x00b1  @mcquin Thanks for all the great feedback! I'm new to Python, so the coding specifics are very helpful. I'll be able to pursue these changes in a couple of weeks, once I've caught up with everything I put on the back burner.

In the meantime, I'd like to get you a set of images and a pipeline so you have the ability to review its function if you have time. Please let me know the best way to approach this.

mcquin:
A Dropbox or Google Drive link to a pipeline and images is perfect.

DanRuderman:
Here is a [dropbox link](https://www.dropbox.com/sh/2azckn6mc4qxe36/AABi_E-szt0gnRQs6kfXaeZka?dl=0) to a directory with an example project and a subdirectory with the images.

Let me know if you have any difficulty.

cellprofiler/modules/__init__.py
```diff
@@ -108,6 +108,7 @@
                         'straightenworms': 'StraightenWorms',
                         'matchtemplate': 'MatchTemplate',
                         'trackobjects': 'TrackObjects',
+                        'tram': 'TrAM',
```
DanRuderman: Makes sense. Can we decide on a name now, since we need to lock it down in the chapter we're writing. How about TrackQuality, TrackMetrics, TrackQC? Other suggestions?

DanRuderman: Going with "TrackQuality"

mcquin: Good choice.

AnneCarpenter:
Does this module produce metrics only? If so does "MeasureTrackQuality" make more sense? I think it fits our naming conventions better. And, is it in the "Measurement" category for modules?

DanRuderman:
Renamed MeasureTrackQuality. It's now in the Measurement category.

cellprofiler/modules/measuretrackquality.py
```diff
@@ -0,0 +1,641 @@
...
+split, then this value is 1. Otherwise it is 0.</li>
+</ul>
+<p>
+See Patsch, K <i>et al.</i>, <a href=https://www.nature.com/articles/srep34785>Single cell
+dynamic phenotyping</a>, Scientific Reports 6:34785 (2016)
+""" % (MIN_TRAM_LENGTH)
+
+logger = logging.getLogger(__name__)
+
+class TrackQuality(cpm.Module):
```
mcquin: Conventionally, module class names match `module_name`:

```python
class MeasureTrackQuality(cpm.Module):
```

DanRuderman:
@mcquin @0x00b1 Finding all `TrackObjects_Label` values to be `nan` in a particular project, though there other `TrackObjects` data available. Does this ring a bell? I can send details but thought I'd check with you in case it's something obvious.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+with partial tracks or those for whom <i>Is_Parent</i> is 1. A histogram of
+all computed TrAM values is displayed to help define a cutoff.</li>
+<li><i>Labels:</i> Each tracked item that has a lineage from the first to the
+last frame is assigned a TrAM label on the last frame. If the final object
+does not arise from a split during tracking, then it has this same unique label
+for its entire track. That label is not assigned to any other objects. But if
+an object arises from a split, then its ancestor object(s) will be assigned
+multiple labels (combined from its progeny). These labels are separated by
+a "|" symbol.</li>
+<li><i>Is_Parent:</i> If the object splits into daughters during its track then
```
DanRuderman: So this is redundant and I should remove?

mcquin:
> Finding all TrackObjects_Label values to be nan in a particular project, though there other TrackObjects data available. Does this ring a bell? I can send details but thought I'd check with you in case it's something obvious.

Hey @DanRuderman, I wouldn't expect `TrackObjects_Label` values to all be `nan`. Can you provide more information?

braymp:
@mcquin: It can happen if the ``Filter objects by lifetime?`` setting in TrackObjects is greater than the track life/lives; see the help for this setting.

bethac07:
Hi Dan- see [this forum thread](http://forum.cellprofiler.org/t/minimum-lifetime-filter/4616)

DanRuderman:
OK - the one thing that turns out to be a bit tricky is handling groups corresponding to disjoint sets of images. Looks like there is some code in `TrackObjects.get_group_image_numbers(self, workspace)` that returns just the image numbers for the current group.  Is that the  code I should use to find the images? If so, does it belong somewhere that it can be more easily shared?

mcquin:
@DanRuderman:

> OK - the one thing that turns out to be a bit tricky is handling groups corresponding to disjoint sets of images. Looks like there is some code in TrackObjects.get_group_image_numbers(self, workspace) that returns just the image numbers for the current group. Is that the code I should use to find the images? If so, does it belong somewhere that it can be more easily shared?

I don't understand what you mean by "group corresponding to disjoint sets of images". Could you please provide an example of this, and explain what you are trying to accomplish?

mcquin:
@DanRuderman I'm trying to run your example pipeline. Unfortunately the name change from TrAM to MeasureTrackQuality means the pipeline doesn't load correctly. I replaced the TrAM module with MeasureTrackQuality, but I wasn't sure which measurements to select. I ran the pipeline twice: once using only TrackObjects features, once using all computed features. I used the default values for the remaining settings. Each time I got this display and error message:

screen shot 2017-05-17 at 1 40 48 pm

```
Worker 3: Wed May 17 13:31:42 2017: Image # 25, module MeasureTrackQuality # 9: 0.00 secs
Progress Counter({u'FinishedWaitingMeasurements': 25})
Worker 3: /Users/mcquin/code/CellProfiler/CellProfiler/cellprofiler/modules/trackobjects.py:2611: RuntimeWarning: invalid value encountered in divide
Worker 3:   linearity[index] = tot_distance / integrated[index]
Worker 3: /usr/local/lib/python2.7/site-packages/numpy/lib/function_base.py:3858: RuntimeWarning: Invalid value encountered in median
Worker 3:   r = func(a, **kwargs)
Worker 3: /Users/mcquin/code/CellProfiler/CellProfiler/cellprofiler/modules/measuretrackquality.py:255: RuntimeWarning: divide by zero encountered in double_scalars
Worker 3:   inv_devs = numpy.diag([1 / tad[k] for k in tram_feature_names]) # diagonal matrix of inverse typical deviation
```

DanRuderman:
@mcquin Regarding the grouping issue, here is an example pipeline and images via dropbox:
https://www.dropbox.com/sh/3icamzpkbogp18v/AAA6ghgI8txPLWLli0GFZ46na?dl=0

The idea is that we are grouping images by the metadata "Field" into two disjoint groups. The TrAM algorithm will therefore be run twice (once per group) and will need to know which images to pay attention to each time. The method `TrackObjects.get_group_image_numbers(self, workspace)` seems to take care of this in the TrackObjects code. I wanted advice on how best to replicate that functionality.

mcquin:
> The TrAM algorithm will therefore be run twice (once per group) and will need to know which images to pay attention to each time. 

I understand now, thanks for the clarification! I don't have an answer for you just yet (I am learning the TrackObjects code as you work on this). I will provide my thoughts soon!

DanRuderman:
@mcquin Thanks, Claire.

Regarding your difficulties getting TrAM up and running, please have a look in that same Dropbox folder for a project named Baseline_short_v3_DR_working.cpproj . Hopefully it will "just work" :)

mcquin:
@DanRuderman It looks like there's a more straightforward way to get the image numbers by group:

```python
measurements = workspace.measurements

group_number = measurements.get_group_number()
# group_number <-- 0

grouping_tags = measurements.get_grouping_tags()
# grouping_tags <-- [u'Metadata_Field']

groupings = measurements.get_groupings(grouping_tags)
# groupings <-- [
#     ({u'Metadata_Field': u'01'}, [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]), 
#     ({u'Metadata_Field': u'02'}, [11, 12, 13, 14, 15, 16, 17, 18, 19, 20])
# ]

image_numbers = groupings[group_number][1]
# image_numbers <-- [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
```

DanRuderman:
@mcquin Thanks for this bit of code (and the recent edit).

I have a question on how to use `Measurements.get_measurement(self, object_name, feature_name, image_set_number)` in this context. In particular the `image_set_number` argument isn't clear to me. Is an image set the same as a group? Do I use the group number as an argument or do I use a list of image numbers?

The documentation says:

> the current image set by default, a single
                           image set number to get measurements for one
                           image set or a sequence of image numbers to
                           return measurements for each of the image sets
                           listed.

I'm a bit confused by the phrase "sequence of image numbers" because it is followed by "each of the image _sets_". Thanks.

mcquin:
@DanRuderman I think the `image_set_number` is the same as the image number. Or you could try using the `image_numbers` defined in the previous code snippet.

DanRuderman:
> @DanRuderman You’ll need to write unit tests before this can be properly reviewed. You should add a test module for both your CellProfiler module and for the “MeasurementMultiChoiceForCategory” setting. If you need some guidance, let us know.

@0x00b1 Yes, guidance would be great. Could do by phone for higher bandwidth next week. My questions relate to:

1. I would want to include some imagery and a pipeline to test the full TrAM functionality. How can I do this (an existing example would be great). Alternatively, I could test just the core algorithm without it.

1. For the “MeasurementMultiChoiceForCategory” setting let me know if there is an existing example of gui testing. I only see an html test under gui tests.

DanRuderman:
> While you may find a number of CellProfiler numbers use global variables, we have asked module authors to avoid them because it impedes automatic refactoring. For example, it is difficult (and sometimes impossible) to automatically refactor documentation strings because of global variables.

@mcquin Took care of this. I'd like to access the value of one of these variables in the `__doc__` string, which is defined in the source code before the variable. (I believe it is standard to define this string at the top of the file?) So the value can't be found. Can you help me with this chicken-and-egg problem?

mcquin:
> I'd like to access the value of one of these variables in the `__doc__` string, which is defined in the source code before the variable. (I believe it is standard to define this string at the top of the file?) So the value can't be found. Can you help me with this chicken-and-egg problem?

@DanRuderman You'll have to define the variable before the `__doc__` string and include it with `"Some text including {ref} and ...".format(**{ref: GLOBAL_VARIABLE})`.

tests/test_formatreader.py
```diff
@@ -0,0 +1,184 @@
+# Python-bioformats is distributed under the GNU General Public
```
mcquin: @DanRuderman Did you write these tests?

DanRuderman: @mcquin Oops. Those do not belong here. Good catch. Will fix.

DanRuderman:
> @DanRuderman You'll have to define the variable before the __doc__ string and include it with "Some text including {ref} and ...".format(**{ref: GLOBAL_VARIABLE}).

Got it, but my understanding is the code base is moving away from use of global variables. That's why I moved it inside the class definition.

mcquin:
@DanRuderman You could move the variable back outside the class (be sure to uniquely name it so that it doesn't conflict with another existing global). Or, you could remove the variable entirely from the doc string and use a hard coded value.

cellprofiler/modules/tram.py
```diff
@@ -0,0 +1,673 @@
...
+This module must be placed downstream of a module that identifies objects
+(e.g., <b>IdentifyPrimaryObjects</b>) and a <b>TrackObjects</b> that tracks
+them. There must be at least %d frames to perform a TrAM analysis. The TrAM
+statistic reflects how typical the maximum deviation from smooth time series
+a chosen set of measurements are. Typical fluctuations are determined from
+measurement differences in adjacent time points of objects whose trajectories
+are complete and without splitting events.
+
+<p><b>TODO</b>For an example pipeline using TrAM see the CellProfiler
+<a href="http://www.cellprofiler.org/examples.html#TrAM">Examples</a> webpage.</p>
```
DanRuderman: @mcquin @N3llz Some possible options on an example for `MeasureTrackQuality`:

1. One option would be to tack a `MeasureTrackQuality` node on the existing example for `TrackObjects`. Hopefully there is a range of track quality that TrAM would reveal. Are there plans to update this existing pipeline for the v3 release? It's loading with some errors since it's not up to date.

1. Another would be to upload the example from the book chapter we submitted. The image data are big, though, like >200MB. Multiple fluorophores. Could cut in half by examining just one of the two microscope fields.

1. Last option would be to delay uploading an example until after v3 is released.

Thoughts?

mcquin:
@DanRuderman -- Regarding testing:

> I would want to include some imagery and a pipeline to test the full TrAM functionality. How can I do this (an existing example would be great). Alternatively, I could test just the core algorithm without it.

I think that's really important! It can be a bit tricky to get set up correctly. I'm working on an example set up for tests for you, which I should be able to share soon. In the meantime, it'd be great to have unit tests for the methods which don't rely on any of the CellProfiler internals (i.e., methods called by `post_group`). We'd prefer you use the [pytest](https://docs.pytest.org/en/latest/) framework for writing tests.

To run only one module's tests:

```sh
python setup.py test -a tests/modules/test_measuretrackquality.py
```

> For the “MeasurementMultiChoiceForCategory” setting let me know if there is an existing example of gui testing. I only see an html test under gui tests.

The test for this would be under `test_settings.py`. However, I don't see one. Honestly (and it isn't often I'll say this), I wouldn't focus on testing this.

mcquin:
@DanRuderman We would like to transfer this module to the [CellProfiler-plugins](https://github.com/CellProfiler/CellProfiler-plugins) repository. Could you please open a pull request with these changes there? Let me know if you have questions!

DanRuderman:
> @DanRuderman We would like to transfer this module to the CellProfiler-plugins repository. Could you please open a pull request with these changes there? Let me know if you have questions!

@mcquin Sure. Some questions:

1. I added the class `MeasurementMultiChoiceForCategory` in settings.py, which is not part of the CellProfiler-plugins repository. Should that remain in this pull request, or close this pull request and create a new one with that change?

1. I was going to create some unit tests on the internal TrAM computations. Where would those go?

1. Will the way that `MeasureTrackQuality` is accessed by the user still be the same? If not we'll need to try to revise our example usage in the book chapter we submitted.

mcquin:
> I added the class MeasurementMultiChoiceForCategory in settings.py, which is not part of the CellProfiler-plugins repository. Should that remain in this pull request, or close this pull request and create a new one with that change?

Let's move `MeasurementMultiChoiceForCategory` to the plugins repository as well. I think the best thing to do would be to create the following directory structure:

```
cellprofiler_plugins/
    __init__.py
    settings.py
measuretrackquality.py
...
```

In `settings.py` you can add your setting class:

```python
import cellprofiler.settings

class MeasurementMultiChoiceForCategory(cellprofiler.setting.MeasurementMultiChoice):
    ...
```

Then, in your module:

```python
import cellprofiler.module
import cellprofiler.settings

import cellprofiler_plugins.settings

class MeasureTrackQuality(cellprofiler.module.Module):
    def create_settings(self):
        self.tram_measurements = cellprofiler_plugins.settings.MeasurementMultiChoiceForCategory(...)
        ...
```

That may not be 100% correct, but should be close. I'm happy to help you get the structure right. :)

> I was going to create some unit tests on the internal TrAM computations. Where would those go?

These would go under a `tests/` directory, similar to how the tests are structured in this repository. For an example, you can check out [this yeast identification pull request](https://github.com/CellProfiler/CellProfiler-plugins/pull/8/files).

> Will the way that MeasureTrackQuality is accessed by the user still be the same? If not we'll need to try to revise our example usage in the book chapter we submitted.

It'll be very easy to add plugins to CellProfiler. Users will need to clone the CellProfiler-plugins repository and provide the local path to CellProfiler. There are instructions in the [README](https://github.com/CellProfiler/CellProfiler-plugins/blob/master/README.md).

mcquin:
Closing in favor of https://github.com/CellProfiler/CellProfiler-plugins/pull/12

PR #2743, opened 2017-05-03
fixed version mismatch message

### LINKED ISSUE #2674: Pipeline version mismatch shows funky numbering
Running the example pipeline by clicking the welcome window in the nightly build yields the attached error. The error itself is probably fine because we won't reconcile the pipeline w the latest version until release. However, the numbering is weird, referring to 3.0 as 300. Could lead to some confused Forum posts from people claiming to be using CP300!

warning

0x00b1:
resolves #2674

### DIFF
--- a/cellprofiler/pipeline.py
+++ b/cellprofiler/pipeline.py
@@ -34,6 +34,7 @@ import urllib
 import urllib2
 import re
 import numpy
+import pkg_resources
 
 logger = logging.getLogger(__name__)
 pipeline_stats_logger = logging.getLogger("PipelineStatistics")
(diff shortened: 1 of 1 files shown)


cellprofiler/pipeline.py
```diff
@@ -948,7 +949,7 @@ def rl():
                         "You can ignore this warning if you do not plan to save\n"
                         "this pipeline or if you will only use it with this or\n"
                         "later versions of CellProfiler."
-                    ).format(git_hash, pipeline_date)
+                    ).format(pkg_resources.get_distribution("CellProfiler").version)
```
mcquin: This will display the current version of CellProfiler, not the version the pipeline was saved with. You want to display which version of CellProfiler this pipeline was saved with.

cellprofiler/pipeline.py
```diff
@@ -951,6 +951,12 @@ def rl():
                     ).format(git_hash, pipeline_date)
                     logging.warning(message)
                 else:
+                    # pipeline versions pre-3.0.0 have unpredictable formatting
+                    if pipeline_version == 300:
```
mcquin: I wonder where the conversion to `300`. It would be nice if the full version information were captured (e.g., "3.0.0rc1"). Does Python have any semantic version support, like version comparison?

0x00b1: Yes, but it wouldn’t work with bundled applications since it assumes the package (and version information) is available from the Python path.

PR #2744, opened 2017-05-03
updated telemetry dialog message

0x00b1:


cellprofiler/gui/dialog.py
```diff
@@ -49,7 +49,7 @@ def __init__(self):
 
         self.SetExtendedMessage(extended_message)
 
-        self.SetYesNoLabels("Enable Telemetry", "Disable Telemetry")
+        self.SetYesNoLabels("Send diagnostic information", "Stop sending diagnostic information")
```
mcquin: Perhaps `"Don't send diagnostic information"`?

PR #2747, opened 2017-05-03
Save images cleanup.

mcquin:
The following changes were made for a saner upgrade to 3D:

* Documentation formatting (`.format` is favored over `% globals`).
* Remove import aliases and optimize imports.
* Remove previous `upgrade_settings`. This module was last updated 7 years ago. Presumably, all active pipelines are up to date.
* Remove support to export images as .MAT, .BMP.
* Remove support to export movies as .AVI, .MOV.
* Update extensions: .TIF -> .TIFF and .JPG -> .JPEG
* Remove support for saving objects. Use ConvertObjectsToImage.
* Remove support for rescaling. Use RescaleIntensity (it works on multi-channel data, too!).
* Remove unimplemented ability to save a figure.
* Update incorrect documentation.


AnneCarpenter:
This is amazing! And refreshing. I laughed out loud about whether it's safe to ditch upgrade_settings. 

I wanted to flag this for super-users @braymp and @bethac07 to ponder whether it'd present a problem:
"Remove support for saving objects. Use ConvertObjectsToImage."

Does that workaround work for the most common cases you're aware of?

braymp:
At the risk to poking the hornet's nest, the removal of saving objects made me spew my coffee. 

This will def. break saving overlapping objects, since each overlapping object was saved into a different plane of a multiframe TIF. 

I'm assuming that an image created with ConvertObjectsToImage + set to "uint16" can be re-loaded as objects by the Input modules?

0x00b1:
> This will def. break saving overlapping objects, since each overlapping object was saved into a different plane of a multiframe TIF.

How were you using this downstream? Is there another application that understands what “overlapping objects” are? It sounds like an abuse of TIFF and would lead to inadvertent corruption.

> I'm assuming that an image created with ConvertObjectsToImage, with the image type as "uint16" can be re-loaded as objects by the Input modules?

Yes

0x00b1:
It’s like a time machine configured to take us on a one-way trip to 2017!

@mcquin I’ll review. But let’s wait for more feedback (e.g. @braymp or @bethac07) before merging.

0x00b1:
A few immediate (possibly controversial) comments:

- We could probably remove the bit-depth setting.
- JPEG should probably have a quality setting (e.g. a slider from 0 to 100%).
- We should consider an option for sRGB color space conversion.
- We should consider (but probably not for long) a transparency setting for PNG.

0x00b1:
I like limiting options to “JPEG, PNG, or TIFF” since they cover user needs:

- Use JPEG if you need compression
- Use PNG if you need lossless
- Use TIFF if you need to save a volume or frames

But I wonder whether we actually **need** JPEG. Do users need a lossy format? If they do, would only PNG and TIFF (with a setting that’d enable compression) be sufficient?

cellprofiler/modules/saveimages.py
```diff
@@ -633,10 +527,10 @@ def do_save_image(self, workspace, filename, pixels, pixel_type,
 
         channel_names - names of the channels (make up names if not present
         '''
-        write_image(filename, pixels, pixel_type,
-                    c=c, z=z, t=t,
-                    size_c=size_c, size_z=size_z, size_t=size_t,
-                    channel_names=channel_names)
+        bioformats.formatwriter.write_image(filename, pixels, pixel_type,
```
0x00b1: I think we can drop BioFormats. 🤗

mcquin:
That's the plan! I think it will take a good bit of reworking. It looks
like movies are being written frame by frame. :/

mcquin:
Let's make some decisions and file separate issues as enhancements or
feature requests!

0x00b1:
@mcquin sounds good!

0x00b1:
Liz says I need to work on my commitment issues, so let’s do this.

litkowsk:
To help with commit(s)ment http://shop.rebekahbrooks.com/category/antique-engagement-rings

bethac07:
Liz, you're my favorite.

braymp:
>How were you using this downstream? Is there another application that understands what “overlapping objects” are? 

CP is able to take in the objects from the multiframe TIF and register them as overlapping. This was convenient when creating a worm model and it might take them several sessions with CP to hand-segment enough images for training. Subsequently, we'd load them and train on them.

>It sounds like an abuse of TIFF and would lead to inadvertent corruption.

I can't imagine it's worse than multiframe stacks being used for goodness-know-what other applications.

mcquin:
SaveWorms?

braymp:
If you want to isolate the functionality w.r.t worms, that seems like a good option. But the Input modules would still need to recognize the IJV format; is that OK?

mcquin:
They still do :)

0x00b1:
> If you want to isolate the functionality w.r.t worms, that seems like a good option. But the Input modules would still need to recognize the IJV format; is that OK?

Yeah. This won’t go away.

0x00b1:
And LOL @litkowsk

AnneCarpenter:
@litkowsk has amazing taste [in jewelry].

PR #2753, opened 2017-05-10
update button in metadata gives feedback

jhung0:
clicking on update button in metadata shows how many rows were found
(similar to NamesAndTypes)


mcquin:
@braymp Metadata correctly reports 0 when there are no images.

PR #2737, opened 2017-05-02
Changes to address issue #2725

VolkerH:
Made some changes to the batch command creation. Now it should work, even if the images are not grouped. Added an additional commandline option for specifying the batch size. 

Retained some of the original code in else clause, I have doubts whether this code can ever be reached.

PR #2755, opened 2017-05-10
--get-batch-commands supports ungrouped pipelines

### LINKED ISSUE #2725: Inconsistent handling of commandline option --get-batch-commands
Dear all,

I am referring to the issue raised by Tischi in this  forum thread:
http://forum.cellprofiler.org/t/simple-batch-file-list/3338/5

I looked into the Cellprofiler sources and found a few issues in `__main__.py`

There are a few issues that I see:

First, with some command line arguments it would make sense to exit once they are processed and not to run a pipeline. In one example, one could exit after generating the batch commands. currently line 82, following 
```
 if options.batch_commands_file is not None:
        get_batch_commands(options.batch_commands_file)
``` 
 I don't see a use case where one would want  process a pipeline when generating a list of batch commands. I think a similar argument could be made for `print_groups`.

Second, `get_batch_commands(filename)` only works for grouped images.
When the images are grouped, the code block following `if m.has_feature(cellprofiler.measurement.IMAGE, cellprofiler.measurement.GROUP_NUMBER):` (currently line 552) is executed and the function returns at the end of the code block.
This `if` clause doesn't have an `else` clause, but as the above code block returns the remaining code of the function is a de-facto `else` clause.  In this remaining code, there is a call to `get_groupings` which returns the grouping `ImageNumber`.

```
(Pdb) m.get_grouping_tags()
['ImageNumber']
(Pdb) m.get_groupings(['ImageNumber'])
[({'ImageNumber': u'1'}, [1]), ({'ImageNumber': u'2'}, [2]), ({'ImageNumber': u\
'3'}, [3]), ({'ImageNumber': u'4'}, [4]), ({'ImageNumber': u'5'}, [5])]
```

I assume that `ImageNumber` is an internally assigned grouping.
Consequently, the batch commands that are created look like this
```

CellProfiler -c -r -p Batch_data.h5 -g ImageNumber=1
CellProfiler -c -r -p Batch_data.h5 -g ImageNumber=2
CellProfiler -c -r -p Batch_data.h5 -g ImageNumber=3
CellProfiler -c -r -p Batch_data.h5 -g ImageNumber=4
CellProfiler -c -r -p Batch_data.h5 -g ImageNumber=5
```

However, CellProfiler doesn't recognize this internally assinged grouping tag `ImageNumber` and creates the following error:

```
$ ./clustercp.sh -c -r  -p Batch_data.h5 -g ImageNumber=1
Times reported are CPU times for each module, not wall-clock time
  File "/g/almfscreen/ALMFCellprofiler/CellProfiler/CellProfiler.py", line 4, in <modu\
le>
    cellprofiler.__main__.main()
  File "/g/almfscreen/ALMFCellprofiler/CellProfiler/cellprofiler/__main__.py", line 15\
5, in main
    run_pipeline_headless(options, args)
  File "/g/almfscreen/ALMFCellprofiler/CellProfiler/cellprofiler/__main__.py", line 70\
8, in run_pipeline_headless
    initial_measurements=initial_measurements
  File "/g/almfscreen/ALMFCellprofiler/CellProfiler/cellprofiler/pipeline.py", line 15\
63, in run
    initial_measurements=measurements):
  File "/g/almfscreen/ALMFCellprofiler/CellProfiler/cellprofiler/pipeline.py", line 16\
81, in run_with_yield
    for group_number, group_index, image_number, closure in __group:
  File "/g/almfscreen/ALMFCellprofiler/CellProfiler/cellprofiler/pipeline.py", line 15\
73, in group
    raise ValueError("The grouping keys specified on the command line (%s) must be the\
 same as those defined by the modules in the pipeline (%s)" % (", ".join(grouping.keys\
()), ", ".join(keys)))
ValueError: The grouping keys specified on the command line (ImageNumber) must be the \
same as those defined by the modules in the pipeline ()
```

I envisage an implementation in which batch commands are simply created based on image number,
e.g. with the `--first-image-set` and `--last-image-set` options.

mcquin:
Resolves #2725 

Author: @VolkerH 
Original PR: https://github.com/CellProfiler/CellProfiler/pull/2737

> Made some changes to the batch command creation. Now it should work, even if the images are not grouped. Added an additional commandline option for specifying the batch size.
> Retained some of the original code in else clause, I have doubts whether this code can ever be reached.


cellprofiler/__main__.py
```diff
@@ -575,18 +587,23 @@ def get_batch_commands(filename):
...
-
-    for grouping in groupings:
-        group_string = ",".join(["%s=%s" % (k, v) for k, v in grouping[0].iteritems()])
+        if len(metadata_tags) == 1 and metadata_tags[0] == 'ImageNumber':
+            for i in range(0, len(image_numbers), n_per_job):
+                first = image_numbers[i]
+                last = image_numbers[min(i+n_per_job-1, len(image_numbers)-1)]
+                print "CellProfiler -c -r -p %s -f %d -l %d" % (filename, first, last)
+        else:
+            # LoadData w/ images grouped by metadata tags
```
mcquin: I learned that this code block is executed when a user creates a pipeline using LoadData and elects to group images by metadata tags.

/cc @VolkerH

VolkerH: Thank you. 
Do I still need to do anything ?

another question regarding the coding style, I notice a lot of empty lines in the CP sources. I actually like them for structuring the code as well. However, I think I read that CP follows the PEP8 style guide. When I enable PEP8 checking in PyCharm it complains about all these empty lines.

mcquin: I re-read https://www.python.org/dev/peps/pep-0008/#blank-lines. We do adhere to

> Surround top-level function and class definitions with two blank lines.
>
> Method definitions inside a class are surrounded by a single blank line.  

These are the most well-defined PEP08 style recommendations on blank lines.

> Use blank lines in functions, sparingly, to indicate logical sections. 

I suppose we could debate about what defines a "logical section", and I choose to leave that decision up to the contributor. It is my opinion that functions should be simple enough to accomplish one task. A function is a sequence of operations. Generally, each operation is its own logical section and rarely do these logical units span more than one line. If there are many clumps, or logical sections, then it's a good indication refactoring may be needed.

mcquin: Nothing further is needed from you, @VolkerH. I'm excited about the improvements you've made to the CLI! Thanks for your contributions!

VolkerH:
thanks

PR #2772, opened 2017-05-22
optional downsampling

### LINKED ISSUE #2768: Watershed shouldn't implicitly rescale before segmenting.
Watershed downscales the images before processing, but this can lead to incorrect segmentation in some cases.  Users wishing to speed up the 3D processing should explicitly rescale instead.

0x00b1:
resolves #2768

### DIFF
--- a/cellprofiler/modules/watershed.py
+++ b/cellprofiler/modules/watershed.py
@@ -13,6 +13,7 @@ import numpy
 import scipy.ndimage
 import skimage.color
 import skimage.feature
+import skimage.filters
 import skimage.measure
 import skimage.morphology
 import skimage.transform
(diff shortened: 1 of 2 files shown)


0x00b1:
cc: @bethac07

cellprofiler/modules/watershed.py
```diff
@@ -116,8 +125,15 @@ def run(self, workspace):
         if self.operation.value == "Distance":
             original_shape = x_data.shape
 
-            if x.volumetric:
-                x_data = skimage.transform.resize(x_data, (original_shape[0], 256, 256), order=0, mode="edge")
+            factor = self.downsample.value
+
+            if factor > 1:
+                if x.volumetric:
+                    factors = (factor, factor, factor)
```
mcquin: The z-dimension should be preserved: `(1, factor, factor)`.

cellprofiler/modules/watershed.py
```diff
@@ -116,8 +125,15 @@ def run(self, workspace):
...
+                else:
+                    factors = (factor, factor)
+
+                x_data = skimage.transform.downscale_local_mean(x_data, factors)
```
mcquin: This operation does not produce a binary image when provided a binary image as input.

PR #2786, opened 2017-06-03
Adds more channels to the display of Split mode in ColorToGray

bethac07:
For rationale, see [RFC](https://github.com/CellProfiler/RFC/pull/2)


mcquin:
Nice @bethac07 ! Can you post a screenshot of an example output?

bethac07:
Here's an 18 channel split (CP's maximum it handles for channels is 19, something I may try to address in a future RFC if it's still true in master, which I actually think it may not be.)  The repeated channel names are not a bug, it's because I was lazy and fudged with the pipeline in a text editor 🙃 

pull2786

mcquin:
Awesome! Looks great! :)

Issue #2255, opened 2016-08-26
Create a new module: ExportImagesPerObject


AnneCarpenter:
This would save an individual image for each object that has been identified in an upstream module. 

This is useful for training deep learning models on individual cell images that have been segmented using CellProfiler. It can also generate the image data needed for data exploration tools designed for imaging flow cytometry, where each cell is imaged separately (though we do not yet know of any requests for this).

INPUT: 
— objects (with XY locations)
— corresponding image(s) to save (should handle RGB or grayscale, masked or not - i.e., any masking or color merging/splitting should happen upstream of this module)

SETTINGS:
— specify the input, as detailed above
— needs an “add another image” option (for multiple grayscale images, for example)
— fixed size tiles (then, specify the pixel X, Y size) or per-object size (each tile fits the object with no blank rows/columns around it) or full size image (the entire size of the original image, with the one little object in the middle - this only makes sense if the input is masked or else it saves the entire original image for every object)
— Rotate major axis of object to the XY axes of the image (two options: limit to 90deg only, vs. any angle). Default = no rotation

DISPLAY:
- nothing really needs to be displayed, but maybe could show “Number of individual images saved to disk: NNN”.  

OUTPUT: files saved to disk, plus maybe we need to save file names to the per-object measurements

AnneCarpenter:
We should consider whether this module should also be allowed to spit individual object images to a remote service rather than save to disk.

braymp:
Would it make sense to split this functionality up across two modules?
* A new TileObjects module which captures the input and settings you suggest (and is reminiscent of TileImage); and 
* SaveImages with a "Tile" option under "Type of input"

The reason for the suggestion is that a module with this function sounds a little esoteric (even though I know the request for this functionality is long-standing). 

On the other hand, carrying out this functionality across two modules might be harder for novice user to find.

AnneCarpenter:
Interesting and clever thought. We should ponder that!

afolarin:
@AnneCarpenter @braymp sounds great. I think this would prove a useful extension of CP. We were looking for a quick way of getting data into a TensorFlow CNN.

0x00b1:
@afolarin This is our motivation too.

0x00b1:
@afolarin And for that reason I think we should export NumPy arrays rather than images.

afolarin:
:+1:

afolarin:
@AnneCarpenter 

[...]

PR #2790, opened 2017-06-12
Changed the CropObjects module to save the cropped object from the original image.

SariNusier:
I believe that issue #2255 is referring to being able to crop the original image for each object. I've made some small changes to the CropObjects module to create an image for each individual object. This crops a rectangle around the object and saves that as an image.

Please let me know if this is something you are looking to implement in the future.


mcquin:
Hi @SariNusier -

The option to export per-object masks is really import and we need to keep that functionality available. I am not against introducing new options to this module which allows users to export images cropped to bounding boxes.

Also, I suggest using [scikit-image region properties](http://scikit-image.org/docs/dev/api/skimage.measure.html#skimage.measure.regionprops) to determine bounding boxes.

SariNusier:
Hi @mcquin 

Thanks for the reply. The thing that I wrote wasn't really meant for a PR, it was just a mod I did as a workaround. If you want I can implement it properly, but I wasn't sure whether you guys were planning on doing it as well.  I would be more than happy to implement it.

mcquin:
Yeah! I think it would be great to have this functionality in addition to the per-object masks. It's not high on my current list of priorities; I'd welcome the contribution! :)

AnneCarpenter:
Hi everyone, I'm a little stuck following the issues/comments for this... did anyone end up submitting a PR for anything to add this functionality? AFAICT we just have savecroppedobjects.py which only saves masks, not the original pixel intensity data.

SariNusier:
Hi @AnneCarpenter. I've submitted a PR #2831 which was adding the functionality to also save the cropped image. I think there was a suggestion left which was to add cropping for 3D objects as well. I remember starting to work on that, but noticing someone else adding the feature. Can you please confirm if this is still needed or if someone else has done it already? I am more than happy to submit the change for 2D (and later on follow-up with the change for 3D).

AnneCarpenter:
Thanks!!

I suspect @0x00b1 might be the only one who knows whether someone else added the feature. I'm pretty sure not.

If not, then your 2D version's PR would be super welcome, 2D is probably ~95% of the use cases.

PR #2794, opened 2017-06-19
Don't raise exception when structuring element size is cleared.

### LINKED ISSUE #2349: Structuring element errors when size is cleared
When I clear the "size" value in the structuring element setting, I get the error:

```
  File "/Users/mcquin/code/CellProfiler/CellProfiler/cellprofiler/gui/moduleview.py", line 692, in on_set_size
    v.size = int(event.GetString())
ValueError: invalid literal for int() with base 10: ''
```

I expect instead no exception but to have a little red "x" next to my module with the blank setting in the pipeline view.

mcquin:
Additionally, validate the size is set and is a positive integer.

Resolves #2349

### DIFF
--- a/cellprofiler/gui/moduleview.py
+++ b/cellprofiler/gui/moduleview.py
@@ -694,7 +694,8 @@ class ModuleView(object):
             )
 
             def on_set_size(event):
-                v.size = int(event.GetString())
+                size = event.GetString()
+                v.size = int(size) if size else ""
                 new_value = v.value_text
                 self.notify(SettingEditedEvent(v, self.__module, new_value, event))
 
--- a/cellprofiler/setting.py
+++ b/cellprofiler/setting.py
@@ -1709,11 +1709,23 @@ class StructuringElement(Setting):
 
     @property
     def size(self):
-        return int(self.value_text.split(",")[1])
+        _, size = self.value_text.split(",")
+
+        return int(size) if size else None
 
     @size.setter
     def size(self, value):
         self.value_text = ",".join((self.shape, str(value)))
+
+    def test_valid(self, pipeline):
+        if self.size is None:
+            raise ValidationError("Missing structuring element size. Please enter a positive integer.", self)
+
+        if self.size <= 0:
+            raise ValidationError(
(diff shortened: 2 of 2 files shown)


cellprofiler/gui/moduleview.py
```diff
@@ -694,7 +694,8 @@ def on_select_shape(event, setting=v, control=shape_control):
             )
 
             def on_set_size(event):
-                v.size = int(event.GetString())
+                size = event.GetString()
+                v.size = int(size) if size else ""
```
0x00b1: Empty string for the alternate? I would assume it would be something like `0` or `None`.

cellprofiler/setting.py
```diff
@@ -1709,12 +1709,24 @@ def shape(self, value):
 
     @property
     def size(self):
-        return int(self.value_text.split(",")[1])
+        _, size = self.value_text.split(",")
```
0x00b1: Ah, I see. You use an empty string because you use `.split`.

It might make more sense to use `None` and check whether `value_text` exists.

mcquin: Unfortunately, `str(None)` becomes `u'None'`.

cellprofiler/gui/moduleview.py
mcquin: Unfortunately, `str(None)` becomes `u'None'`.

cellprofiler/setting.py
0x00b1: Where is `str(None)` called?

mcquin: Additionally, we can't set `value_text` to `None` because it contains the structuring element shape information.

mcquin: @0x00b1 In the size setter (here)[https://github.com/CellProfiler/CellProfiler/blob/a2b23730cd51e9d43816e7fdaf1fcea42205ac2a/cellprofiler/setting.py#L1716-L1718]

```python
@size.setter
def size(self, value):
    self.value_text = ",".join((self.shape, str(value)))
```

0x00b1: Sounds good

PR #2795, opened 2017-06-19
Formatting changes for MeasureImageAreaOccupied

mcquin:
* Remove import aliases and optimize imports
* Use str.format
* Remove '% globals'
* PEP08 compliance


0x00b1:
so fresh

0x00b1:
so clean

PR #2799, opened 2017-06-26
clear border (image segmentation) module

0x00b1:


cellprofiler/modules/clearborder.py
```diff
@@ -0,0 +1,79 @@
...
+
+import skimage.measure
+import skimage.segmentation
+
+import cellprofiler.module
+import cellprofiler.object
+import cellprofiler.setting
+
+
+class ClearBorder(cellprofiler.module.ImageSegmentation):
```
mcquin: I think it makes more sense for the input of this module to be objects, rather than an image. Easy to fix by inheriting from `cellprofiler.module.ObjectProcessing`. :)

tests/modules/test_clearborder.py
```diff
@@ -0,0 +1,57 @@
...
+import skimage.transform
+import skimage.util
+
+import cellprofiler.image
+import cellprofiler.modules.clearborder
+
+instance = cellprofiler.modules.clearborder.ClearBorder()
+
+
+def test_run(image, module, image_set, workspace):
```
mcquin: Does it work on volumes?

0x00b1: Yep. :)

cellprofiler/modules/clearborder.py
0x00b1: Nice!

Issue #2837, opened 2017-08-16
Closing module adds padding to images


karhohs:
The closing module will add padding or a border around an image the size of the structuring element.

![image](https://user-images.githubusercontent.com/156232/29383484-96a976da-829e-11e7-84fb-59ad63fbbb4b.png)

This is not a display issue, because subsequent operations on the output of Closing treats this border as part of the image. For instance, Gaussian smoothing blurs the border.

![image](https://user-images.githubusercontent.com/156232/29383546-d1126e62-829e-11e7-8412-aa6f7f3322ca.png)

0x00b1:
Hi, @karhohs. Would you mind sharing the original image?

karhohs:
No problem. 
[issue_2837.zip](https://github.com/CellProfiler/CellProfiler/files/1231972/issue_2837.zip)

mcquin:
@karhohs We use `skimage.measure.closing` as the closing operation. I believe padding the image border with 0-values is the default behavior of `skimage.morphology.closing`.

karhohs:
@mcquin I believe the [skimage.morphology.closing example](http://scikit-image.org/docs/dev/api/skimage.morphology.html#skimage.morphology.closing) does not show this padding.

in:
```
[[0, 0, 0, 0, 0],
[0, 0, 0, 0, 0],
[1, 1, 0, 1, 1],
[0, 0, 0, 0, 0],
[0, 0, 0, 0, 0]]
```

out:
```
[[0, 0, 0, 0, 0],
[0, 0, 0, 0, 0],
[1, 1, 1, 1, 1],
[0, 0, 0, 0, 0],
[0, 0, 0, 0, 0]]
```

karhohs:
Ah ha! The issue is not related to *skimage.morphology.closing*, but instead ***skimage.morphology.binary_closing***.

https://gist.github.com/karhohs/bad263764e0b3d4dd175204d53a247fc

mcquin:
Interesting! I would expect the functions to product identical output.

There's not much we can do here, except to use `closing` in favor of `binary_closing`.

0x00b1:
@mcquin I was intrigued by this one too!

karhohs:
I updated the [gist](https://gist.github.com/karhohs/bad263764e0b3d4dd175204d53a247fc) to show that all binary operations have this border effect, which differs from the non-binary morphology operations.

karhohs:
[...]

PR #2839, opened 2017-08-18
Issues/2837

karhohs:
The *binary* versions or dilation, erosion, opening, and closing within skimage.morphology introduced a border of *False* values around the perimeter of an image. This is in contrast to the *non-binary* versions of dilation, erosion, opening, and closing that do not introduce this artifact. Since the non-binary versions of these morphological operations still work on boolean images the binary versions were replaced.

This is related to https://github.com/CellProfiler/CellProfiler/issues/2837

### DIFF
--- a/cellprofiler/modules/closing.py
+++ b/cellprofiler/modules/closing.py
@@ -54,7 +54,9 @@ class Closing(cellprofiler.module.ImageProcessing):
             raise NotImplementedError("A 3D structuring element cannot be applied to a 2D image.")
         else:
             if x.pixel_data.dtype == numpy.bool:
-                self.function = skimage.morphology.binary_closing
+                # issue #2837
+                # self.function = skimage.morphology.binary_closing
+                self.function = skimage.morphology.closing
             else:
                 self.function = skimage.morphology.closing
 
@@ -66,7 +68,9 @@ def planewise_morphology_closing(x_data, structuring_element):
 
     for index, plane in enumerate(x_data):
         if x_data.dtype == numpy.bool:
-            y_data[index] = skimage.morphology.binary_closing(plane, structuring_element)
+            # issue #2837
+            # y_data[index] = skimage.morphology.binary_closing(plane, structuring_element)
+            y_data[index] = skimage.morphology.closing(plane, structuring_element)
         else:
             y_data[index] = skimage.morphology.closing(plane, structuring_element)
 
--- a/cellprofiler/modules/closing.py
+++ b/cellprofiler/modules/closing.py
@@ -71,12 +71,6 @@ def planewise_morphology_closing(x_data, structuring_element):
 
     for index, plane in enumerate(x_data):
 
-        if x_data.dtype == numpy.bool:
-
-            y_data[index] = skimage.morphology.closing(plane, structuring_element)
-
-        else:
-
-            y_data[index] = skimage.morphology.closing(plane, structuring_element)
+        y_data[index] = skimage.morphology.closing(plane, structuring_element)
 
     return y_data
--- a/cellprofiler/modules/closing.py
+++ b/cellprofiler/modules/closing.py
@@ -60,13 +60,7 @@ class Closing(cellprofiler.module.ImageProcessing):
 
         else:
 
-            if x.pixel_data.dtype == numpy.bool:
-
-                self.function = skimage.morphology.closing
-
-            else:
-
-                self.function = skimage.morphology.closing
+            self.function = skimage.morphology.closing
 
         super(Closing, self).run(workspace)
 
--- a/cellprofiler/modules/closing.py
+++ b/cellprofiler/modules/closing.py
@@ -45,33 +45,44 @@ class Closing(cellprofiler.module.ImageProcessing):
     def run(self, workspace):
 
         x = workspace.image_set.get_image(self.x_name.value)
+
         is_strel_2d = self.structuring_element.value.ndim == 2
+
         is_img_2d = x.pixel_data.ndim == 2
 
         if is_strel_2d and not is_img_2d:
+
             self.function = planewise_morphology_closing
+
         elif not is_strel_2d and is_img_2d:
+
             raise NotImplementedError("A 3D structuring element cannot be applied to a 2D image.")
+
         else:
+
             if x.pixel_data.dtype == numpy.bool:
-                # issue #2837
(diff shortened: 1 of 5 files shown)


0x00b1:
Hey, @karhohs! Thanks! This looks good. However, would you mind removing the comments? We don't need them. It looks like you need to change the expected values in the filterobjects tests. See the Travis CI output:

https://travis-ci.org/CellProfiler/CellProfiler/builds/266089055?utm_source=github_status&utm_medium=notification

karhohs:
@0x00b1 I think filterobjects.py might have been using the binary_erode bug as a feature. I've been looking into the expected results and something seems off. Is there a reason [skimage.segmentation.clear_border](http://scikit-image.org/docs/dev/api/skimage.segmentation.html?highlight=border#clear-border) isn't being implemented here, line 833 in filterobjects.py?

0x00b1:
@karhohs No reason. It sounds like clear_borders would fix the problem.

cellprofiler/modules/closing.py
```diff
@@ -45,29 +45,44 @@ def visible_settings(self):
...
+
             self.function = planewise_morphology_closing
+
         elif not is_strel_2d and is_img_2d:
+
             raise NotImplementedError("A 3D structuring element cannot be applied to a 2D image.")
+
         else:
+
             if x.pixel_data.dtype == numpy.bool:
```
mcquin: This `if ... else ...` is redundant. Set `self.function = skimage.morphology.closing` regardless of data type.

cellprofiler/modules/closing.py
```diff
@@ -45,29 +45,44 @@ def visible_settings(self):
...
         super(Closing, self).run(workspace)
 
 
 def planewise_morphology_closing(x_data, structuring_element):
+
     y_data = numpy.zeros_like(x_data)
 
     for index, plane in enumerate(x_data):
+
         if x_data.dtype == numpy.bool:
```
mcquin: Redundant `if ... else ...`. Use `skimage.morphology.closing` regardless of data type.

cellprofiler/modules/dilation.py
```diff
@@ -39,30 +39,46 @@ def visible_settings(self):
...
+
             self.function = planewise_morphology_dilation
+
         elif not is_strel_2d and is_img_2d:
+
             raise NotImplementedError("A 3D structuring element cannot be applied to a 2D image.")
+
         else:
+
             if x.pixel_data.dtype == numpy.bool:
```
mcquin: Use `skimage.morphology.dilation` regardless of data type.

cellprofiler/modules/dilation.py
```diff
@@ -39,30 +39,46 @@ def visible_settings(self):
...
         super(Dilation, self).run(workspace)
 
 
 def planewise_morphology_dilation(x_data, structuring_element):
+
     y_data = numpy.zeros_like(x_data)
 
     for index, plane in enumerate(x_data):
+
         if x_data.dtype == numpy.bool:
```
mcquin: Use `skimage.morphology.dilation` regardless of data type.

cellprofiler/modules/erosion.py
```diff
@@ -39,30 +39,46 @@ def visible_settings(self):
...
+
             self.function = planewise_morphology_erosion
+
         elif not is_strel_2d and is_img_2d:
+
             raise NotImplementedError("A 3D structuring element cannot be applied to a 2D image.")
+
         else:
+
             if x.pixel_data.dtype == numpy.bool:
```
mcquin: Use `skimage.morphology.erosion` regardless of data type.

cellprofiler/modules/erosion.py
```diff
@@ -39,30 +39,46 @@ def visible_settings(self):
...
         super(Erosion, self).run(workspace)
 
 
 def planewise_morphology_erosion(x_data, structuring_element):
+
     y_data = numpy.zeros_like(x_data)
 
     for index, plane in enumerate(x_data):
+
         if x_data.dtype == numpy.bool:
```
mcquin: Use `skimage.morphology.erosion` regardless of data type.

cellprofiler/modules/opening.py
```diff
@@ -40,30 +40,46 @@ def visible_settings(self):
...
+
             self.function = planewise_morphology_opening
+
         elif not is_strel_2d and is_img_2d:
+
             raise NotImplementedError("A 3D structuring element cannot be applied to a 2D image.")
+
         else:
+
             if x.pixel_data.dtype == numpy.bool:
```
mcquin: Use `skimage.morphology.opening` regardless of data type.

cellprofiler/modules/opening.py
```diff
@@ -40,30 +40,46 @@ def visible_settings(self):
...
         super(Opening, self).run(workspace)
 
 
 def planewise_morphology_opening(x_data, structuring_element):
+
     y_data = numpy.zeros_like(x_data)
 
     for index, plane in enumerate(x_data):
+
         if x_data.dtype == numpy.bool:
```
mcquin: Use `skimage.morphology.erosion` regardless of data type.

mcquin:
FWIW: `scipy.ndimage.binary_x` has the same effect as `skimage.morphology.binary_x`. Likewise, `scipy.ndimage.gray_x` and `skimage.morphology.x`. This makes me think that the behavior of `skimage.morphology.binary_x` is a bug.

If the added border to a `skimage.morphology.binary_x` operation is undesired, then we should use `skimage.morphology.x` in its place. However, I would not suggest modifying the binary erosion trick in FilterObjects. It's very useful for quickly computing pixels touching the image border.

karhohs:
@mcquin Sounds good. I'll remove the filter objects edits. Do you agree that we should use `skimage.morphology.x` over `skimage.morphology.binary_x`? When I'm working with thresholded images, I would like the behavior of `skimage.morphology.x`.

mcquin:
Your call! If you like `skimage.morphology.x` better than `skimage.morphology.binary_x`, and you think other users will feel the similarly, then the correct behavior for us is `skimage.morphology.x`.

karhohs:
If no one else disagrees, let's go with my preference. I'm betting it is the generally preferred behavior.

mcquin:
Go for it, @karhohs :)

mcquin:
Nice work, dude!

PR #2841, opened 2017-08-21
Volumetric support for MeasureTexture

### LINKED ISSUE #2454: MeasureTexture should support three-dimensional images

mcquin:
Resolves #2454

### DIFF
--- a/cellprofiler/modules/measuretexture.py
+++ b/cellprofiler/modules/measuretexture.py
@@ -76,15 +76,6 @@ Available measurements
    -  *45:* Diagonal
    -  *135:* Anti-diagonal
 
--  *Gabor “wavelet” features:* These features are similar to wavelet
-   features, and they are obtained by applying so-called Gabor filters
-   to the image. The Gabor filters measure the frequency content in
-   different orientations. They are very similar to wavelets, and in the
-   current context they work exactly as wavelets, but they are not
-   wavelets by a strict mathematical definition. The Gabor features
-   detect correlated bands of intensities, for instance, images of
-   Venetian blinds would have high scores in the horizontal orientation.
-
 .. _here: http://murphylab.web.cmu.edu/publications/boland/boland_node26.html
 
 Technical notes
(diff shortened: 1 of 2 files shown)


cellprofiler/modules/measuretexture.py
```diff
@@ -710,201 +539,111 @@ def display(self, workspace, figure):
...
+                if image.has_mask:
+                    mask, m2 = cellprofiler.object.size_similarly(labels, image.mask)
+                    mask[~m2] = False
                 else:
-                    mask, m2 = cellprofiler.object.size_similarly(labels, mask)
+                    mask = m1
 
-                    mask[~m2] = False
+        pixel_data[~mask] = 0
+        # mahotas.features.haralick bricks itself when provided a dtype larger than uint8 (version 1.4.3)
```
0x00b1: Is this true regardless of your available system memory? http://mahotas.readthedocs.io/en/latest/faq.html#i-ran-out-of-memory-computing-haralick-features-on-16-bit-images-is-it-not-supported

mcquin:
In the issue, Anne suggests it's fine to drop Gabor.

cellprofiler/modules/measuretexture.py
0x00b1: You could also use scikit-image to compute the Haralick features (from the GLCM).

mcquin: How much memory is required is unclear. However, I will take the 12-bit range allowance into consideration.

Thanks for the skimage suggestion. The mahotas function is much easier to use. I'll stick with it.

0x00b1: > How much memory is required is unclear. However, I will take the 12-bit range allowance into consideration.

If I’m reading the FAQ correctly, it’s `(ngrey, ngrey)` so 16-bit is `(2^16 − 1, 2^16 − 1)`.

0x00b1:
Nice work dude. I be this feels good. 👍

PR #2842, opened 2017-08-21
Warn when selecting an incompatible structuring element.

### LINKED ISSUE #2351: Structuring element doesn’t return an error when selecting a three-dimensional shape for a two-dimensional image

mcquin:
Resolves #2351 

Generate a validation warning when a structuring element of an incompatible dimensionality is selected. E.g., when a 3D structuring element is configured for a 2D pipeline:
screen shot 2017-08-21 at 4 00 39 pm

...or a 2D structuring element is selected for a 3D pipeline and the module does not support planewise operations.
screen shot 2017-08-21 at 4 00 12 pm

### DIFF
--- a/cellprofiler/gui/moduleview.py
+++ b/cellprofiler/gui/moduleview.py
@@ -2106,9 +2106,12 @@ class ModuleView(object):
     def __on_radiobox_change(self, event, setting, control):
         if not self.__handle_change:
             return
+
         setting.on_event_fired(control.GetStringSelection() == "Yes")
-        self.on_value_change(
-            setting, control, control.GetStringSelection(), event)
+
+        self.on_value_change(setting, control, control.GetStringSelection(), event)
+
+        self.__module.on_setting_changed(setting, self.__pipeline)
 
     def __on_combobox_change(self, event, setting, control):
         if not self.__handle_change:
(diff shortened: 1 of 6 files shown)


cellprofiler/gui/moduleview.py
```diff
@@ -4464,18 +4467,21 @@ def __init__(self, pipeline, module, callback):
...
-        d = weakref.WeakValueDictionary()
-        setattr(request_pipeline_cache, "d", d)
-    settings_hash = pipeline.settings_hash()
-    result = d.get(settings_hash)
-    if result is None:
-        result = d[settings_hash] = pipeline.copy(False)
-    return result
+#
+# Cacheing is not compatible with the current pattern of sharing state between GUI components. E.g., the pipeline object
+# stored in the ModuleView constructor will never be updated, yet it is used to validate current state.
```
mcquin: I decided to leave this as a comment, but I am not sure how useful it will be in the future. This change was critical to solving the problem where the pipeline would be updated but the change would not propagate through the GUI. 

I am not certain how critical pipeline caching is (especially for large pipelines). It is probably worth investigating. However, I think the larger issue is not the size of the pipeline but the number of generated validation requests.

PR #2854, opened 2017-08-24
PyInstaller (Windows)

0x00b1:
Use:

```sh
pyinstaller.exe CellProfiler.spec
cp C:\Python27\Lib\site-packages\zmq\libzmq.pyd .\dist\CellProfiler\
```

### DIFF
--- a/CellProfiler.spec
+++ b/CellProfiler.spec
@@ -31,7 +31,7 @@ hiddenimports += [
 
 a = Analysis(
     [
-        'CellProfiler.py'
+        "CellProfiler.py"
     ],
     binaries=[],
     cipher=block_cipher,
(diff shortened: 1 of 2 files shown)


CellProfiler.spec
```diff
@@ -67,11 +63,11 @@ exe = EXE(
...
     a.scripts,
     # options,
-    console=False,
+    console=True,
     debug=False,
     exclude_binaries=True,
-    icon=os.path.join("cellprofiler", "data", "CellProfilerIcon.ico"),
-    name='CellProfiler-App',
+    icon=os.path.join("cellprofiler", "data", "images", "CellProfilerIcon.ico"),
+    name="cp",
```
mcquin: What's the purpose of the `name` field? 

`"cp"` seems a little terse. Is `"cellprofiler"` more appropriate?

0x00b1: It’s the executable inside the distribution. Because the distribution includes the CellProfiler package (`/cellprofiler`) I cannot use the name `cellprofiler`. In practice this is fine, since the application name uses the name from the bundled distribution (`CellProfiler`).

mcquin: Ok! Makes sense. :)

PR #2863, opened 2017-08-25
[WIP] Build help with sphinx

### LINKED ISSUE #2840: help.py should be updatable with Sphinx
The main GUI help at `cellprofiler/gui/help.py` would be great if we could include in the Sphinx documentation upgrade.

mcquin:
Resolves #2840


mcquin:
OK -- All the help I could find is updated to reST. It may need some formatting help. Could you please take a look @bethac07 ?

It renders in the GUI just fine. I am having trouble getting the images to link correctly in Sphinx. It seems to only accept paths relative to the source directory: http://www.sphinx-doc.org/en/stable/rest.html#images

> When used within Sphinx, the file name given (here gnu.png) must either be relative to the source file, or absolute which means that they are relative to the top source directory. For example, the file sketch/spam.rst could refer to the image images/spam.png as ../images/spam.png or /images/spam.png.

PR #2888, opened 2017-08-28
Reformatted CorrectIlluminationApply

vchernys:


mcquin:
Nice! Indentation in setting text is off. Settings with documentation spanning multiple lines should have the format:

```python
    def create_settings():
        self.my_setting = cellprofiler.setting.MySetting(
            "setting text",
            setting_value,
            doc="""\
This setting does a particular thing.
When it has this value it does this.
When it has that value it does that.
"""
        )
```

cellprofiler/modules/correctilluminationapply.py
```diff
@@ -1,7 +1,14 @@
 # coding=utf-8
 
 """
+CorrectIlluminationApply
+========================
+
+<<<<<<< HEAD
```
mcquin: @vchernys You'll need to remove lines like these. They indicate merge conflicts:

```
<<<<<<<< HEAD
a version
=======
another version
>>>>>> COMMIT SHA
```

PR #2900, opened 2017-08-30
corrected a few indentations

vchernys:


cellprofiler/modules/measureobjectintensity.py
```diff
@@ -1,7 +1,10 @@
 # coding=utf-8
 
 """
-**Measure Object Intensity** measures several intensity features for
+MeasureObjectIntensity
+========================
```
mcquin: The `===` line should be the same length as the line above.

PR #2901, opened 2017-08-30
changed format

vchernys:


cellprofiler/modules/measureimageintensity.py
```diff
@@ -109,22 +112,25 @@ def add_image_measurement(self, can_remove=True):
...
+Select *%(YES)s* to measure only those pixels within an object of choice.
+""" % globals()))
 
         group.append("object_name", cps.ObjectNameSubscriber(
-                "Select the input objects", cps.NONE, doc='''
-            <i>(Used only when measuring intensity from area enclosed by objects)</i><br>
-            Select the objects that the intensity will be aggregated within. The intensity measurement will be
-            restricted to the pixels within these objects.'''))
+                "Select the input objects", cps.NONE, doc="""\               
+*(Used only when measuring intensity from area enclosed by objects)*
```
mcquin: Add a newline between *(Used only when...)* and the next line to add a line break in the rendered documentation. Otherwise, all text appears on the same line.

Issue #2909, opened 2017-08-30
Workspace Viewer should be resurrected.


bethac07:
Please and thank you!

braymp:
Also see #2748.

mcquin:
https://github.com/CellProfiler/CellProfiler/pull/2911
https://github.com/CellProfiler/CellProfiler/pull/2957

bethac07:
A request from [the forum](https://forum.image.sc/t/announcing-cellprofiler-3-1-9/29401/9?u=bcimini) for when this returns - make the viewer "pickle-a-ble", ala the workspace files in CPA, which you can save some file with workspace viewer properties (ie which images, objects, and measurements are loaded)

PR #2911, opened 2017-08-30
removed workspace viewer button

0x00b1:


0x00b1:
> Additionally, we should comment out/remove references to the workspace viewer in the documentation to prevent confusion.

done

0x00b1:
> Maybe it would be better to comment this code out, so it will be easy to add back in #2909.

What if I just update the issue with the commit for the changeset?

cellprofiler/gui/help.py
```diff
@@ -1493,186 +1493,6 @@
...
-object's pixels. Each object is assigned a color using the default color map
-initially. You can choose the color map by
-selecting <i>Color</i> from the objects' menu and choosing one of the available
-color maps. You can change the transparency of the overlay by choosing
-<i>Alpha</i> from the objects' menu.</li>
-</ul></p>
-
-""" % globals()
-
-WV_FIGURE_HELP = tuple(list(FIGURE_HELP) + [(
```
mcquin: Removing `WV_FIGURE_HELP` will break this: https://github.com/CellProfiler/CellProfiler/blob/e9170b747a26a1d546171cbffb6d89604b935e02/cellprofiler/gui/viewworkspace.py#L229

mcquin:
> What if I update the issue with the commit for the changes?

That's fine, too.

cellprofiler/gui/help.py
0x00b1: Fixed

mcquin:
@0x00b1 This line should also be removed: https://github.com/CellProfiler/CellProfiler/blob/e9170b747a26a1d546171cbffb6d89604b935e02/cellprofiler/gui/cpframe.py#L635

PR #2931, opened 2017-08-31
Load and export NPY arrays

### LINKED ISSUE #2822: Export illumination correction images as NPY arrays which can be loaded in NamesAndTypes
When trying to save an image, using the SaveImages module, created by the CorrectIlluminationCalculate module, I observe an error that states...

`Cannot save image, because values are are not between -1 and 1`.

I am trying to save a 64-bit floating point TIFF. The documentation for the CorrectIlluminationCalculate module states that the values are intentionally calculated to be >1 (so that the corrected image is still between 0 and 1 after division).

mcquin:
Resolves #2822


mcquin:
Fixing Travis errors.

@karhohs Has verified this works for him.

PR #2939, opened 2017-08-31
Edit documentation FlagImage revised

minh-doan:


cellprofiler/modules/flagimage.py
```diff
@@ -1,6 +1,9 @@
 # coding=utf-8
 
 """
+FlagImage
+=========
+
 **Flag Image** allows you to flag an image based on properties that you
```
bethac07: Please change **Flag Image** to **FlagImage**

PR #2941, opened 2017-08-31
Restructured text and grammatical edits

N3llz:


cellprofiler/modules/makeprojection.py
```diff
@@ -64,73 +66,75 @@ class MakeProjection(cpm.Module):
...
         self.frequency = cps.Float(
-                "Frequency", 6.0, minval=1.0, doc="""
-            <i>(Used only if %(P_POWER)s is selected as the projection method)</i><br>
-            This setting controls the frequency at which the power
-            is measured. A frequency of 2 will respond most strongly to
-            pixels that alternate between dark and light in successive
-            z-stack slices. A frequency of N will respond most strongly
-            to pixels whose brightness cycle every N slices.""" % globals())
+                "Frequency", 6.0, minval=1.0, doc="""\
+*(Used only if %(P_POWER)s is selected as the projection method)*
```
bethac07: Please put an empty line between the *(Used only...)* line and the next one.

cellprofiler/modules/makeprojection.py
```diff
@@ -1,11 +1,13 @@
...
+==============
+**MakeProjection** combines several two-dimensional images of the same
+field of view by performing a mathematical operation
 upon the pixel values at each pixel position.
 
-This module combines a set of images by performing a mathematic
+This module combines a set of images by performing a mathematical
 operation of your choice at each pixel position; please refer to the
 settings help for more information on the available operations. The
 process of averaging or summing a Z-stack (3D image stack) is known as
```
bethac07: At the end of this paragraph, for sake of clarification please add:

**MakeProjection** is meant to be used only on single image Z-slices across image cycles; it will not work on images that have been loaded as 3D volumes in **NamesAndTypes**.

cellprofiler/modules/makeprojection.py
mcquin: Add quotes around setting text: `"%(P_POWER)s"`

PR #2945, opened 2017-08-31
Updates to filterobjects

bethac07:


bethac07:
Something weird happened here.  Closing and repulling.

PR #2947, opened 2017-08-31
Load versioned example pipeline.

### LINKED ISSUE #2943: Load an ExampleFly pipeline matching the CellProfiler version.
When running CellProfiler stable or beta, an example pipeline corresponding to that version should be loaded. This will prevent "Your pipeline was saved with X but you are running Y" messages. CellProfiler should grab the latest example pipeline when running from nightly or source.

We will support this feature starting with 3.0 beta release.

We can use GitHub tags in CellProfiler/examples to host versioned pipelines.

mcquin:
Resolves #2943

Loads a version of the ExampleFly pipeline corresponding to the running
version of CellProfiler. If no matching pipeline is found, the pipeline
is run from master.

Assumes CellProfiler/examples is tagged with the a CellProfiler version
string. When releasing a new version of CellProfiler, remember to update
and tag CellProfiler/examples with the new version.

### DIFF
--- a/cellprofiler/gui/html/content.py
+++ b/cellprofiler/gui/html/content.py
@@ -148,7 +148,7 @@ startup_main = '''<html>
 </tr>
 <tr>
     <td width="1">&nbsp;</td>
-    <td colspan="2"><a href="loadexample:http://cellprofiler.org/ExampleFlyImages/ExampleFlyURL.cppipe">Load</a> an example pipeline, then click on the "Analyze Images" button.</td>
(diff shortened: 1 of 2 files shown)


mcquin:
> This seems like a special case, why don't we just keep the pipeline in cellprofiler/data/ and load it with pkg_resources?

@0x00b1 If we did that, then we'd need to maintain multiple versions of that pipeline. One in this repository, and one in CellProfiler/examples.

PR #2951, opened 2017-08-31
Documentation morph

bethac07:
Nested tables in restructured text don't seem to work that well, so I had to do a list-table with a single table for each function (rather than before and after).  I also had to redo all the tables by hand, so people should check them to make sure they're right.


cellprofiler/modules/morph.py
```diff
@@ -884,35 +320,34 @@ def add_function(self, can_remove=True):
...
-            <p>Select <i>%(YES)s</i> to rescale the transformed values to lie between 0 and 1.
-            This is the option to use if the distance transformed image is to be used
-            for thresholding by an <b>Identify</b> module or the like, which assumes
-            a 0-1 scaling.</p>
-            <p>Select <i>%(NO)s</i> to leave the values in absolute pixel units.
-            This useful in cases where the actual pixel distances are to be used
-            downstream as input for a measurement module.</p>
-            """ % globals()))
+                "Rescale values from 0 to 1?", True, doc="""\
+*(Used only for the %(F_DISTANCE)s operation).*
```
mcquin: `"%(F_DISTANCE)s"`

cellprofiler/modules/morph.py
```diff
@@ -852,30 +287,31 @@ class Morph(cpm.Module):
...
+the previous operation the number of times you select before applying
+the operation added by this button.""")
 
         self.functions = []
         self.add_function(can_remove=False)
 
     CUSTOM_REPEATS_TEXT = "Repetition number"
-    CUSTOM_REPEATS_DOC = "<i>(Used only if Custom selected)</i><br>Enter the number of times to repeat the operation"
+    CUSTOM_REPEATS_DOC = """\
+*(Used only if Custom selected)*
```
mcquin: `"Custom"`

PR #2952, opened 2017-08-31
Resize format change

vchernys:


cellprofiler/modules/resize.py
```diff
@@ -93,45 +86,40 @@ def create_settings(self):
...
 
         self.specific_image = cellprofiler.setting.ImageNameSubscriber(
             "Select the image with the desired dimensions",
             cellprofiler.setting.NONE,
-            doc=""""
-            <i>(Used only if resizing by specifying desired final dimensions using an image)</i><br>
-            The input image will be resized to the dimensions of the specified image.
-            """
-        )
+            doc=""""\
```
mcquin: Extra quote. There should only be 3.

cellprofiler/modules/resize.py
```diff
@@ -4,7 +4,8 @@
 Resize
 ======
 
-**Resize** resizes images (changes their resolution).
+**Resize** resizes images (changes their resolution). This module is
+compatible for both 2D and volumetric image analysis.
```
bethac07: Can you change this sentence to "compatible with 2D and 3D/volumetric images."

PR #2964, opened 2017-09-01
Doc/contributing

vzloy:
Dear Authors,

Would it be possible to have short mathematical descriptions of all modules to avoid (or decrease) the experiments with different parameters of modules; to make the optimisation more intelligent? For example, for Otsu method, we have on Wikipedia https://en.wikipedia.org/wiki/Otsu's_method . It would be grate to have at least links related to different methods or research papers; in optimum - short formal introduction to the methods based ZFC set theory, so everybody with mathematical background can predict the effect of different parameters.

Cheers,
Volodymyr

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,9 +2,11 @@
 
 Hello! We're glad you're interested in contributing to CellProfiler.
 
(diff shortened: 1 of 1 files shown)


0x00b1:
Hi, did you intend to make an issue? You created a pull request without any changes.

PR #2992, opened 2017-09-02
add padding between subplots (WIP)

### LINKED ISSUE #2899: Display padding between elements doesn't take titles into account in 3.0
Using today's nightly on Windows 10; I've verified it happens in other modules too (every one I've tried so far) not just CorrectIlluminationCalculate:

2.2:

display_gauss_22

3.0:

display_gauss_30

0x00b1:
resolves #2899 (???)


mcquin:
Group note: we need to decide if this is an improvement.

PR #3025, opened 2017-09-05
Change the output message when naming images/data in module.py

### LINKED ISSUE #3009: Avoid using term "output" when naming images/data
Because in each instance below it is easy to read "image output" as an adjective-noun pair instead of noun-verb pair - and once you've done that the rest of the sentence cannot be parsed and is confusing.

In CellProfiler/cellprofiler/module.py
Change this line: "Enter the name you want to call the image output by this module."
to this: "Enter the name you want to call the image produced by this module."

and change this line: "Enter the name you want to call the object output by this module."
to this: "Enter the name you want to call the object produced by this module."

In CellProfiler/cellprofiler/modules/loaddata.py
Change this line:    '''Return column definitions for measurements output by this module'''
to this:      '''Return column definitions for measurements produced by this module'''

DuaaAli:
Resolved #3009

### DIFF
--- a/cellprofiler/module.py
+++ b/cellprofiler/module.py
@@ -909,7 +909,7 @@ class ImageProcessing(Module):
         self.y_name = cellprofiler.setting.ImageNameProvider(
             "Name the output image",
             self.__class__.__name__,
-            doc="Enter the name you want to call the image output by this module."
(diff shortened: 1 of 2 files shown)


mcquin:
Thanks, @DuaaAli !! Looks great.

PR #3034, opened 2017-09-05
ShrinkObjectsToPoints, a 2d/3d version for shrinking objects to a point 

karhohs:
ExpandOrShrinkObjects has the functionality to shrink a set of objects to a single point. Often an image is heavily processed before objects are identified using a watershed based method implemented in IdentifyPrimaryObjects or Watershed. Such processing can blur or warp boundaries between objects, resulting in a segmentation that does not reflect the original image with exactness. Seeding with single points upon the original image can yield a more accurate segmentation. This approach was possible in 2D, but not in 3D since ExpandOrShrinkObjects is not 3D compatible. This pull request is being made in the main CellProfiler project, as opposed to a plugin, because I think it could be of value for many 3D segmentation pipelines.


cellprofiler/modules/shrinkobjectstopoints.py
```diff
@@ -0,0 +1,128 @@
...
+
+import cellprofiler.image
+import cellprofiler.module
+import cellprofiler.setting
+import numpy
+import skimage.measure
+
+
+class ShrinkObjectsToPoints(cellprofiler.module.ObjectProcessing):
+
```
mcquin: No space is necessary between the class/method/function declaration and the next line.

E.g., 
```python
class ShrinkObjectsToPoints(cellprofiler.module.ObjectProcessing):
    module_name = "ShrinkObjectsToPoints"
```

```python
def create_settings(self):
    super(ShrinkObjectsToPoints, self).create_settings()
```

cellprofiler/modules/shrinkobjectstopoints.py
```diff
@@ -0,0 +1,128 @@
...
+    category = "Advanced"
+
+    variable_revision_number = 1
+
+    def create_settings(self):
+
+        super(ShrinkObjectsToPoints, self).create_settings()
+
+        self.x_name.text = "Input objects"
```
mcquin: This is already defined as

```python
self.x_name = cellprofiler.setting.ObjectNameSubscriber(
    "Select the input object",
    doc="Select the object you want to use."
)
```

Perhaps only `self.x_name.doc` needs updating?

cellprofiler/modules/shrinkobjectstopoints.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+    def visible_settings(self):
+
+        __settings__ = super(ShrinkObjectsToPoints, self).visible_settings()
+
+        return __settings__
+
+
+    def display(self, workspace, figure):
```
mcquin: You can use the super class' `display` method. Change these:

`workspace.display_data.x_data = input_objects.segmented`
`workspace.display_data.y_data = output_objects.segmented`

cellprofiler/modules/shrinkobjectstopoints.py
```diff
@@ -0,0 +1,128 @@
...
+    def run(self, workspace):
+
+        input_objects = workspace.object_set.get_objects(self.x_name.value)
+
+        output_objects = cellprofiler.object.Objects()
+
+        output_objects.segmented = self.shrink_ray(input_objects.segmented)
+
+        if input_objects.has_small_removed_segmented:
+
```
mcquin: A blank link is not needed between `if/else` and the following line. E.g.,

```python
if input_objects.has_small_removed_segmented:
    output_objects.small_removed_segmented = self.shrink_ray(input_objects.small_removed_segmented)
```

cellprofiler/modules/shrinkobjectstopoints.py
```diff
@@ -0,0 +1,128 @@
...
+        input_centroids = [numpy.int_(obj["centroid"]) for obj in input_props]
+
+        output_segmented = numpy.zeros_like(label_image)
+
+        for ind, arr in enumerate(input_centroids):
+
+            output_segmented[tuple(arr)] = ind + 1
+
+        return output_segmented
```
mcquin: A blank line is required at the end of the file.

cellprofiler/modules/shrinkobjectstopoints.py
```diff
@@ -0,0 +1,128 @@
...
+        if self.show_window:
+
+            workspace.display_data.input_objects_segmented = input_objects.segmented
+
+            workspace.display_data.output_objects_segmented = output_objects.segmented
+
+            workspace.display_data.dimensions = input_objects.dimensions
+
+
+    def shrink_ray(self, label_image):
```
mcquin: I love this method name, it's really cute! However, good method names should be self documenting. Something boring like `find_centroids` would be more obvious.

tests/modules/test_shrinkobjectstopoints.py
```diff
@@ -0,0 +1,101 @@
...
+    labels[0:5, 0:8, 12:18] = 2
+
+    labels[4:9, 12:18, 0:8] = 3
+
+    labels[1:8, 12:20, 12:20] = 4
+
+    return labels
+
+
+def test_resize_by_factor_shrink_image_labels(image_labels, module, object_set, objects, workspace):
```
mcquin: I don't understand `resize_by_factor` here. Should the test name be `test_shrink_image_labels`?

PR #3066, opened 2017-09-07
Focus on running module in test mode.

### LINKED ISSUE #3062: Highlight the current module in TestMode
It would be nice to show users which module is about to run in TestMode. For example, if we paused TestMode on IdentifySecondaryObjects, the module name could be highlighted. Alternatively, an icon to the left of the pause icon could be added to indicate which module is about to run next. Another suggestion is to make the module text bold.

screen shot 2017-09-07 at 11 54 13 am

mcquin:
Resolves #3062 

@DuaaAli and I figured out how to improve Test Mode visualization. The "running module" is the module that will run the next time "Step" is pressed. That module receives a blue/gray rectangle around it. You can select another module and it's highlighted in blue/gray (like normal), and the running module still gets a blue/gray box around it.

In this picture, we're in test mode. The running module is "IdentifyPrimaryObjects", but the module we're looking at is "Crop".

screen shot 2017-09-07 at 2 07 29 pm

We've verified this functionality on Windows.

### DIFF
--- a/cellprofiler/gui/pipelinelistview.py
+++ b/cellprofiler/gui/pipelinelistview.py
@@ -1470,6 +1470,10 @@ class PipelineListCtrl(wx.PyScrolledWindow):
                     text_color = text_color_selected
                 else:
                     text_color = text_color
+            if self.test_mode and index == self.running_item:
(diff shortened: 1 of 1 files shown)


bethac07:
🎉

PR #3085, opened 2017-09-08
Documentation fixes for namesandtypes

bethac07:


cellprofiler/modules/namesandtypes.py
```diff
@@ -958,6 +975,28 @@ def settings(self):
...
+            self.assignments_count,
+            self.single_images_count,
+            self.manual_rescale,
+            self.process_as_3d,
+            self.x,
+            self.y,
+            self.z
+        ]
+
+        for assignment in self.assignments:
```
mcquin: This will result in duplicate documentation for multiple assignments. It's probably best to use the settings from `self.assignments[0]`.

PR #3088, opened 2017-09-10
Don't retain object outlines.

### LINKED ISSUE #3035: Object processing modules should not return outline images
AFAIK, the new convention was that outline images would have to be explicitly generated in OverlayOutlines, rather than being generated by Object Processing modules.  The outline generation has been removed from IdentifyPrimaryObjects and FilterObjects, but remains in

- [x] EditObjectsManually
- [x] ExpandOrShrinkObjects
- [x] FilterObjects
- [x] IdentifyObjectsManually
- [x] IdentifySecondaryObjects
- [x] IdentifyTertiaryObjects
- [x] MaskObjects
- [x] SplitOrMergeObjects

mcquin:
Resolves #3035 

Object identification modules no longer produce object outlines. Use **OverlayOutlines** to produce outline images.

### DIFF
--- a/cellprofiler/modules/untangleworms.py
+++ b/cellprofiler/modules/untangleworms.py
@@ -112,8 +112,18 @@ from centrosome.outline import outline
 from cellprofiler.preferences import standardize_default_folder_names, \
     DEFAULT_INPUT_FOLDER_NAME, DEFAULT_OUTPUT_FOLDER_NAME, NO_FOLDER_NAME, \
     ABSOLUTE_FOLDER_NAME
-from cellprofiler.modules._help import NAMING_OUTLINES_HELP, RETAINING_OUTLINES_HELP, USING_METADATA_GROUPING_HELP_REF, \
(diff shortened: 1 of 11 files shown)


cellprofiler/modules/untangleworms.py
```diff
@@ -112,9 +112,19 @@
 from cellprofiler.preferences import standardize_default_folder_names, \
     DEFAULT_INPUT_FOLDER_NAME, DEFAULT_OUTPUT_FOLDER_NAME, NO_FOLDER_NAME, \
     ABSOLUTE_FOLDER_NAME
-from cellprofiler.modules._help import NAMING_OUTLINES_HELP, RETAINING_OUTLINES_HELP, USING_METADATA_GROUPING_HELP_REF, \
+from cellprofiler.modules._help import USING_METADATA_GROUPING_HELP_REF, \
```
bethac07: Does the NAMING_OUTLINES_HELP need to be replaced also?

mcquin: Nope. It was unused.

PR #3090, opened 2017-09-10
Docfix helpcontent

### LINKED ISSUE #3022: Update icon variables in help/content to use new icons rather than old icons

bethac07:
Resolves #3022 and contributes towards #2840, among others


bethac07:
@mcquin with so many other doc changes coming in to `content.py` does it make sense to just review and pull now and I'll make a new PR for the next batch?  Just trying to avoid imminent merge crises.  LMK.

mcquin:
@bethac07 That's probably a good idea. I'll merge now.

PR #3091, opened 2017-09-11
Made cell-profiler output wall-clock times...

oleksandr-pavlyk:
alongside the cpu-times.

Wall-clock times are more meaningful.

Cpu-times include spinning times (when a thread is active, rather than sleeping, but not doing any work).

Example of the output:

```
Mon Sep 11 10:16:50 2017: Image # 1, module Images # 1: CPU_time = 0.00 secs, Wall_time = 0.00 secs
Mon Sep 11 10:16:50 2017: Image # 1, module Metadata # 2: CPU_time = 0.00 secs, Wall_time = 0.00 secs
Mon Sep 11 10:16:50 2017: Image # 1, module NamesAndTypes # 3: CPU_time = 96.90 secs, Wall_time = 3.47 secs
Mon Sep 11 10:16:53 2017: Image # 1, module Groups # 4: CPU_time = 0.00 secs, Wall_time = 0.00 secs
```


mcquin:
@oleksandr-pavlyk Cool!

I am curious about the behavior of this on Windows. From the [documentation on `os.times`](https://docs.python.org/2/library/os.html#os.times) (emphasis mine):

> Return a 5-tuple of floating point numbers indicating accumulated (processor or other) times, in seconds. The items are: user time, system time, children’s user time, children’s system time, and elapsed real time since a fixed point in the past, in that order. See the Unix manual page times(2) or the corresponding Windows Platform API documentation. **On Windows, only the first two items are filled, the others are zero.**

Will the reported wall time always be 0 on Windows?

mcquin:
@oleksandr-pavlyk Yes! This looks like a great alternative.

oleksandr-pavlyk:
@mcquin I have added use of `timeit.default_timer`.

oleksandr-pavlyk:
@mcquin If you would like me to rebase and squash commits, let me know.

0x00b1:
Hi, @oleksandr-pavlyk! Thanks for the commit. Looks great!

PR #3092, opened 2017-09-11
Update run multiple pipelines documentation

mcquin:
Running multiple pipelines only works with pipeline (.cppipe) files with image plane details. Use File > Export > Pipeline... to export "CellProfiler pipeline and file list".

### DIFF
--- a/cellprofiler/gui/help/content.py
+++ b/cellprofiler/gui/help/content.py
@@ -821,7 +821,7 @@ RUN_MULTIPLE_PIPELINES_HELP = u"""\
 The **Run multiple pipelines** dialog lets you select several
 pipelines which will be run consecutively. Please note the following:
 
--  Pipelines from CellProfiler 2.0 and lower are supported.
+-  Pipeline files (.cppipe) are supported.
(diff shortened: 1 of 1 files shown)


cellprofiler/gui/help/content.py
```diff
@@ -821,12 +821,11 @@ def __image_resource(filename):
 The **Run multiple pipelines** dialog lets you select several
 pipelines which will be run consecutively. Please note the following:
 
--  CellProfiler 2.1 project files are not currently supported.
 -  Pipelines from CellProfiler 2.0 and lower are supported.
--  If you want to use a pipeline made using CellProfiler 2.1, then you
-   need to include the project file list with the pipeline, by selecting
-   *Export > Pipeline…*, and under the “Save as type” dropdown, select
-   “CellProfiler pipeline and file list”.
+-  Project files (.cpproj) from CellProfiler 2.1 or newer are not supported.
```
bethac07: For the sake of an overabundance of clarity, can you say that project files are not supported but pipelines are, rather than simply saying project files aren't?

PR #3094, opened 2017-09-11
Documentation - Identify Primary

AnneCarpenter:


cellprofiler/modules/identifyprimaryobjects.py
```diff
@@ -79,22 +79,25 @@
...
+typically the most important but complex step in creating a good pipeline,
+so do not be discouraged: other modules are easier to configure!
 Using **IdentifyPrimaryObjects** with *'Use advanced settings?'* set to *'No'* 
-allows you to quickly try to segment your objects based only the size and input
-image; CellProfiler will then use its built-in defaults to decide how to set the 
-threshold and how to break clumped objects apart.  If you are happy with the 
-segmentation produced by the default settings, you can then move on to 
+allows you to quickly try to identify your objects based only their typical size;
+CellProfiler will then use its built-in defaults to decide how to set the 
+threshold and how to break clumped objects apart. If you are happy with the
```
mcquin: @AnneCarpenter "If you are unhappy"?

AnneCarpenter: Nope, "happy". I'm always trying to make ppl happy but in this case it's justified as is :)

mcquin: Cool! I wanted to double check. :)

PR #3098, opened 2017-09-12
Fix example log config

mcquin:


cellprofiler/gui/help/content.py
```diff
@@ -898,34 +898,35 @@ def __image_resource(filename):
...
-    args=('~/CellProfiler/logfile.log','w')
+    formatter=detailed
+    args=('cellprofiler.log', 'w')
 
     [formatter_detailed]
-    format=[%(asctime)s] %(name)s %(levelname)s %(message)s
+    format=%(asctime)s - %(name)s - %(levelname)s - %(message)s
     datefmt=
 
 The above file would print warnings and errors to the console for all
```
bethac07: The console currently is suppressed.  https://github.com/CellProfiler/CellProfiler/issues/3099  If we think that's easy to fix, we can leave docs as is, otherwise we should should adjust.

mcquin: I'm not sure what you mean. When I run CellProfiler from the command line (headless), I see information logged to my terminal. 

Do you see any output when you run CellProfiler from the command line?

bethac07: Sorry, I was not clear, my bad!  The console that pops up with CP is currently suppressed for anyone running from prebuilt. Running the prebuilt from the command line (by cd-ing into the install directory and running `cellprofiler -L DEBUG` from there) gives no output to the console you ran it from- rather it (presumably) goes to some hidden console .  Running from source with command line does give output.

Since far more people run prebuilt than source, we should either indicate that these instructions are for running from source only (in which case, what do we advise people running prebuilt to do?) or fix #3099.

mcquin: I understand now.

I don't seem to be able to run the build from the command line on OS X. Maybe @0x00b1 can provide us with more details.

mcquin: I have verified functionality on OS X. See #3099.

0x00b1:
It looks fine and requires only a minor rebase.  @bethac07 would you mind re-reviewing?

cellprofiler/gui/help/content.py
0x00b1: Thanks, @mcquin.

bethac07:
Console is re-enabled, so this seems fine to me.

0x00b1:
Thanks, @bethac07!

0x00b1:
Oops. I screwed up the rebase. Standby.

PR #3100, opened 2017-09-12
Undeprecate LoadData.

mcquin:
LoadImages and LoadSingleImages are still considered deprecated.


AnneCarpenter:
minor Q, should we have removed a file associated with "HelpContent_LegacyLoad_image0"?

mcquin:
@AnneCarpenter No. That variable references an image that is used elsewhere (the add module button).

PR #3110, opened 2017-09-13
Documentation formatting for MeasureImageSkeleton

mcquin:


cellprofiler/modules/measureimageskeleton.py
```diff
@@ -1,12 +1,14 @@
...
+====================
+
+**MeasureImageSkeleton** measures the number of branches and endpoints in a
 morphological skeleton. A branch is a pixel with more than two neighbors
 and and an endpoint is a pixel with only one neighbor. You can create a
-morphological skeleton with the MorphologicalSkeleton module from the
-Mathmatical Morphology category. You can use the ouput of this module
-for graph theoretic analysis.
+morphological skeleton with the **MorphologicalSkeleton** module from the
+*Advanced* category. You can use the ouput of this module for graph theoretic analysis.
```
bethac07: Can we add a squishy bio use too, like "graph theoretic analysis; you may also use it to analyze the number of total branches and endpoints for branching objects (such as neurons) in an image."  It's not in advanced, so I think being a bit more descriptive in the docs is nice.  Otherwise, 👍

PR #3119, opened 2017-09-13
Documentation - Adaptive thresholding in threshold.py

### LINKED ISSUE #3111: Update documentation about adaptive thresholding
In threshold.py the description of adaptive thresholding is as follows:
*{TS_ADAPTIVE}:* Partitions the input image into tiles and calculate
   thresholds for each tile. For each tile, the calculated threshold is
   applied only to the pixels within that tile.

Is that really true? I had thought there is a smoothing step; this makes it sound like it's literally a different threshold from one tile to the next which seems like always a bad strategy.

AnneCarpenter:
Resolves #3111


AnneCarpenter:
Solves #3111

PR #3140, opened 2017-09-13
Documentation - ColorToGray

AnneCarpenter:
Many changes are because it was likely originally written only for RGB and now handles other types.

I need two bits of help: 
1) pls check it's correct what I've changed from splitting -> combining
2) there are three similar settings like this: 
Select *"%(YES)s"* to extract the red channel to grayscale."""

My Q is: else what? Is it correct to add "Otherwise, the red channel will be (ignored?) (separated but stored as a color image?)" (if you know the answer please do make the change for red, green, and blue.

### DIFF
--- a/cellprofiler/modules/colortogray.py
+++ b/cellprofiler/modules/colortogray.py
@@ -4,14 +4,14 @@
 ColorToGray
 ===========
 
-**ColortoGray** converts an image with multiple color channels to a set
-of individual grayscale images.
+**ColortoGray** converts an image with multiple color channels to one or more
+grayscale images.
 
-This module converts RGB (Red, Green, Blue) color  and channel-stacked 
+This module converts color and channel-stacked 
 images to grayscale. All channels can be merged into one grayscale image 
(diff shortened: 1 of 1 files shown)


cellprofiler/modules/colortogray.py
```diff
@@ -207,7 +209,7 @@ def add_channel(self, can_remove=True):
 
         group.append("image_name", cps.ImageNameProvider(
                 "Image name", value="Channel%d" % (len(self.channels) + 1), doc="""\
-*(Used only when splitting images)*                
+*(Used only when combining images)*
```
bethac07: This is not correct, the one for combining is listed earlier.

AnneCarpenter:
Thanks Claire! Changes are great. 
For Beth's comment I reverted that text back per her advice but this github GUI makes it seem like it's still waiting on me... I suspect it might just be waiting for Beth to approve my changes, but if there is a way to force that since she is busy please do!

PR #3149, opened 2017-09-14
Documentation - SaveImages

AnneCarpenter:
Note that the 16/32 bit comment deleted from the main intro text does not seem so important as to mention it here (it's mentioned in two settings later), but it's fine to revert if someone thinks this is so crucial to state up front (I imagine most users just want to save their images and don't care). Same rationale for un-bolding 16/32 comment in one of the settings. No other setting I've come across in CP is bold, so it's hard to see why it's warranted here.


bethac07:
I'm ok with the unbolding, but I think we want to mention it up front; I don't feel super-duper strongly about  it though.

PR #3154, opened 2017-09-14
Documentation - MeasureObjectSizeShape

AnneCarpenter:
Some help please:
1) Please someone verify line 71 where I added centroid - is it true? And/or, can anyone elaborate on why it's different than LocationX from Identify?
2) see change max to min in line 102, correct?

### DIFF
--- a/cellprofiler/modules/measureobjectsizeshape.py
+++ b/cellprofiler/modules/measureobjectsizeshape.py
@@ -60,10 +60,7 @@ ellipse with the same second-moments as each object.
 -  *FormFactor:* *(2D only)* Calculated as 4\*π\*Area/Perimeter\ :sup:`2`. Equals 1
    for a perfectly circular object.
 -  *Solidity:* *(2D only)* The proportion of the pixels in the convex hull that are
-   also in the object, i.e., *ObjectArea/ConvexHullArea*. Equals 1 for a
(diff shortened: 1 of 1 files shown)


bethac07:
Someone mathier than me should weigh in on the centroid thing as I think it's not correct but I'm not certain (the definition we report here for how we report center (furthest from any edge) doesn't match the definitions of centroid I'm finding but I'm not sure if that means anything).  FWIW we do say in IPO that we ARE measuring the centroid there (the center of mass of the binary representation), and we say that the center reported in MeasureObjectSizeShape is different than the one in IPO.  

I don't think your change to 102 is right; I don't know why the MaximumRadius would be 1/2 of the _minimum_ width.

AnneCarpenter:
Agreed with Beth, line 102 should be reverted.

mcquin:
Interesting! It can't find *MeasureObjectSizeShape* and yet it's there.

I'll dig into this tomorrow morning.

AnneCarpenter:
Weird! See if you want to also delete the one sentence for #3155 at the same time. I'm nervous to touch anything!

mcquin:
Derp. It failed to load the module because it couldn't find the global variable `No` (note the "o" isn't capitalized). I've fixed it and removed the sentence for #3155

AnneCarpenter:
nice catch!

PR #3157, opened 2017-09-14
Documentation - RelateObjects

AnneCarpenter:
I moved "Distances" to be a child measurement rather than a parent one, please confirm that's accurate.

### DIFF
--- a/cellprofiler/modules/relateobjects.py
+++ b/cellprofiler/modules/relateobjects.py
@@ -112,6 +112,9 @@ speckles to the nuclei that contains them, the speckles are the children.
             D_ALL,
             doc="""\
 Choose the method to calculate distances of each child to its parent.
+For example, these measurements can tell you whether nuclear speckles
+are located more closely to the center of the nucleus or to the nuclear
+periphery.
 
 -  *{D_NONE}:* Do not calculate any distances. This saves computation time.
 -  *{D_MINIMUM}:* The distance from the centroid of the child object to
--- a/cellprofiler/modules/relateobjects.py
+++ b/cellprofiler/modules/relateobjects.py
@@ -12,13 +12,17 @@ objects. This is useful for counting the number of children associated
 with each parent, and for calculating mean measurement values for all
 children that are associated with each parent.
 
-An object will be considered a child even if the edge is the only part
-touching a parent object. If an child object is touching multiple parent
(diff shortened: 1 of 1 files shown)


bethac07:
Haven't checked rest of PR yet but that is accurate.  Nice catch.

cellprofiler/modules/relateobjects.py
```diff
@@ -12,13 +12,17 @@
...
-An object will be considered a child even if the edge is the only part
-touching a parent object. If an child object is touching multiple parent
+An object will be considered a child even if the edge is the only partly
+touching a parent object. If a child object is touching multiple parent
 objects, the object will be assigned to the parent with maximal overlap.
 For an alternate approach to assigning parent/child relationships,
 consider using the **MaskObjects** module.
 
-This module supports 2D and 3D objects
+If you want to include child objects that lie outside but still nearby
```
bethac07: nearby->near

cellprofiler/modules/relateobjects.py
```diff
@@ -157,9 +162,9 @@ def create_settings(self):
             False,
             doc="""\
 Select "*{YES}*" to calculate the per-parent mean values of every upstream
-measurement made with the children objects and stores them as a
-measurement for the parent; the nomenclature of this new measurements is
-“Mean_<child>_<category>_<feature>”. For this reason, this module
+measurement made with the children objects and store them as a
+measurement for the parent; the nomenclature of this new measurement is
+“Mean_<child>_<category>_<feature>”. This module
 should be placed *after* all **Measure** modules that make measurements
```
bethac07: Could you change "should" to "must" (or even "MUST") to make it more clear that this is mandatory?  This is a confusion point for a large number of users.

cellprofiler/modules/relateobjects.py
```diff
@@ -108,7 +113,7 @@ def create_settings(self):
             doc="""\
 Choose the method to calculate distances of each child to its parent.
```
bethac07: Would you be willing to add a one sentence use case? For example:

"For example, these measurements can tell you whether nuclear speckles are located more closely to the center of the nucleus or to the nuclear periphery".

AnneCarpenter:
Great, thanks for these additions. This one is ready to go.

Issue #3152, opened 2017-09-14
Documentation - Is it true Crop is the only module that can create a mask?


AnneCarpenter:
It says so in the help for SaveImages, here:
-  *{IF_MASK}:* Relevant only if the **Crop** module is used. The
   **Crop** module creates a mask of the pixels of interest in the
   image. Saving the mask will produce a binary image in which the
   pixels of interest are set to 1; all other pixels are set to 0.

I was wondering if other modules might create masks too, like Threshold. I don’t know whether an image must just be binary to be able to be saved as a mask here, or whether it needs to be stored explicitly as a mask by CellProfiler in order to be eligible to be saved as described here (in which case it's more plausible that Crop is the only module). But we have modules called MaskImage and MaskObjects too.

mcquin:
> I was wondering if other modules might create masks too, like Threshold

*Threshold* produces a binary image, but it does not produce a "mask" in the sense the word is used here.

> But we have modules called MaskImage and MaskObjects too.

Yes, these modules produce masks that you can save with **SaveImages**. It's worth updating the documentation to reflect this.

AnneCarpenter:
I'm on it.

mcquin:
https://github.com/CellProfiler/CellProfiler/pull/3175

Issue #3151, opened 2017-09-14
Documentation - SaveImages: metadata allowed?


AnneCarpenter:
For this chunk of code ~line 216, most similar settings have the blurbs about metadata that tell you that you can use metadata tags. Is that also true here? If so, need to paste in the proper blurbs (note PR #3149 is awaiting merge). 

        self.file_name_suffix = cellprofiler.setting.Text(
            "Text to append to the image name",
            "",
            metadata=True,
            doc="""\
*(Used only when constructing the filename from the image filename)*

Enter the text that should be appended to the filename specified above."""
        )

mcquin:
It appears that metadata tags are supported. We should update the documentation.

<img width="1072" alt="screen shot 2017-09-14 at 4 47 07 pm" src="https://user-images.githubusercontent.com/4721755/30455033-5c95f4f2-996c-11e7-866e-bfe6bfc1f10a.png">

AnneCarpenter:
I'm on it, will do in a few minutes.

mcquin:
https://github.com/CellProfiler/CellProfiler/pull/3175

PR #3175, opened 2017-09-14
Documentation - SaveImages

AnneCarpenter:
The metadata fixes solve #3151
The masking fixes solve #3152

I'm flying blind here so please check the results in the GUI (esp the part about metadata!)


mcquin:
Nice, @AnneCarpenter !!

I made a *tiny* modification to create paragraphs. Without newlines, it gets displayed as a single sentence.

PR #3182, opened 2017-09-15
Hide file format setting on saving movies

### LINKED ISSUE #3150: Documentation - Save Images: movie formats?
In this setting ~line 226, it says the setting is only used when saving non-movie files but then it says selecting a movie format is ok, which is a little confusing - I imagine “or movie” needs to be deleted here but wasn’t sure (note PR #3149 is awaiting merge).

       self.file_format = cellprofiler.setting.Choice(
            "Saved file format",
            [
                FF_JPEG,
                FF_NPY,
                FF_PNG,
                FF_TIFF
            ],
            value=FF_TIFF,
            doc="""\
*(Used only when saving non-movie files)*

Select the image or movie format to save the image(s)

mcquin:
Resolves #3150 

The only supported file format for movies is TIFF. Remove the option to chose a file format when a movie is being saved.

screen shot 2017-09-15 at 1 33 48 pm

### DIFF
--- a/cellprofiler/modules/saveimages.py
+++ b/cellprofiler/modules/saveimages.py
@@ -245,7 +245,7 @@ automatically.
             doc="""\
 *(Used only when saving non-movie files)*
 
-Select the image or movie format to save the image(s).
+Select the format to save the image(s).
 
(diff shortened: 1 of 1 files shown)


mcquin:
@0x00b1 Rebased.

PR #3183, opened 2017-09-15
Remove option to save a binary image.

### LINKED ISSUE #3160: Remove "Save binary image" setting in MeasureAreaImageOccupied?
In MeasureAreaImageOccupied
https://github.com/CellProfiler/CellProfiler/edit/master/cellprofiler/modules/measureimageareaoccupied.py

It seems the following option could be replaced and instead people could be guided to create an identical such image using other modules in the pipeline (I'm not the expert on which ones, but thresholding the label matrix might be a good strategy). But maybe I”m missing a common use case for this that explains why it is so handy to have this in this module?

Select *{YES}* if you would like to use a binary image later in the
pipeline, for example in **SaveImages**. The image will display the
object area that you have measured as the foreground in white and the
background in black.

mcquin:
Resolves #3160

### DIFF
--- a/tests/modules/test_measureimageareaoccupied.py
+++ b/tests/modules/test_measureimageareaoccupied.py
@@ -1,6 +1,7 @@
 import unittest
 
 import numpy
+import StringIO
 
 import cellprofiler.image
 import cellprofiler.measurement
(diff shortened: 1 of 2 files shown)


tests/modules/test_measureimageareaoccupied.py
```diff
@@ -1,6 +1,7 @@
 import unittest
 
 import numpy
+import StringIO
```
0x00b1: You should consider using the StringIO package available from `six` to ease the Python 3 transition.

mcquin:
@0x00b1 considered.

PR #3189, opened 2017-09-15
Documentation - Crop: adding X,Y to references to coordinates

AnneCarpenter:


mcquin:
This might be confusing because the GUI asks for "left" and "right" and "top" and "bottom", instead of "X" and "Y".

AnneCarpenter:
ooooh, excellent catch! I withdraw my PR in shame.

Issue #3194, opened 2017-09-17
Reconcile documentation between Smooth and CorrectIlluminationCalculate


AnneCarpenter:
There is some redundancy that needs pulling into a help blob. Please wait for PR #3195 to be accepted first.

Also consider the order in which the options are displayed and which should be default (in both modules). @bethac07 hopefully has an opinion on this.

AnneCarpenter:
The options in each context are a bit different so only some of the text can be exactly shared. Here are the settings held in common with text we want to use for each:

-  *%(SM_FIT_POLYNOMIAL)s:* This method is fastest but does not allow
   a very tight “fit” compared to the other methods. Thus, it will usually be less
   accurate. The method treats the intensity of the image
   pixels as a polynomial function of the x and y position of each
   pixel. It fits the intensity to the polynomial, *A x* :sup:`2` *+ B
   y* :sup:`2` *+ C xy + D x + E y + F*. This will produce a smoothed
   image with a single peak or trough of intensity that tapers off
   elsewhere in the image. For many microscopy images (where the
   illumination of the lamp is brightest in the center of field of
   view), this method will produce an image with a bright central region
   and dimmer edges. But, in some cases the peak/trough of the
   polynomial may actually occur outside of the image itself.

-  *%(SM_MEDIAN_FILTER)s* and *%(SM_GAUSSIAN_FILTER)s:* 
   We typically recommend
   *%(SM_MEDIAN_FILTER)s* vs. *%(SM_GAUSSIAN_FILTER)s* because the
   median is less sensitive to outliers, although the results are also
   slightly less smooth and the fact that images are in the range of 0
   to 1 means that outliers typically will not dominate too strongly
   anyway. The *%(GAUSSIAN_FILTER)s* convolves the image with a
   Gaussian whose full width at half maximum is the artifact diameter
   entered. Its effect is to blur and obscure features smaller than the
   specified diameter and spread bright or dim features larger than the
   specified diameter. The *%(MEDIAN_FILTER)s* finds the median pixel value within
   the diameter you specify. It removes bright or dim features
   that are significantly smaller than the specified diameter.

PR #3195, opened 2017-09-17
Documentation - CorrectIlluminationCalculate

AnneCarpenter:
1) The smoothing method setting said "See the **EnhanceOrSuppressFeatures** module help for more details. " That doesn't make sense to me, I assume it must've meant **Smooth**? Separately, I've made an issue to merge the settings documentation for smooth-related options within CorrecIllumCalc (#3194).
2) Can someone update the docs on Splines and Convex Hull... are they recommended, are they slow? Right now they are just described by pros/cons are not explained.
3) Note line ~321 I changed to FI_OBJECT_SIZE because I think it was wrong before, someone please confirm. Here is the context in case line numbers change: 

        self.object_width = cps.Integer(
            "Approximate object size", 10, doc="""\
*(Used only if %(FI_OBJECT_SIZE)s is selected for smoothing filter size calculation)*

Enter the approximate width of the artifacts to be smoothed, in pixels.

### DIFF
--- a/cellprofiler/modules/correctilluminationcalculate.py
+++ b/cellprofiler/modules/correctilluminationcalculate.py
@@ -201,8 +201,8 @@ are all equal to or greater than 1. You have the following options:
 Calculate a separate function for each image, or one for all the
 images? You can calculate the illumination function using just the
 current image or you can calculate the illumination function using all
-of the images in each group. The illumination function can be calculated
-in one of the three ways:
+of the images in each group (or in the entire experiment). The
+illumination function can be calculated in one of the three ways:
 
 -  *%(EA_EACH)s:* Calculate an illumination function for each image
    individually.
@@ -215,7 +215,9 @@ in one of the three ways:
    pipeline, but also means that you will not have the ability to filter
    out images (e.g., by using **FlagImage**). The input images need to
    be assembled using the **Input** modules; using images produced by
-   other modules will yield an error.
+   other modules will yield an error. Thus, typically,
+   **CorrectIlluminationCalculate** will be the first module after the
+   input modules.
 -  *%(EA_ALL_ACROSS)s:* Calculate an illumination function across all
    cycles in each group. This option takes any image as input; however,
    the illumination function will not be completed until the end of the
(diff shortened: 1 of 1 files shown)


cellprofiler/modules/correctilluminationcalculate.py
```diff
@@ -299,16 +311,17 @@ def create_settings(self):
...
 """ % globals())
 
         self.object_width = cps.Integer(
             "Approximate object size", 10, doc="""\
-*(Used only if %(FI_AUTOMATIC)s is selected for smoothing filter size calculation)*
+*(Used only if %(FI_OBJECT_SIZE)s is selected for smoothing filter size calculation)*
 
-Enter the approximate width of the artifacts to be smoothed, in pixels.
+Enter the approximate width of the largest typical object, in pixels.
+The smoothing size will be set to preserve objects of that size.
```
bethac07: I don't believe that this is correct, I believe this is the size of things you wish to smooth out.  Note it calls to `%(FI_OBJECT_SIZE)s`, whose definition is 
```
The module will calculate the smoothing size  based on the width of artifacts to be smoothed.
```

AnneCarpenter: Ok, I am nearly certain that %(FI_AUTOMATIC)s ought to have been changed to %(FI_OBJECT_SIZE)s so I think we are all set on that. 

As for how it's described, yeah, I think the documentation has been unclear about whether we are talking about the size of objects (that you want to preserve) or artifacts (that you want to capture as the illum function). I made this one change but if it's correct we'd need to change the other line you refer to as well (and should check the rest of text too). 

Does anyone know for certain what the code is actually doing and if it seems more like the size of objects (that you want to preserve) or artifacts (that you want to capture as the illum function)?

mcquin:
> The smoothing method setting said "See the EnhanceOrSuppressFeatures module help for more details. " That doesn't make sense to me, I assume it must've meant Smooth?

Yes, Smooth seems more appropriate.

> Can someone update the docs on Splines and Convex Hull... are they recommended, are they slow? Right now they are just described by pros/cons are not explained.

I'm not particularly familiar with these and how long they take to compute. Do @bethac07 or @karhohs have any experience?

> Note line ~321 I changed to FI_OBJECT_SIZE because I think it was wrong before, someone please confirm.

Confirmed! ([source](https://github.com/AnneCarpenter/CellProfiler/blob/c9e20811ba054ab9b23c16062e64475a5341c87a/cellprofiler/modules/correctilluminationcalculate.py#L507-L508))

cellprofiler/modules/correctilluminationcalculate.py
mcquin: My reading of the documentation is minimum size of objects to preserve/maximum size of artifacts to capture (remove or ignore). I haven't had the time to sit down with the code, but generally the smoothing filter size determines the size of features to keep. Features larger than the smoothing filter size are generally preserved.

AnneCarpenter:
I think it'd be best to merge the documentation for this module. There are two things undone and I can make issues for them separately; the second one esp could get messy:

1. Can someone update the docs on Splines and Convex Hull... are they recommended, are they slow? Right now they are just described by pros/cons are not explained.

2. (anne) Does anyone know for certain what the code is actually doing and if it seems more like the size of objects (that you want to preserve) or artifacts (that you want to capture as the illum function)?
(mcquin) My reading of the documentation is minimum size of objects to preserve/maximum size of artifacts to capture (remove or ignore). I haven't had the time to sit down with the code, but generally the smoothing filter size determines the size of features to keep. Features larger than the smoothing filter size are generally preserved

PR #3210, opened 2017-09-18
Documentation - MeasureImageSkeleton

AnneCarpenter:
Two Qs:
1) In rc5 this module doesn't have a nice bold heading when help is opened in CP's GUI. But the code looks like other modules that do (e.g., ConvertObjectsToImage) so I'm not sure why; maybe this was fixed later? But I'm looking at rc5 (I think).
2) I don't fundamentally understand how this differs from MeasureObjectSkeleton. Is the idea here that you measure endpoints and branchpoints irrespective of objects being identified? If so, please add this to the end of the intro help "Unlike **MeasureObjectSkeleton**, **MeasureImageSkeleton** measures branchpoints and endpoints throughout the entire image rather than associated with individual objects."


mcquin:
@0x00b1 has the most context on this, as the primary author.

AnneCarpenter:
Hopefully @0x00b1 can resolve the conflicts? LMK if you need anything from me.

0x00b1:
@AnneCarpenter Yep!

Issue #2278, opened 2016-09-01
Labels should be on top rather than to the left of inputs


0x00b1:
![forms_ _bootstrap](https://cloud.githubusercontent.com/assets/315821/18169536/8751ea16-7028-11e6-8bed-1cd36510a9df.png)

braymp:
Why?

0x00b1:
> Why?

…

0x00b1:
Since you’re curious, it simplifies box sizing. Moreover, it permits dynamic sizing (e.g. long labels and form elements clip if the label is too long).

mcquin:
It would also make settings which take multiple inputs (like Binary or https://github.com/CellProfiler/CellProfiler/pull/2280) easier to read.

0x00b1:
And it wouldn’t break every baseline.

0x00b1:
e.g. 
![h_and_v_and_window_and_5e92039e-702e-11e6-9728-36ae0b7ff6de_png__png_image__364 x 64_pixels_](https://cloud.githubusercontent.com/assets/315821/18171197/0dbe2532-702f-11e6-8665-83a80e094153.png)

braymp:
Gotcha. Part of me was wondering two things:
* Whether it makes the overall top-bottom panel content would become just that much lengthier (since they'd be stacked)
* I assumed that the straight-up left-to-right flow was easier to read/interpret, but I'm not dogmatic on that.

mcquin:
If a module take more than a screen of options to configure, then something is wrong.

0x00b1:
> I assumed that the straight-up left-to-right flow was easier to read/interpret, but I'm not dogmatic on that.

I think this is the least of our usability concerns.

AnneCarpenter:
This would solve #2135 so closing that one.

PR #3213, opened 2017-09-18
Documentation - MeasureObjectIntensityDistribution

AnneCarpenter:
Note that in the interface the question "Calculate intensity Zernikes?" is cutoff and appears as just  "Calculate intensity". I don't know how to fix that.


mcquin:
@AnneCarpenter That's, unfortunately, an annoying wx issue. We're considering displaying setting text above settings to avoid this issue. See #2278 .

PR #3216, opened 2017-09-18
Documentation - FilterObjects

AnneCarpenter:
Did a decent first pass here but will make a separate issue to clean up the settings help/logic for Classifier vs Rules.

I see a funky "no newline" symbol at line 1109, not sure if that's going to cause trouble.


AnneCarpenter:
Yeah, no dice, it seems I cannot fix the last line on my own.

cellprofiler/modules/filterobjects.py
```diff
@@ -21,6 +21,9 @@
 the original objects. Any measurements on the new object set will need
 to be made post-filtering by the desired measurement modules.
 
+See also any of the **MeasureObject** modules, **MeasureTexture**,
+**MeasureColocalization**, and **CalculateRatios**.
```
mcquin: CalculateRatios isn't a module.

Do you mean CalculateStatistics?

AnneCarpenter:
Good catch on "CalculateRatios" I updated to "CalculateMath".

PR #3220, opened 2017-09-19
Documentation - ExportToDatabase (part 1)

AnneCarpenter:
1) Can someone add a prefix to "At the secure shell" - the question I had here was "why are you telling me to do this?" Should the prefix be something like "Once you have completed the pipeline containing **ExportToDatabase** you will be ready to load (launch?) the database as follows:"
2) Please confirm my warning about the password is correct? (line ~731)


mcquin:
@AnneCarpenter Your warning about the password is correct. I saved a .cppipe and .cpproj and the password isn't obfuscated in either file.

PR #3229, opened 2017-09-19
Rename 'Label' as 'ConvertImageToObjects'

### LINKED ISSUE #3223: Change Label module's name
It's currently an Advanced module but it has functionality currently missing in CP's main modules. 

New name (draft): ConvertImageToObjects 
New category: Object processing

@mcquin: I think this is useful as a standalone module if your objects are neatly separated after thresholding. yes, this is essentially IdPrimaryObjects w/o declumping and filtering objects by size. tl;dr move to object processing? 
me: IIUC this is essentially ConvertImageToObjects? I know that name is a little silly, but if it's accurate that is probably better than Label bc the typical user never thinks about labels.
@mcquin: That would work! Also, we should make sure this module records the standard measurements for identified objects.

It is redundant with IdentifyPrimaryObjects except that module isn't 3D yet. Calling it "Convert" makes it sound appropriately much simpler than "Identify". Both modules should reference each other as "see also" 

For now (3.0 release) we can change name/category; we can make a new issue to update the m'ments made and improve documentation.

mcquin:
Resolves #3223 

screen shot 2017-09-19 at 1 11 01 pm


cellprofiler/modules/convertimagetoobjects.py
```diff
@@ -0,0 +1,35 @@
...
+"""
+
+import skimage
+import skimage.measure
+
+import cellprofiler.module
+
+
+class ConvertImageToObjects(cellprofiler.module.ImageSegmentation):
+    category = "Object Processing"
```
AnneCarpenter: Please revert the category to "Advanced"

Issue #1110, opened 2014-04-18
Document ExportToSpreadsheet saves Experiment.csv to default output folder


dlogan:
If ExportToSpreadsheet saves to a subfolder of the Default Output folder, and that folder uses Metadata tags, then the Experiment.csv is saved to the DOF, and not the specified subfolder.  All the other CSV files are put in the correct folder.

(Note: I used the Welcome Screen Fly images to test this.)

Attachments:
http://cellprofiler.org/issues/uploaded/cBQc5K/project_Experiment_CSV.cpproj

AnneCarpenter:
Given that at least one reputable person in the world likes the current behavior, let's just document the behavior in the help and be done with it.

AnneCarpenter:
(Could I ask @dlogan or @braymp to do this, if either of you still has an understanding of the issue?)

dlogan:
Yes, I can add a little Help clarification for this soon.

AnneCarpenter:
Hey @dlogan if you can just type the correct text in a comment I'm happy to put it in properly as a pull request.

AnneCarpenter:
I will do this (as best I can)

AnneCarpenter:
Strangely, I don't seem to have the power to close this issue @mcquin

mcquin:
Closed #3232

dlogan:
sorry @AnneCarpenter I lost track of this.  Your text edits read fine to me.

PR #3232, opened 2017-09-19
Documentation - ExportToSpreadsheet DOF behavior

AnneCarpenter:
Solves #1110
I'm flying blind so I've no idea if I rendered the URL properly. Please check it in the GUI!


mcquin:
I fixed the URL on merge :)

PR #3241, opened 2017-09-19
Documentation - SplitOrMergeObjects

AnneCarpenter:
How hard would it be to change "Unify" throughout, to "Merge" to be consistent with the module name? I was a little nervous to muck about because "Unify" is a setting choice so not sure if it affects variable revision numbers. Also change Unification to Merging.

### DIFF
--- a/cellprofiler/modules/splitormergeobjects.py
+++ b/cellprofiler/modules/splitormergeobjects.py
@@ -6,16 +6,18 @@ SplitOrMergeObjects
 
 **SplitOrMergeObjects** separates or combines a set of objects that
 were identified earlier in a pipeline. Objects that share a label, but
-are not touching will be relabeled into separate. Objects that share a
-boundary will be combined into a single object.
+are not touching will be relabeled into separate objects. Objects that
+share a boundary will be combined into a single object.
 
 Objects and their measurements are associated with each other based on
 their object numbers (also known as *labels*). Typically, each object is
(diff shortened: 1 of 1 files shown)


mcquin:
I can make those changes in another pull request.

bethac07:
Disagree about changing Unify to Merge- Merge to me makes it sound like
you're going to fill in the spaces between all the objects too, whereas
with Unify I understand better that you're now calling them all one big
super-object but ultimately not changing the pixels involved. Could just be
my odd brain though, willing to be overruled.

mcquin:
@bethac07 "Merge" is consistent with the module title. Should the module name be changed to SplitOrUnifyObjects?

AnneCarpenter:
It was indeed contentious, but luckily we already debated it: #632

bethac07:
If other people are happy with it, I'm sure it's fine, just something that
popped into my head. I'm boarding a flight rn so can't check that the
module help looks like, but if there isn't already a good figure showing
what the behavior is a  compromise could be to change wording to merge but
add a figure. If there is already a figure, then my objection is withdrawn.

AnneCarpenter:
I just submitted another PR to clarify in the intro help: #3248

bethac07:
If there isn't a figure I can likely make one today

AnneCarpenter:
A figure would be great too!

PR #3249, opened 2017-09-20
Documentation - EnhanceEdges

AnneCarpenter:
Is it possible to move settings to be this order in the help? 
Select the input (already first)
Name the output (already second)
Select an edge-finding method (ought to be third, right now a few settings are shown that are specific to dift methods so it's a little confusing)

I did so but feel free to revert if it causes problems.


mcquin:
@AnneCarpenter We can make that change; it needs to happen elsewhere in the code. No need to revert, necessarily. 

I'll handle the help setting order.

mcquin:
New help setting order:

screen shot 2017-09-20 at 9 08 27 am

AnneCarpenter:
perfect!

Issue #3235, opened 2017-09-19
Broken help button - Default output folder


AnneCarpenter:
1) Click on "View output settings" button in CP's main window
2) click the ? button to the left of "Default output folder" - nothing happens. The other ? button works fine.

AnneCarpenter:
I believe the same help content is accessed from Preferences window, which has ? buttons on it also. I don't know where its content is stored, but this is what it says: 

The Default Output Folder is accessible by pressing the “View output settings” button at the bottom of the pipeline panel. The Default Output Folder is the folder that CellProfiler uses to store the output file it creates. Also, several File Processing modules (e.g., SaveImages or ExportToSpreadsheet) provide the option of saving analysis results to this folder on a default basis unless you specify, within the module, an alternate, specific folder on your computer. Within modules, we recommend selecting the Default Output Folder as much as possible, so that your pipeline will work even if you transfer your images and pipeline to a different computer. If, instead, you type specific folder path names into a module’s settings, your pipeline will not work on someone else’s computer until you adjust those pathnames within each module. 
Use the Browse button (to the right of the text box) to specify the folder you would like to use as the Default Output Folder, or type the full folder path in the edit box. If you type a folder path that cannot be found, the message box below will indicate this fact until you correct the problem. If you want to specify a folder that does not yet exist, type the desired name and click on the New folder icon to the right of the Browse folder icon. The folder will be created according to the pathname you have typed.

mcquin:
Interesting! This works for me from source. I'll need to verify in the build. Which release candidate are you using?

AnneCarpenter:
Well that is bizarre! It says rc4 on it but Allen mentioned it's actually the code for rc5. I can reproduce (restarted CP just to be sure). This is not going to kill anyone if it's not easy to figure out.

mcquin:
Okay. I'll look at RC5 tomorrow.

There's a secondary issue here. The help doesn't convert reST to HTML. It's a one-liner fix. I'll get it tomorrow.

mcquin:
Very odd. It certainly doesn't work in RC4 or 5, though it works from the RC4 and 5 source tags.

mcquin:
The "Default Input Folder" button also doesn't work. Both are the only buttons instantiated by `__make_folder_panel`. I think the problem is somewhere in there.

mcquin:
@AnneCarpenter is this resolved in RC 7?

AnneCarpenter:
[...]

PR #3253, opened 2017-09-20
Display help text as HTML in preferences view.

mcquin:
Partial fix for #3235: 

Fixes the rendering of the help documentation in the GUI. Does not resolve the missing help on button press in builds.


mcquin:
I've bound the help event to the help button. This doesn't change the behavior of the GUI from source (which is good). Perhaps this will fix the behavior of the GUI in the compiled version. We should verify in the next release and, if fixed, close #3235 .

0x00b1:
@mcquin I'll queue a new build.

PR #3265, opened 2017-09-20
Documentation - FlagImages

AnneCarpenter:


AnneCarpenter:
Oh, whoa, I hadn't thought about that interpretation. Let me go test the module and see what it actually does and adjust accordingly.

AnneCarpenter:
Ok, thanks for the pointer - I confirmed that the name of the field stored in measurements indeed has nothing to do with the measurement name upon which the flag is based, it's solely based on the flag name you provide, so i've updated help to be more clear about that.

mcquin:
That makes a lot more sense.

PR #3270, opened 2017-09-20
add a few clarifying words

karhohs:


AnneCarpenter:
I don't even understand what stacked means here... are they all just placed in whatever RGB channels they were originally in?

cellprofiler/modules/graytocolor.py
```diff
@@ -65,8 +65,8 @@ def create_settings(self):
    blue intensities. The magenta image adds equally to the red and blue
    intensities. The yellow image adds equally to the red and green
    intensities.
--  *%(SCHEME_STACK)s*: The channels are stacked in order. An arbitrary
-   number of channels is allowed.
+-  *%(SCHEME\_STACK)s*: The channels are stacked in the order listed,
```
mcquin: `SCHEME_STACK`

mcquin:
@karhohs It looks like Anne would like more clarification on what "stacked" means.

Does the word "concatenated" make more sense, @AnneCarpenter ? E.g., 

```
The channels are concatenated in the order listed,
from top to bottom. An arbitrary number of channels is allowed.

For example, you could create a 5-channel image by providing
5 grayscale images. The first grayscale image you provide will fill 
the first channel, the second grayscale image you provide will fill
the second channel, and so on.
```

AnneCarpenter:
ohhhhh!!  Stacked is fine actually, it's the description that was really needed.

AnneCarpenter:
I do now!

PR #3280, opened 2017-09-21
updated classifier docs

### LINKED ISSUE #3217: Documentation - FilterObjects by Rules vs Classifier
Looks like the Classifier option was added without updating documentation relating to settings. 

In the GUI, when you change from Rules to Classifier, the options all stay exactly the same except _Class number_ changes to _Class name_.

But the settings help is confusing, at the least we need to update "Select the location of the rules file" to be "Select the location of the rules file or classifier file", but more importantly we need to change but under _Rules file name_, the settings help says stuff like this "Each line of this file should be a rule naming a measurement " which doesn't make sense for a classifier file. 

I don't know enough about the mode to be super helpful here.

0x00b1:
Resolves #3217

### DIFF
--- a/cellprofiler/modules/filterobjects.py
+++ b/cellprofiler/modules/filterobjects.py
@@ -242,7 +242,7 @@ maximal child is assigned. The choices are:
             doc="""\
 *(Used only if a per-object filtering method is selected)*
 
-This setting selects the container (i.e., parent) objects for the
+This setting selects the container (i.e., parent) objects for the 
 *{FI_MAXIMAL_PER_OBJECT}* and *{FI_MINIMAL_PER_OBJECT}* filtering
 choices.""".format(**{
                 "FI_MAXIMAL_PER_OBJECT": FI_MAXIMAL_PER_OBJECT,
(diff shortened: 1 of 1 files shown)


cellprofiler/modules/filterobjects.py
```diff
@@ -296,17 +299,18 @@ def set_directory_fn(path):
...
 *(Used only when filtering using {MODE_RULES} or {MODE_CLASSIFIERS})*
 
-The name of the rules file. This file should be a plain text file
-containing the complete set of rules.
+The name of the rules or classifier file. 
+
+A rules file is a plain text file containing the complete set of rules.
 
-Each line of this file should be a rule naming a measurement to be made
+Each line of the rules file should be a rule naming a measurement to be made
```
AnneCarpenter: Changes look great throughout... this was the part I was most uncertain about. Do all files from Classifier definitely have this structure? I would think those produced using GentleBoosting (the original default classifier type) would, but I'm not sure if you can now use SVM and other options from CPA's Classifier in CP here... and if so, are those types accurately described by this wording? I suspect not in which case I think we can readily resolve by just saying, "For example, if you chose GentleBoosting as the classifier in CellProfiler Analyst, each line of the file....". 

I can't rule out that my questions even make sense, because I don't have full understanding of this module!

For this line specifically, I think we probably don't want to say "the rules", wouldn't it be "the rules or classifier" (or just leave as is)?

0x00b1: Keep reading! :D

AnneCarpenter: bwaaa haaa haaaa

Issue #3224, opened 2017-09-19
Documentation - elaborate on Smooth/CorrectIlluminationCalculate options


AnneCarpenter:
Re. options: Splines and Convex Hull... are they recommended, are they slow? Right now they are just described by pros/cons are not explained. I'm hoping someone has hands-on experience to add, but no big deal if not: @karhohs @bethac07

braymp:
Convex Hull was designed with the BF worms in a well in mind. However, I've occasionally found it useful for other IAD projects with similar intensity profiles.

I forget what project Splines was implemented for. Depending on the settings, it can be slow; the more control points included, the slower it is to compute. However, for *really* irregular illumination/background profiles, it can help since it has the most flexibility. The original paper has more use case: http://www.cb.uu.se/~joakim/publications/SCIA_2001_J_Lindblad.pdf

AnneCarpenter:
Super, I will add this to the help. Thanks a lot!

PR #3284, opened 2017-09-22
Add extra guidance for convex hull and spline smoothing. #3224

karhohs:


AnneCarpenter:
Looks great.

0x00b1:
Nice addition

PR #3287, opened 2017-09-22
update smooth module for consistency

karhohs:
Smooth is to be consistent with CorrectIlluminationCalculate, which share smoothing algorithms. #3194


AnneCarpenter:
Super - whenever there is identical help text it can be pulled into a variable and called so that it's only present in one location in the code base. Ask @vchernys for help on that if needed!

Issue #3123, opened 2017-09-13
Documentation: ResizeObjects needs an assay dev's attention


AnneCarpenter:
I'm not 100% sure what it does, I guess it's ExpandOrShrink but it's only for 3D? The relationship between the two should be clarified and the documentation edited in general.

mcquin:
It's Resize but for objects instead of images.

AnneCarpenter:
Ohhhhhh, that never even occurred to me. Definitely needs elaboration then!

bethac07:
Presumably resolved by #3288

PR #3288, opened 2017-09-22
elaborate on difference between ResizeObjects and ExpandOrShrinkObjects

karhohs:
#3123 

This is beyond the scope of this pull-request, but an option to preserve the Z-dimension, or not, would be a useful addition to make this module more general. The documentation for this module has been expanded to make it clear that the z-dimension is not resized, but the name of the module does not hint at this specific behavior. The same option would be useful in **Resize**.

### DIFF
--- a/cellprofiler/modules/resizeobjects.py
+++ b/cellprofiler/modules/resizeobjects.py
@@ -4,16 +4,25 @@
 ResizeObjects
 =============
 
-Upsize or downsize an object’s label matrix by a factor or by specifying
-the final dimensions in pixels. ResizeObjects uses nearest neighbor
+**ResizeObjects** will upsize or downsize an object’s label matrix by a factor or by specifying
+the final dimensions in pixels. **ResizeObjects** is similar to **ResizeImage**, but
+**ResizeObjects** is specific to CellProfiler objects created by modules such as
+**IdentifyPrimaryObjects** or **Watershed**. **ResizeObjects** uses nearest neighbor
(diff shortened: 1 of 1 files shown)


karhohs:
Also, note that the "Process as 3D" x-y-z ratios would need to be updated after resizing x-y and not z. I don't think this is currently being tracked within CellProfiler.

cellprofiler/modules/resizeobjects.py
```diff
@@ -4,16 +4,25 @@
...
+the original depth (or z-dimension) will be kept. This 3D behavior was chosen, because in most
+cases the number of slices in a z-stack is much fewer than the number of pixels that define the
+x-y dimensions. Otherwise, a significant fraction of z information would be lost during downsizing.
+
+**ResizeObjects** is useful for processing very large or 3D data to reduce computation time. You
+might downsize a 3D image with **ResizeImage** to generate a segmentation, then use
+**ResizeObjects** to stretch the segmented objects to their original size
+before computing measurements with the original 3D image. **ResizeObjects** differs
+from **ExpandOrShrinkObjects** and **ShrinkToObjectCenters** in that the overall dimensions
+of the object label matrix, or image, is changed. In contrast, **ExpandOrShrinkObjects**
```
AnneCarpenter: should be _are_ changed.

AnneCarpenter:
This is an awesome description! I suggested one grammar change.
Let's wait and see if anyone specifically requests the behavior you describe before making it an option.

Issue #3215, opened 2017-09-18
Documentation - "See also" section should have a heading


AnneCarpenter:
I just realized one reason it's natural to put sections out of order (see #3177) is because the "See also" section does not have a major heading. Sometimes it seems like it's in a good spot and other times not, depending on what real section it gets appended to. 

We should rampage through all modules and find & replace "See also" with the proper heading structure.

AnneCarpenter:
This would be...
FIND:
`See also
`
REPLACE WITH: 
```
See also
========

See also
```


So that a typical section looks like this:
```
See also
========

See also other **Display** modules and data tools.
```

N3llz:
resolved by #3290 
@0x00b1 : it won't let me resolve this, but we've already merged the branch that did, so if you want to resolve this issue, it's ready.

PR #3290, opened 2017-09-22
Documentation - "See also" section should have a heading (Issues #3215)

N3llz:


0x00b1:
Nice work, @N3llz. Thanks for the PR.

PR #3308, opened 2017-09-28
License information

0x00b1:


AnneCarpenter:
Will think on it. In the meantime there seems to be a funky "licenseED" find and replace error?

AnneCarpenter:
I don't have any brilliant insight as  to where we should link to the license. There might have been a link from CellProfiler GUI menu bar at some point? 

But aside from that I don't think we need to change from what we've done before unless you think we need to. I thought the license was always just in the repo.

PR #3313, opened 2017-09-29
Documentation/measureimagequality_issue3146

### LINKED ISSUE #3146: Documentation - MeasureImageQuality: add rescale value?
The help says the image’s rescaling value is stored as a quality control metric, shouldn’t it be added to the list in section “measurements made by this module”? It’s optional but I don't think that affects a listing.

minh-doan:
Resolve #3146 
Image rescaling value is stored as a quality control metric

### DIFF
--- a/cellprofiler/modules/measureimagequality.py
+++ b/cellprofiler/modules/measureimagequality.py
@@ -90,7 +90,7 @@ Measurements made by this module
    -  *MinIntensity, MaxIntensity:* Minimum and maximum of pixel
       intensity values.
    -  *TotalArea:* Number of pixels measured.
-   -  Image’s rescaling value: if *Yes* is chosen for "Include the image rescaling value?",
+   -  *Image’s rescaling value*: if *Yes* is chosen for "Include the image rescaling value?",
       image’s rescaling value will be stored as a quality control metric.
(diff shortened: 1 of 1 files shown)


AnneCarpenter:
@minh-doan Can you please check the exact name of the metric? Right now it says "*Image’s rescaling value*" but usually this is a one-word name to be stored in output files so I am surprised if it's multiple words with a apostrophe.

Also a Q about the second change here: I think you deleted something that is still true. I think the rescaling value is only available to this module if the image was directly loaded by input modules (vs an image that went through another module, like Smooth, before arriving at this module). Can you put that back? Also I don't think you need to add the comment that the metric is only saved for images you choose in the module because that's always true. So basically, I think we can just revert the second part of the changed documentation back to its original unless I misunderstand something about how this all works!

0x00b1:
@AnneCarpenter and @minh-doan It would be easier to send a new pull request with the changes. @AnneCarpenter, I'll wait for you to approve once @minh-doan sends it.

minh-doan:
@AnneCarpenter 
Ah, sorry for the confusion. It was written as "**Input** modules" with bolded Input, making me thinking "Input" is a name of a module, which is not there. That's why... It's clear now, I'll make the changes as suggested.

Thanks !

PR #3325, opened 2017-10-02
Revert "Invert"'s behavior to subtract from 1

bethac07:


cellprofiler/modules/imagemath.py
```diff
@@ -404,7 +404,7 @@ def run(self, workspace):
             if opval == O_AVERAGE:
                 output_pixel_data /= sum(image_factors)
         elif opval == O_INVERT:
-            output_pixel_data = output_pixel_data.max() - output_pixel_data
+            output_pixel_data = 1.0 - output_pixel_data
```
0x00b1: Nice. However, would you mind using `skimage.utils.invert`? It is a little safer (it computes the maximum dtype value) and can deal with binary images.

tests/modules/test_imagemath.py
```diff
@@ -217,7 +217,7 @@ def test_invert(image_a, module, workspace):
 
     actual = output.pixel_data
 
-    expected = image_a.pixel_data.max() - image_a.pixel_data
+    expected = 1.0 - image_a.pixel_data
```
0x00b1: @bethac07 You should change this too.

tests/modules/test_imagemath.py
```diff
@@ -1050,7 +1050,7 @@ def fn(module):
 
         numpy.random.seed(0)
         image = numpy.random.uniform(size=(10, 10)).astype(numpy.float32)
-        expected = image.max() - image
+        expected = 1.0 - image
```
0x00b1: @bethac07 And this. :)

Issue #3310, opened 2017-09-28
Document per well options better in ExportToDatabase


bethac07:
The per-well means/medians/stdevs are only usable for MySQL (and CSV/MySQL, but I'm not sure the current status of #2961 ), not SQLite (per bb13353c36108bb18e1d524a91ab223ad7fc082a); this should be better documented in the help for ExportToDatabase.

minh-doan:
If no one has taken this yet, I'll try

bethac07:
Closed in #3337

PR #3337, opened 2017-10-09
Documentation/ExportToDatabase_issue3310

minh-doan:
Solve #3310 
At line 57, I really wanted to add "However, if you aim to use CellProfiler Analyst for the exported output, please consider using database type SQLite"
Probably things will change later ?


bethac07:
CPA can absolutely be used with MySQL too, not just SQLite, so I think leaving that line out was the correct decision. :)

PR #3354, opened 2017-10-20
Create output subdirectories in SaveCroppedObjects

### LINKED ISSUE #3353: SaveCroppedObjects directory behavior inconsistent with SaveImages
SaveCroppedObjects directory option should:

- default to "Default Output Folder" (currently defaults to "Elsewhere"),
- when "Default [Input|Output] Folder sub-folder" or "Elsewhere":
  - recursively create subdirectories (e.g., `foo/bar` should be created if missing) rather than failing,
  - populate metadata tags in pathnames (e.g., `METATAG/foo` should populate with `VALUE/foo`) rather than ignoring the metadata tag chunk of the path.

mcquin:
Resolves #3353


0x00b1:
Thanks, @mcquin. The two failures (#3355 and #3356) are unrelated to your patch.

PR #3361, opened 2017-10-26
Restore help manual GUI link.

### LINKED ISSUE #3296: Add the manual to Help > Online Manual
Restore [Help > Online Manual](https://github.com/CellProfiler/CellProfiler/blob/master/cellprofiler/gui/help/menu.py#L21-L25). 

The manual is [linked to CloudFront](https://github.com/CellProfiler/CellProfiler/blob/master/cellprofiler/gui/help/menu.py#L188-L191). Let's consider using a URL we can update so that we can updated the manual after the release.

mcquin:
Resolves #3296 

I've published the [CellProfiler manual on RTD](http://cellprofiler.readthedocs.io/en/latest/) and enabled a webhook to automatically build the latest documentation whenever there's a change to CellProfiler (we can configure this, if this seems too aggressive). 

This PR re-enables the GUI's Help > Online Manual. It's configured such that is tries to access a version of the documentation matching the running CellProfiler version (e.g., "v3.0.0"). If that fails, it defaults to "latest". I handle the logic here since it doesn't seem configurable in RTD.

There's only one manual step (which I'd love to figure out whether or not it can be automated). When a new version is tagged/released, RTD will update "latest" but it won't automatically make the tagged version available (e.g., "v3.0.0"). You have to [manually activate the newly published version](https://readthedocs.org/dashboard/cellprofiler/versions/). Anyone with a RTD account who is a maintainer can activate and deactivate versions. (@0x00b1 let's get you added and update the release process instructions with this step.)

**Note:** The published documentation is partially broken right now, because it's missing the `requirements.txt` file added by this PR. The "latest" documentation should build fine once this is merged.

### DIFF
--- a/cellprofiler/gui/help/menu.py
+++ b/cellprofiler/gui/help/menu.py
@@ -18,11 +18,10 @@ class Menu(cellprofiler.gui.menu.Menu):
             event_fn=lambda _: self.frame.show_welcome_screen(True)
         )
 
-        # TODO: Requires updated online help manual
-        # self.append(
-        #     "Online Manual",
-        #     event_fn=self.__on_help_online_manual
-        # )
+        self.append(
+            "Online Manual",
+            event_fn=self.__on_help_online_manual
+        )
 
         self.AppendSeparator()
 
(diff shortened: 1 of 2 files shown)


AnneCarpenter:
Super! I have a Q:  if the manual fails to build/publish to the HTML site, would we be notified in some way? Or is this something we ought to add to our checklist every time we release a new version?

mcquin:
@AnneCarpenter I can set up notifications. As mentioned, there's still a manual step for making the new version build visible on RTD. We can add a note mentioning that you should check the build was successful.

cellprofiler/gui/help/menu.py
```diff
@@ -185,10 +184,17 @@ def __on_help_developers_guide():
         import webbrowser
         webbrowser.open("https://github.com/CellProfiler/CellProfiler/wiki")
 
-    # @staticmethod
-    # def __on_help_online_manual(event):
-    #     import webbrowser
-    #     webbrowser.open("http://d1zymp9ayga15t.cloudfront.net/CPmanual/index.html")
+    @staticmethod
+    def __on_help_online_manual(event):
+        import requests
```
0x00b1: You could bump these imports to the module space since `requests` is imported elsewhere and `webbrowser` is part of the Python standard library.

requirements.txt
```diff
@@ -0,0 +1,3 @@
+--index-url https://pypi.python.org/simple/
```
0x00b1: Is this necessary? It looks like it’s just calling `pip`.

0x00b1:
> @0x00b1 let's get you added and update the release process instructions with this step.

Awesome. My username on RTD is `0x00B1`.

0x00b1:
Awesome! Thanks, @mcquin!

PR #3368, opened 2017-10-31
Fix ZeroDivisionError introduced by SciPy 1.0

0x00b1:


cellprofiler/modules/enhanceorsuppressfeatures.py
```diff
@@ -188,7 +188,7 @@ def create_settings(self):
         self.smoothing = cellprofiler.setting.Float(
             'Smoothing scale',
             value=2.0,
-            minval=0,
+            minval=0.1,
```
0x00b1: @mcquin What do you think about this default? Do you think epsilon is preferable?

mcquin: Machine epsilon is probably better than `0.1`. 

Alternatively, the smoothing step could be skipped if the value is 0. That would better preserve the current behavior.

cellprofiler/modules/enhanceorsuppressfeatures.py
```diff
@@ -556,6 +556,9 @@ def enhance_dic(self, image, angle, decay, smoothing):
 
             return result
 
+        if smoothing == 0:
```
0x00b1: I found that the issue was bubbling up from inside centrosome (i.e. centrosome wasn’t protecting the denominator) so I added this to address the problem. It has the benefit of not modifying the existing `smoothing` behavior.

PR #3374, opened 2017-11-07
Only convert floating point data to binary.

mcquin:
Additionally, added some simple documentation.

### DIFF
--- a/cellprofiler/modules/removeholes.py
+++ b/cellprofiler/modules/removeholes.py
@@ -1,17 +1,27 @@
 # coding=utf-8
 
 """
-Remove holes
-============
+RemoveHoles
+===========
+
+**RemoveHoles** fills holes smaller than the specified diameter.
+
+This module works best on binary and integer-labeled images (i.e., the output of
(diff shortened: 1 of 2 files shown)


cellprofiler/modules/removeholes.py
```diff
@@ -60,7 +71,8 @@ def run(self, workspace):
 def fill_holes(image, diameter):
     radius = diameter / 2.0
 
-    image = skimage.img_as_bool(image)
+    if image.dtype.kind == "f":
+        image = skimage.img_as_bool(image)
```
AetherUnbound: We would ideally like to use this module while preserving labels that are created during segmentation. It seems like there might be two ways to accomplish this if the original image comes in as an integer array: 
1. Use the output of `remove_small_holes` as a mask 
2. Use `remove_small_objects` instead of `remove_small_holes` 
I'd be happy to add those changes so we can use this module with labeled arrays.

AetherUnbound: I take it back, it looks like `remove_small_holes` couldn't be used as a mask in this case, and `remove_small_objects` wouldn't fill the holes.

Issue #3369, opened 2017-10-31
RemoveHoles should fill holes in labeled images


derekthirstrup:
Here is an image documenting the issue.  Also posted this issue to the forum.

![image](https://user-images.githubusercontent.com/20077409/32252592-4aec1e64-be53-11e7-8b67-e5d6930a272d.png)

mcquin:
Verified the behavior with the following pipeline:

1. Threshold
2. Label
3. ConvertObjectsToImage (16bit)
4. RemoveHoles

The output is a blank image. When provided a binary image (e.g., the output of Threshold), the filled binary image is returned.

AetherUnbound:
@mcquin What does the output of that filled binary image look like? Since a number of our images don't have distinct borders (like the one @derekthirstrup posted), that image would come back nearly all white. 
![image](https://user-images.githubusercontent.com/10214785/32566987-abbe9868-c46e-11e7-8522-16dd5dc69f1d.png)


I have an idea for how holes could be removed in a labeled image, and could add it as an additional module if that seems sensible.

mcquin:
@AetherUnbound Yes, in this case the returned image would basically be all white which, though technically correct behavior, is not the result you and @derekthirstrup are looking for.

I'd absolutely encourage you to submit a module which fills holes in labeled objects. :)

AetherUnbound:
@mcquin You're right - the technically correct behavior is not the desired behavior! 

Awesome, I'll go ahead and start work on that module. I know your PR is related to this issue, but I'd like to keep the issue open until I've submitted a PR with the labeled-filler module if that's alright.

mcquin:
Sounds good. I'll assign you to the issue.

AetherUnbound:
For context, I'm a software developer that works closely with @derekthirstrup at the Allen Institute for Cell Science. We've got a few modules we'd like to add and we may come across other changes/modifications. If I'm not overstepping myself, I'd love to be added as a collaborator! Otherwise, I'll continue to work through the external revision process.

mcquin:
Nice to meet you! Since we don't have a defined policy on adding collaborators to the project, I think it'd be best if you continued to contribute externally. I know it's super annoying. It's definitely something we should figure out.

AetherUnbound:
Gotcha. I think per AICS's standards we would want to add modules through the external contribution channel anyway. I only ask as a convenience for things like labeling @derekthirstrup's issues, adding reviewers, etc :)

[...]

PR #3382, opened 2017-11-14
FillObjects modules addition

AetherUnbound:
This module fills holes on a per-object basis, essentially preserving object labels in hole filling. This is one of two PRs to address Issue #3369


cellprofiler/modules/fillobjects.py
```diff
@@ -0,0 +1,90 @@
...
+
+import numpy
+import skimage.morphology
+
+import cellprofiler.image
+import cellprofiler.module
+import cellprofiler.setting
+
+
+class FillObjects(cellprofiler.module.ImageProcessing):
```
mcquin: Typically, modules with "Objects" in their name operate on CellProfiler `Objects`, rather than a CellProfiler `Image`. The `Objects` have a property `segmented` which returns the label matrix (essentially, the equivalent of **ConvertObjectsToImage** when the color format is *uint16*). 

With few modifications, this module could be reworked to operate on `Objects` by inheriting from `cellprofiler.module.ObjectProcessing`. This would save users the extra steps of converting their objects to and image, then back again. Measurements relating the original objects to the filled objects would also be recorded. Additionally, this module won't have to worry about whether or not it was passed the correct type of image data. 

What do you think?

cellprofiler/modules/fillobjects.py
```diff
@@ -0,0 +1,90 @@
...
+        ]
+
+    def run(self, workspace):
+        self.function = lambda image, diameter: \
+            fill_object_holes(image, diameter)
+
+        super(FillObjects, self).run(workspace)
+
+
+def fill_object_holes(image, diameter):
```
mcquin: If you inherited from `cellprofiler.module.ObjectProcessing`, this `image` parameter would be a label matrix. It would have 2 dimensions or 3 dimensions, corresponding to 2D or 3D objects. You wouldn't have to check if the third dimension represented channels.

cellprofiler/modules/fillobjects.py
AetherUnbound: I am all for it! The other implementations for object operations (`SplitOrMergeObjects, FilterObjects`) looked quite complex when I glanced over them, and I wanted to make sure that the CellProfiler team felt this module would actually be beneficial to the software before spending a ton of time on it. I'll work to refactor the module to operate on `Objects` instead (as the name already implied :stuck_out_tongue: ).

cellprofiler/modules/fillobjects.py
AetherUnbound: Good to know!

AetherUnbound:
@mcquin Thanks for the excellent feedback! Yes, I'd be happy to write some tests for this module once it operates on `Objects`.

AetherUnbound:
@mcquin It looks like there are a few different testing frameworks in place - which should I emulate? I'm referencing `test_resizeobjects`, `test_removeholes`, and `test_classifyobjects`.

mcquin:
@AetherUnbound You'll want to follow the pattern of `test_resizeobjects`.

Be sure to include this:

```python
@pytest.fixture(scope="module")
def image():
    return cellprofiler.image.Image()
```

There are a number of pytest fixtures defined in `tests/conftest.py`. The `image` fixture is parameterized such that each test including `image` (or a fixture which takes `image` as a parameter, such as `workspace`) will run three times: once with a grayscale 2D image, multichannel 2D image, and 3D grayscale image. Overriding the definition in `tests/conftest.py` ensures tests only run once.

cellprofiler/modules/fillobjects.py
```diff
@@ -0,0 +1,114 @@
+# coding=utf-8
+
+"""
+FillObjects
+===========
+
+**FillObjects** fills holes within all objects in an image.
+
+This module works best on integer-labeled images (i.e., the output of **ConvertObjectsToImage**
+when the color format is *uint16*).
```
mcquin: This paragraph is no longer necessary.

cellprofiler/modules/fillobjects.py
```diff
@@ -0,0 +1,114 @@
...
+
+The output of this module is a labeled image of the same data type as the input.
```
mcquin: This sentence can be removed.

AetherUnbound:
@mcquin @0x00b1 I'm getting a pytest error with the basic unit test I wrote. As far as I can tell, I've copied Claire's format almost exactly. Any thoughts?
```
image = <cellprofiler.image.Image object at 0x7f288ec83250>, module = <cellprofiler.modules.fillobjects.FillObjects object at 0x7f28b4143910>, workspace = <cellprofiler.workspace.Workspace object at 0x7f28b41432d0>

    def test_run(image, module, workspace):
        module.x_name.value = "example"
        module.y_name.value = "output"
        module.size.value = 6.
    
>       module.run(workspace)

modules/test_fillobjects.py:43: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../cellprofiler/modules/fillobjects.py:85: in run
    super(FillObjects, self).run(workspace)
../cellprofiler/module.py:1308: in run
    x = objects.get_objects(x_name)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <cellprofiler.object.ObjectSet object at 0x7f28b4143710>, name = 'example'

    def get_objects(self, name):
        """Return the objects instance with the given name
            """
>       return self.__objects_by_name[name]
E       KeyError: 'example'

../cellprofiler/object.py:826: KeyError
```

tests/modules/test_fillobjects.py
```diff
@@ -0,0 +1,65 @@
...
+
+labels[0:5, 0:8, 12:18] = 2
+
+labels[4:9, 12:18, 0:8] = 3
+
+labels[1:8, 12:20, 12:20] = 4
+
+
+def test_run(image, module, workspace):
+    module.x_name.value = "example"
```
mcquin: Regarding https://github.com/CellProfiler/CellProfiler/pull/3382#issuecomment-349200052:

[`module.x_name` is an `ObjectNameSubscriber`](https://github.com/CellProfiler/CellProfiler/blob/master/cellprofiler/module.py#L1229-L1232), so you will need to provide the [name of the objects (`"InputObjects"`)](https://github.com/AetherUnbound/CellProfiler/blob/bce15e3e69943921898e966b0574f7139f7f0656/tests/conftest.py#L75) rather than the [name of the image (`"example"`)](https://github.com/AetherUnbound/CellProfiler/blob/bce15e3e69943921898e966b0574f7139f7f0656/tests/conftest.py#L50).

tests/modules/test_fillobjects.py
```diff
@@ -0,0 +1,65 @@
...
+    module.y_name.value = "output"
+    module.size.value = 6.
+
+    module.run(workspace)
+
+    actual = workspace.image_set.get_image("output").pixel_data
```
mcquin: Similarly, the output of the module is and object. You'll want to change this to: 

```python
actual = workspace.object_set.get_objects("output").segmented
```

tests/modules/test_fillobjects.py
mcquin: You'll also need to set the object's label data (e.g., `objects.segmented = labels`).

tests/modules/test_fillobjects.py
AetherUnbound: Gah, thanks!

AetherUnbound:
@mcquin I've created some fixtures with data for common `test_run` usage in `ObjectProcessing` based modules...I'm not sure if that makes sense with this framework, so if you think it's best to just create the objects in the file (like it's done with `test_overlayobjects.py`) I can change it to that. Also, ignore all the `label` assignment above, I have a few more tests I want to write but I wanted to get the default run working first.

AetherUnbound:
@mcquin It seems like the current testing fixtures allowed modules to be run _either_ with the 3 default images via `test_run` _or_ tests with specific arrays once fixtures were overwritten (see `test_resizeobjects.py`). I've created some common "empty" fixtures to allow both kinds of tests to be run for the same module. Unit testing in python is still pretty new to me, so if this doesn't make any sense I'm happy to change it!

Beyond that, I've added some unit tests for this module. If you think more are needed please let me know! If not, I believe this is ready to be merged.

AetherUnbound:
@mcquin Your call! I don't really like the idea of constantly de-parameterizing fixtures in each module test file, so your workaround seems good. It also shows my inexperience with pytest :) I agree that it should be a separate pull request - I'll make an issue regarding it. I also just added the `<x>_empty` fixtures to CellProfiler-plugins, so the same pattern will eventually need to be applied there.

mcquin:
@AetherUnbound I wouldn't identify myself as being experienced with pytest fixtures. We're all learning here. :) Thanks for creating an issue and standardizing fixtures across repositories.

tests/modules/test_fillobjects.py
```diff
@@ -0,0 +1,157 @@
+import numpy
+import numpy.testing
+import skimage.morphology
+import pytest
+
+import cellprofiler.image
+import cellprofiler.measurement
+import cellprofiler.modules.fillobjects
+import cellprofiler.object
+import pdb
```
mcquin: Remove unused import.

AetherUnbound:
No worries! Can I go ahead and merge this?

mcquin:
> No worries! Can I go ahead and merge this?

Please wait. I'm working on a review right now.

tests/modules/test_resizeobjects.py
```diff
@@ -60,16 +41,16 @@ def volume_labels():
     return labels
 
 
-def test_resize_by_factor_shrink_image_labels(image_labels, module, object_set, objects, workspace):
-    objects.segmented = image_labels
+def test_resize_by_factor_shrink_image_labels(image_labels, module,  object_set_empty, objects_empty,  workspace_empty):
```
mcquin: There's some incorrect formatting throughout this file. I'll point out the first occurrence and leave the rest to you. 

There should only be one space between parameters.

tests/modules/test_resizeobjects.py
```diff
@@ -60,16 +41,16 @@ def volume_labels():
...
+    objects_empty.segmented = image_labels
 
     module.x_name.value = "InputObjects"
 
     module.method.value = "Factor"
 
     module.factor.value = 0.5
 
-    module.run(workspace)
+    module.run( workspace_empty)
```
mcquin: Remove the leading space before the argument: `module.run(workspace_empty)`

tests/modules/test_resizeobjects.py
```diff
@@ -81,35 +62,35 @@ def test_resize_by_factor_shrink_image_labels(image_labels, module, object_set,
 
     expected_labels[6:10, 6:10] = 4
 
-    numpy.testing.assert_array_equal(object_set.get_objects("ResizeObjects").segmented, expected_labels)
+    numpy.testing.assert_array_equal( object_set_empty.get_objects("ResizeObjects").segmented, expected_labels)
 
     numpy.testing.assert_array_equal(
-        workspace.measurements.get_current_measurement(
+         workspace_empty.measurements.get_current_measurement(
```
mcquin: This line should be indented by exactly 8 spaces.

tests/modules/test_resizeobjects.py
AetherUnbound: Gah, this is the result of some poorly written `sed` commands...thanks for catching that!

Issue #3381, opened 2017-11-14
Merge seeding artifacts into most significant neighbor


AetherUnbound:
In our seeded watershed segmentation with CellProfiler, we're finding that a number of seeding artifacts are left in the final segmentation. These over-segmented sections can't be removed and filled through normal methods as many of the seeds grow to the edge of the cell. 

Example:  
![image](https://user-images.githubusercontent.com/10214785/32806698-5917183c-c942-11e7-8d24-c2b26fdd1180.png)

minh-doan:
Can you try the module "SplitOrMergeObjects" to see if it helps?

AetherUnbound:
From @mcquin in #3384 

> The merge options for SplitOrMergeObjects indicate different approaches for merging objects than this implementation. Until that module supports 3D, it's sensible to leave this as a separate module.

AnneCarpenter:
Hi all, thanks for adding this functionality! Shouldn't it live in plugins, though, if we foresee it being temporary (i.e. eventually replaced by a more complete/3D SplitOrMerge, or even by IdentifyPrimaryObjects)?

AetherUnbound:
@AnneCarpenter That makes sense, I'm happy to move it there!

PR #3384, opened 2017-11-15
MergeObjects Plugin

AetherUnbound:
This module fills holes on a per-object basis, essentially preserving object labels in hole filling. This is one of two PRs to address Issue #3369. This merges smaller labeled objects into a neighboring object based on the object that contributes the most pixels to said object's border. 

Before merge: 
image

After merge:
image


AetherUnbound:
Since it seems there already exists a `SplitOrMergeObjects` module, does this module duplicate functionality? If it employs a different algorithm, is it useful to add as a different approach? I can also convert this module to operate on `Objects` @mcquin.

AnneCarpenter:
I can't answer your question but just wanted to note that IIUC, I *think* the functionality you are describing is available in IdentifyPrimaryObjects... but only for 2D. So it would not be shocking if you needed to add this functionality in the absence of IdentifyPrimaryObjects being upgraded to 3D (which is beastly).

mcquin:
> Since it seems there already exists a `SplitOrMergeObjects` module, does this module duplicate functionality? If it employs a different algorithm, is it useful to add as a different approach?

The [merge options](https://github.com/CellProfiler/CellProfiler/blob/master/cellprofiler/modules/splitormergeobjects.py#L137-L150) for `SplitOrMergeObjects` indicate different approaches for merging objects than this implementation. Until that module supports 3D, it's sensible to leave this as a separate module.

> I can also convert this module to operate on `Objects`.

Sounds great. :+1:

cellprofiler/modules/mergelabels.py
```diff
@@ -0,0 +1,106 @@
+# coding=utf-8
+
+"""
+MergeLabels
+===========
+
+**MergeLabels** merges labels below a certain threshold into its most prevalent, adjacent neighbor.
```
mcquin: I think you should add a note here stating that objects below the threshold with no adjacent neighbors will be removed.

AetherUnbound:
Per @AnneCarpenter's request in #3381 I've moved this into the plugins folder. This is probably ready to be merged.

AetherUnbound:
I'm going to close this PR and submit a PR to CellProfiler-Plugins instead (didn't realize that was a repo!).

PR #3400, opened 2017-12-05
Load Morph without error.

### LINKED ISSUE #3349: Change Morph's upgrade_settings to allow loading the rest of the pipeline.
When upgrading pipelines from 2.2 to 3.0, the biggest problem I've run across so far is in pipelines that contain Morph; if the Morph option is one that has now moved to its own module, you hit the [raise at line 587](https://github.com/CellProfiler/CellProfiler/blob/master/cellprofiler/modules/morph.py#L587) and whether you hit "cancel" or "continue processing" on the error message (see below) the rest of the pipeline does not load.  This is an especially big problem because Morph is often used near the beginning of a pipeline to clean images before segmentation and measuring, so it is not uncommon to see a case where you can only load the first 5 modules of a 40 module pipeline and then lose the rest.

As far as I can tell, this is the only module that behaves that way; if for example you had a pipeline in 2.2 with ApplyThreshold with a thresholding method that's now been removed (like MCT), loading that pipeline (correctly) shows an error at Threshold when loaded into 3.0 until you give it a new thresholding method, but it doesn't prevent the modules after it from being able to load.  

I'm not sure if the cleaner/better solution would be to just kill the "raise" line or to fix the "Continue Processing" behavior, but for the sake of our older users and power users it would be GREAT if this behavior could be changed by 3.0.1.  It'd be great if we could fully catch it and even put the new "correct" module into the pipeline, but I'd be happy just if it could be bypassed to allow the full pipeline to load.

image

mcquin:
Resolves #3349 

Load a pipeline with Morph without failing fast when encountering
an operation which has been moved to a separate module. Also
includes a tip on how to upgrade your pipeline.

screen shot 2017-12-05 at 10 29 51 am

### DIFF
--- a/cellprofiler/modules/morph.py
+++ b/cellprofiler/modules/morph.py
@@ -238,14 +238,28 @@ The following operations are available:
 
     - Binary
 
-|
+Upgrading:
+~~~~~~~~~~
 
-============ ============
-Supports 2D? Supports 3D?
-============ ============
-YES          NO
-============ ============
+The following **Morph** operations have been extracted to separate modules in CellProfiler 3.0.
+Use the table below to update your pipeline to use the corresponding module and, where appropriate,
+setting and value.
 
+=================  ======================  ===========  =======================
+Morph operation    Module                  Setting      Value
+=================  ======================  ===========  =======================
+bothat             TopHatTransform*        *Operation*  Black top-hat transform
+close              Closing
+dilate             Dilation
+erode              Erosion
+fill small holes   RemoveHoles
+invert             ImageMath               *Operation*  Invert
+open               Opening
+skel               MorphologicalSkeleton
+tophat             TopHatTransform*        *Operation*  White top-hat transform
+=================  ======================  ===========  =======================
+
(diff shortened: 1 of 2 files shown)


bethac07:
🎊

cellprofiler/modules/morph.py
```diff
@@ -238,14 +238,28 @@
...
+The following **Morph** operations have been extracted to separate modules in CellProfiler 3.0.
+Use the table below to update your pipeline to use the corresponding module and, where appropriate,
+setting and value.
+
+=================  ======================  ===========  =======================
+Morph operation    Module                  Setting      Value
+=================  ======================  ===========  =======================
+bothat             TopHatTransform*        *Operation*  Black top-hat transform
+close              Closing
+dilate             Dilation
```
AetherUnbound: I'm currently working on a `DilateObjects` module. Since it seems like these have been converted to plugins, should that module go to CellProfiler-plugins as well?

mcquin: I think DilateObjects makes sense in CellProfiler-plugins.

PR #3402, opened 2017-12-07
DilateObjects module

AetherUnbound:
This PR adds a `DilateObjects` module, and also creates a distinction between this and the `Dilation` module by renaming the latter to `DilatePixels`. This is added as there are use cases for applying dilation to objects as well as images. This same format could be applied to the other "morph" operations (`Erosion`, `Opening`, etc). 

These modules are functionally the same, they just inherit from separate classes. Is there a way to allow the single dilation operation to apply to either images or objects?


mcquin:
> This PR adds a `DilateObjects` module, and also creates a distinction between this and the Dilation module by renaming the latter to `DilatePixels`.

`XPixels` is not the convention for naming image processing modules which have an object processing counterpart. There are two conventions within the project:

1. `X` and `XObjects` (e.g., `Resize` and `ResizeObjects`)
1. `XImage` and `XObjects` (e.g., `MaskImage` and `MaskObjects`)

Pick your favorite.

---

> These modules are functionally the same, they just inherit from separate classes. Is there a way to allow the single dilation operation to apply to either images or objects?

Not directly. This is mostly limited by needing to know if input should be specified via a `ImageNameSubscriber` or `ObjectNameSubscriber`. There's no joint "`Subscriber`" which allows you to choose an image or objects (and I don't know that adding one would solve more problems than it would create, but we can discuss this). 

If you wanted to DRY up these modules, here are a few ideas:

1. Create a `dilate` utility function that can be shared between the two modules.
1. Create a base module which defines the `dilate` function and initializes some of the shared state (e.g., the structuring element), and make the image and object processing versions of the module inherit from that. You might not want to lose some of the automation behind the measurements `ObjectProcessing` modules make. Perhaps you can (carefully!!) leverage multiple inheritance.
1. Add an option "Image or objects" which allows the user to specify if the input/output will be an image or objects. The I/O logic would switch on this condition. 

---

Object dilation is a strange concept to me. Could you please explain your use case? We've talked about it some internally, but always settled on the `ConvertObjectsToImage` + `Dilate` pattern. 

Additionally, I think this module could be a better fit in CellProfiler-plugins.

cellprofiler/modules/dilateobjects.py
```diff
@@ -0,0 +1,87 @@
+# coding=utf-8
+
+"""
+DilateObjects
+===================
```
mcquin: reST will complain if the `===`s isn't the same length as the line above. Please adjust the length of this line appropriately.

cellprofiler/modules/dilatepixels.py
```diff
@@ -1,10 +1,10 @@
 # coding=utf-8
 
 """
-Dilation
+DilatePixels
```
mcquin: reST will complain if the ===s isn't the same length as the line above. Please adjust the length of this line appropriately.

cellprofiler/setting.py
```diff
@@ -1679,7 +1679,7 @@ def test_valid(self, pipeline):
 
 class StructuringElement(Setting):
     def __init__(self, text="Structuring element",
-                 value="disk,1",
+                 value="Disk,1",
```
mcquin: What necessitated this change?

cellprofiler/modules/__init__.py
```diff
@@ -29,7 +29,8 @@
                         'createbatchfiles': 'CreateBatchFiles',
                         'crop': 'Crop',
                         'definegrid': 'DefineGrid',
-                        'dilation': 'Dilation',
```
mcquin: When renaming modules, you'll need to add a line to [substitutions](https://github.com/AetherUnbound/CellProfiler/blob/f6d2f7896154d23b11bede3e8408de81848d31e4/cellprofiler/modules/__init__.py#L215) mapping the old name to the new name.

AetherUnbound:
Thanks for all the excellent comments @mcquin! I like `DilateImage` and `DilateObjects`, that convention makes much more sense to me. Your ideas for shared inheritance make sense, I'll experiment with your ideas and see what fits best. 

---

As for use cases, here's two I've encountered so far:

1. The seeds we generate for the watershed module are created by the `ShrinkObjectsToCenter` module, which only generates single pixel seeds. We've found that dilating these objects helps our watershed algorithm greatly. 
2. When using edge detection for segmentation, some objects may be split by just a few pixels. Dilating the objects can help bridge these small gaps and allow hole filling to be more accurate.

I think the main motivator for having this as an object processing module is so that we don't lose any measurements about the objects. IIRC `ConvertObjectsToImage` drops some of the measurements/history of the objects in the conversion process. If further object manipulation is desired later down the pipeline, this would require a re-conversion via `ConvertImageToObjects`. This way of thinking about objects/images may be incorrect though.

cellprofiler/setting.py
AetherUnbound: This was merely so that the structure elements would show up capitalized in the UI. It's entirely personal preference, so I can revert this if you like.

cellprofiler/modules/__init__.py
AetherUnbound: Good to know!

cellprofiler/modules/dilateimage.py
```diff
@@ -0,0 +1,56 @@
...
+import cellprofiler.image
+import cellprofiler.module
+import cellprofiler.setting
+import cellprofiler.utilities.sharedmethods
+
+
+class DilateImage(cellprofiler.module.ImageProcessing):
+    category = "Advanced"
+
+    module_name = "DilatePixels"
```
mcquin: `module_name = "DilateImage"`

cellprofiler/modules/dilateimage.py
```diff
@@ -0,0 +1,56 @@
...
+
+============ ============ ===============
+Supports 2D? Supports 3D? Respects masks?
+============ ============ ===============
+YES          YES          NO
+============ ============ ===============
+"""
+
+import numpy
+import skimage.morphology
```
mcquin: `skimage.morphology` is unused.

cellprofiler/modules/dilateimage.py
```diff
@@ -0,0 +1,56 @@
...
+|
+
+============ ============ ===============
+Supports 2D? Supports 3D? Respects masks?
+============ ============ ===============
+YES          YES          NO
+============ ============ ===============
+"""
+
+import numpy
```
mcquin: `numpy` is unused.

cellprofiler/modules/dilateobjects.py
```diff
@@ -0,0 +1,58 @@
...
+
+============ ============ ===============
+Supports 2D? Supports 3D? Respects masks?
+============ ============ ===============
+YES          YES          NO
+============ ============ ===============
+
+"""
+
+import numpy
```
mcquin: `numpy` is unused.

cellprofiler/modules/dilateobjects.py
```diff
@@ -0,0 +1,58 @@
...
+import skimage.morphology
```
mcquin: `skimage.morphology` is unused.

cellprofiler/modules/dilateobjects.py
```diff
@@ -0,0 +1,58 @@
...
+import skimage.segmentation
```
mcquin: `skimage.segmentation` is unused.

cellprofiler/setting.py
mcquin: It's okay. In the future, please try to limit changes based on personal preference.

cellprofiler/utilities/sharedmethods.py
```diff
@@ -0,0 +1,33 @@
+"""
+Shared methods used by multiple modules
```
mcquin: Let's create functionality-specific modules. A `cellprofiler.utilities.morphology` module would be specific.

cellprofiler/utilities/sharedmethods.py
```diff
@@ -0,0 +1,33 @@
...
+"""
+
+import numpy
+import skimage.morphology
+
+
+def dilation(x_data, structuring_element):
+
```
mcquin: I understand you didn't introduce this formatting, but do you mind fixing it? There shouldn't be a newline after `def` and `if`/`elif`/`else`. 

Additionally, `elif` and `else` are unnecessary after a `return` statement.

AnneCarpenter:
Question for @AetherUnbound : does ExpandOrShrinkObjects not have the functionality you need?

AetherUnbound:
@AnneCarpenter At this time I believe `ExpandOrShrinkObjects` does not support 3D data. Furthermore, the morphological structures used to define the dilation are more robust with `skimage.morphology.dilation` (it looks like `ExpandOrShrinkObjects` uses `binary_shrink` in 2D).

AnneCarpenter:
Thanks for explaining! It'd be awesome to make ExpandOrShrinkObjects 3D soon! Do you think we should switch ExpandOrShrinkObjects to use skimage.morphology.dilation, is it better in some way?

AetherUnbound:
I think it *might* be better to use skimage because there seems to be more configuration available with the structuring element. Also, it seems like @0x00b1 would like to move away from using centrosome anyway (which ExpandOrShrinkObjects uses): #2215

Issue #3407, opened 2017-12-07
Help search is currently broken


bethac07:
From [the forum](http://forum.cellprofiler.org/t/filenames-issue-in-cellprofiler-3-0-on-windows/5371/10?u=bcimini)

Using the help search (at least for the couple of keywords I tried) gives the following error message, at least on Windows 10 in both 3.0.0 built and today's master.

```
Traceback (most recent call last):
  File "cellprofiler\gui\help\search.py", line 129, in <lambda>
    lambda _: __on_search(self.results_view)
  File "cellprofiler\gui\help\search.py", line 89, in __on_search
    html = search_module_help(search_text)
  File "cellprofiler\gui\help\search.py", line 234, in search_module_help
    help_text = module.get_help()
  File "cellprofiler\module.py", line 255, in get_help
    "MODULE_DOC": self._rst_to_html_fragment(self.__doc__),
  File "cellprofiler\module.py", line 212, in _rst_to_html_fragment
    parts = docutils.core.publish_parts(source=source, writer_name="html")
  File "C:\Python27\lib\site-packages\docutils\core.py", line 448, in publish_parts
    enable_exit_status=enable_exit_status)
  File "C:\Python27\lib\site-packages\docutils\core.py", line 662, in publish_programmatically
    output = pub.publish(enable_exit_status=enable_exit_status)
  File "C:\Python27\lib\site-packages\docutils\core.py", line 217, in publish
    self.settings)
  File "C:\Python27\lib\site-packages\docutils\readers\__init__.py", line 71, in read
    self.input = self.source.read()
  File "C:\Python27\lib\site-packages\docutils\io.py", line 426, in read
    return self.decode(self.source)
  File "C:\Python27\lib\site-packages\docutils\io.py", line 99, in decode
    data_encoding = self.determine_encoding_from_data(data)
  File "C:\Python27\lib\site-packages\docutils\io.py", line 142, in determine_encoding_from_data
    if data.startswith(start_bytes):
AttributeError: 'NoneType' object has no attribute 'startswith'
```

AetherUnbound:
After some debugging it looks like `self.__doc__` is none during the `self._rst_to_html_fragment(self.__doc__)` call.

AetherUnbound:
It looks like this is the caused by:
[...]

PR #3408, opened 2017-12-08
Fix help search 

AetherUnbound:
Addresses #3407. Fixing this issue actually exposed a number of other issues, namely:

- `previous.png` (previous match arrow) wasn't displaying properly
- There was extra whitespace between the previous text arrow and the match term
- The first match title was duplicated
- No indication of count was actually displayed (it was only present in the title)

### DIFF
--- a/cellprofiler/gui/help/search.py
+++ b/cellprofiler/gui/help/search.py
@@ -273,6 +273,11 @@ def search_module_help(text):
 
         start_match = re.search(r"<\s*body[^>]*?>", help_text, re.IGNORECASE)
 
+        # Some pages don't have in-line titles
+        # Not matching "<h1>" here for cases that have "<h1 class='title'>", etc.
+        if not help_text.startswith(u"<h1"):
+            body += u"<h1>{}</h1>".format(title)
+
         if start_match is None:
             start = 0
         else:
--- a/cellprofiler/gui/help/search.py
+++ b/cellprofiler/gui/help/search.py
@@ -262,10 +262,7 @@ def search_module_help(text):
 
     match_num = 1
 
-    prev_link = u"""\
-<a href="#match%d" title="Previous match">
-    <img alt="previous match" src=memory:previous.png">
-</a>"""
+    prev_link = u"""<a href="#match%d" title="Previous match"><img alt="previous match" src="memory:previous.png"></a>"""
 
     anchor = u"""<a name="match%d"><u>%s</u></a>"""
 
--- a/cellprofiler/gui/help/search.py
+++ b/cellprofiler/gui/help/search.py
@@ -249,7 +249,7 @@ def search_module_help(text):
     <title>{count} match{es} found</title>
 </head>
 <body>
-    <h1>Match{es} found</h1><br>
+    <h1>Match{es} found ({count} total)</h1><br>
     <ul></ul>
 </body>
 </html>
--- a/cellprofiler/gui/help/search.py
+++ b/cellprofiler/gui/help/search.py
@@ -271,9 +271,6 @@ def search_module_help(text):
     for title, help_text, pairs in matching_help:
         top += u"""<li><a href="#match{:d}">{}</a></li>\n""".format(match_num, title)
 
-        if help_text.find("<h1>") == -1:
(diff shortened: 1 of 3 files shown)


AetherUnbound:
Ah, you're right. The images don't seem to render for me 
image

AetherUnbound:
Upon further inspection it looks like none of the help images are rendering for me on Ubuntu. Do they look fine for you @bethac07? If so this is a separate issue.

cellprofiler/modules/measureobjectintensitydistribution.py
```diff
@@ -82,10 +58,31 @@ def __image_resource(filename):
...
-                "MeasureObjectIntensityDistribution_Magnitude_Phase":MeasureObjectIntensityDistribution_Magnitude_Phase,
-                "MeasureObjectIntensityDistribution_Edges_Centers":MeasureObjectIntensityDistribution_Edges_Centers
-})
+"""
+
+import os
+import pkg_resources
+
+
+def __image_resource(filename):
```
mcquin: You can use `__image_resource` defined in [`cellprofiler.modules._help`](https://github.com/CellProfiler/CellProfiler/blob/master/cellprofiler/modules/_help.py#L228-L241).

cellprofiler/modules/measureobjectintensitydistribution.py
```diff
@@ -167,6 +164,10 @@ def __image_resource(filename):
 
 
 class MeasureObjectIntensityDistribution(cellprofiler.module.Module):
+    __doc__ = __doc__.format(**{
```
mcquin: When imports are needed to render documentation, the conventional module structure is:

```python
import foo
import bar
import baz

__doc__ = """
MyModule
========

{SOME_ELEMENT}\
""".format(**{
    "SOME_ELEMENT": foo.render(some_element)
})

class MyModule(cellprofiler.module.Module):
    # module stuff
```

cellprofiler/modules/measureobjectintensitydistribution.py
AetherUnbound: This was leftover code I just moved around, I can go ahead and import this.

cellprofiler/modules/measureobjectintensitydistribution.py
AetherUnbound: Gotcha, my concern was that the imports before the docstring and the `format` call on the docstring were causing python not to set `self.__doc__` properly. 
```
>>> class Foo(object):
...   """Here's my docstring {}""".format("test")
... 
>>> a = Foo()
>>> a.__doc__
>>> a.__doc__ is None
True
```

mcquin: TIL use `__doc__ = """...""".format(...)`:

```python
In [1]: class Foo(object):
   ...:     """Here is my docstring"""
   ...:

In [2]: class Bar(object):
   ...:     """Here is my {}""".format("docstring")
   ...:

In [3]: class Baz(object):
   ...:     __doc__ = """Here is my {}""".format("docstring")
   ...:

In [4]: foo = Foo()

In [5]: foo.__doc__
Out[5]: 'Here is my docstring'

In [6]: bar = Bar()

In [7]: bar.__doc__

In [8]: baz = Baz()

In [9]: baz.__doc__
Out[9]: 'Here is my docstring'
```

cellprofiler/gui/help/search.py
```diff
@@ -274,9 +271,6 @@ def search_module_help(text):
     for title, help_text, pairs in matching_help:
         top += u"""<li><a href="#match{:d}">{}</a></li>\n""".format(match_num, title)
 
-        if help_text.find("<h1>") == -1:
```
AetherUnbound: I thought this was unnecessary, but it turns out some pages have titles that don't exist in the body (hence this _is_ needed). The reason it wasn't matching properly was because some pages also have `<h1 class="title>`.

AetherUnbound:
@mcquin I've made some further changes for the search displaying and I've fixed the `MeasureObjectIntensityDistribution` docstring. I think this is ready to be merged!

bethac07:
I can see images 

image

PR #3415, opened 2017-12-10
Fix Crop with rectangle

### LINKED ISSUE #3358: Crop based on "Mouse" fails if you alter the cropping box
From [the forum](http://forum.cellprofiler.org/t/crop-module-error/5224):

> Hi, I get this error using the latest 3.0.0 release if I try to resize the rectangular selection when cropping, but it goes ahead without an error if I leave it as it is (a square):
‘Error while processing Crop:
slice indices must be integers or None or have an index method’

I've verified on my machine- if you just leave the default crop box it starts with, the module completes, but if you move one of the corners you get the full error trace here:
```
Failed to run module Crop
  File "CellProfiler\cellprofiler\gui\pipelinecontroller.py", line 2887, in do_step
  File "CellProfiler\cellprofiler\pipeline.py", line 2022, in run_module
  File "CellProfiler\cellprofiler\modules\crop.py", line 335, in run
  File "CellProfiler\cellprofiler\modules\crop.py", line 437, in ui_crop
  File "CellProfiler\cellprofiler\modules\crop.py", line 795, in apply_rectangle_cropping
TypeError: slice indices must be integers or None or have an __index__ method
```
Pipeline and image to validate are attached.  
[CP forum question.zip](https://github.com/CellProfiler/CellProfiler/files/1411374/CP.forum.question.zip)

AetherUnbound:
Convert crop location to integer before trying to index, resolves #3358


tests/modules/test_crop.py
```diff
@@ -282,6 +282,19 @@ def test_04_06_crop_color_with_rectangle(self):
...
         self.assertTrue(np.all(output_image.pixel_data == expected_image))
 
+    def test_04_07_crop_with_rectangle_float_bounds(self):
+        x, y = np.mgrid[0:10, 0:10]
+        input_image = (x / 100.0 + y / 10.0).astype(np.float32)
+        expected_image = input_image[2:8, 1:9]
+        workspace, module = self.make_workspace(input_image)
+        module.shape.value = cpmc.SH_RECTANGLE
+        module.horizontal_limits.set_value((1.2, 9.0000003))
+        module.vertical_limits.set_value((2.5, 8.999999))
```
mcquin: I'd expect this to round to 9, rather than 8. 

What about using `int(numpy.round(r[x]))` to round up or down?

AetherUnbound: I was thinking that as well, but I was thinking more of the edge case:
If the left bound is 0.00001, it should go to index 0. If the right bound is 255.99999, and it gets rounded up to 256, that index would be out of bounds for an array of length 256. 

That being said, we're already ensuring the selection is within the bounds so I can change this to be more rational :+1:

AetherUnbound: @mcquin Unfortunately, it looks like `vertical_limits` is of type `IntegerOrUnboundedRange`, so setting it via the `set_value` method automatically rounds the float down...Looks like this test didn't even sufficiently test my change! Do you know of another way we can set these values so that they can actually exist as floats at the time that `apply_rectangle_cropping` is called?

mcquin: Interesting! I suppose the solution would be to implement an equivalent `FloatOrUnboundedRange` setting. However, since the old behavior is to round down then your solution here is sufficient.

Issue #3357, opened 2017-10-23
Remove Morph module's "life" operation


AnneCarpenter:
In the Morph module, all options seem to work fine except “life”… it just hangs indefinitely. I propose we remove the option. 

In Test Mode, it freezes CP so that Force Quit is the only option. In Analyze Images mode, you retain the ability to stop it using Stop Analysis and it ends gracefully.

For extra brownie points, I tested all other operations in Morph and they work fine.

AnneCarpenter:
For even more brownie points, here is a pipeline to demonstrate the behavior:
[Archive.zip](https://github.com/CellProfiler/CellProfiler/files/1408780/Archive.zip)

AetherUnbound:
Since this uses the `cpmorphology` module (which it looks like we're trying to move away from) I also think it's fine to remove.

jhung0:
It's gone https://github.com/CellProfiler/CellProfiler/blob/master/cellprofiler/modules/morph.py#L599

PR #3416, opened 2017-12-11
Remove "life" option from Morph

AetherUnbound:
Addresses #3357

### DIFF
--- a/cellprofiler/modules/morph.py
+++ b/cellprofiler/modules/morph.py
@@ -147,11 +147,6 @@ The following operations are available:
       +---+---+---+----------------+---+---+---+
 
     - Binary
-  * - *Life*
-    - Applies the interaction rules from the `Game of Life
-      <http://en.wikipedia.org/wiki/Conway%27s_Game_of_Life>`__, an example of a
-      cellular automaton.
-    - Binary
   * - *Majority*
     - Each pixel takes on the value of the majority that surround it (keep
       pixel value to break ties):
@@ -286,7 +281,6 @@ F_DISTANCE = 'distance'
 F_ENDPOINTS = 'endpoints'
 F_FILL = 'fill'
 F_HBREAK = 'hbreak'
-F_LIFE = 'life'
 F_MAJORITY = 'majority'
 F_OPENLINES = 'openlines'
 F_REMOVE = 'remove'
@@ -298,7 +292,7 @@ F_THIN = 'thin'
 F_VBREAK = 'vbreak'
 F_ALL = [F_BRANCHPOINTS, F_BRIDGE, F_CLEAN, F_CONVEX_HULL,
          F_DIAG, F_DISTANCE, F_ENDPOINTS, F_FILL,
-         F_HBREAK, F_LIFE, F_MAJORITY, F_OPENLINES, F_REMOVE,
+         F_HBREAK, F_MAJORITY, F_OPENLINES, F_REMOVE,
          F_SHRINK, F_SKELPE, F_SPUR, F_THICKEN, F_THIN, F_VBREAK]
 
 R_ONCE = 'Once'
@@ -474,7 +468,7 @@ input for a measurement module.""" % globals()))
 
         if (function_name in (F_BRANCHPOINTS, F_BRIDGE, F_CLEAN, F_DIAG,
                               F_CONVEX_HULL, F_DISTANCE, F_ENDPOINTS, F_FILL,
-                              F_HBREAK, F_LIFE, F_MAJORITY,
+                              F_HBREAK, F_MAJORITY,
                               F_REMOVE, F_SHRINK, F_SKELPE, F_SPUR,
                               F_THICKEN, F_THIN, F_VBREAK)
             and not is_binary):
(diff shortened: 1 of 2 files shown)


AetherUnbound:
@mcquin I'm still a little confused with the version number upgrades. Do you mind explaining what's going on in that block and why the step sizes are 12?

mcquin:
> Do you mind explaining what's going on in that block and why the step sizes are 12?

First, allow me to provide some context:

```python
def upgrade_settings(self, setting_values, variable_revision_number, module_name, from_matlab):
```

* `setting_values` is a list of values for settings (as strings) as they were loaded from the pipeline or project file.
* `variable_revision_number` is the version of the module the pipeline or project file was last saved with.

When CellProfiler loads a pipeline/project file, it needs to map the `setting_values` to the module's settings. A module's settings are specified by its `settings` method. When settings are added or removed from a module, the `setting_values` need to be adjusted accordingly in the `upgrade_settings` method. The `variable_revision_number` allows us to make assumptions about the structure of `setting_values` in order to make the appropriate modifications.

In version 5, I removed a number of morph operations which were extracted and extended to support volumes in other modules. This rendered a number of additional settings unnecessary; these settings were parameters for the removed morph operations. The unused settings were removed which required reducing `setting_values` similarly.

Here's the [diff](https://github.com/CellProfiler/CellProfiler/commit/366f95c44e29fbe30919d4a049d3ad78432c402e#diff-97fe204cbc55b0c371b3e1902231b341L643) for that change. You can see that in the previous version of this module, there were 12 settings per operation. `settings` returned something that looks like this:

```python
[self.image_name,  # 0
 self.output_image_name,  # 1
 self.functions[0].function,   # 2 
 self.functions[0].repeats_choice,  # 3
 self.functions[0].custom_repeats,  # 4
 self.functions[0].scale,  # 5
 self.functions[0].structuring_element,  # 6
 self.functions[0].x_offset,  # 7
 self.functions[0].y_offset,  # 8
 self.functions[0].angle,  # 9
 self.functions[0].width,  # 10
 self.functions[0].height,  # 11
 self.functions[0].strel,  # 12
 self.functions[0].rescale_values,  # 13
 self.functions[1].function,   # 14
 ...]
```

To get the name for each operation, we get the values at indexes 2, 14, ... . Hence, the step size is 12. 

You can also see from the diff that the `function`, `repeats_choice`, `custom_repeats` and `rescale_values` are the preserved settings. In `upgrade_settings`, we get the [values of those settings per-operation](https://github.com/CellProfiler/CellProfiler/blob/master/cellprofiler/modules/morph.py#L592-L598) and then [stitch them together](https://github.com/CellProfiler/CellProfiler/blob/master/cellprofiler/modules/morph.py#L600) to form a `setting_values` list that can be mapped to the current `settings`, which looks like:

```python
[self.image_name,  # 0
 self.output_image_name,  # 1
 self.functions[0].function,   # 2 
 self.functions[0].repeats_choice,  # 3
 self.functions[0].custom_repeats,  # 4
 self.functions[0].rescale_values,  # 5 (was 13)
 self.functions[1].function,   # 6 (was 14)
 ...]
```

It's really confusing. I hope that makes a little bit of sense. Let me know if there's anything I can clear up. I should add a clarifying comment stating which settings were removed between versions 3 and 4.

Luckily for you, this change doesn't require modifying `setting_values`. Since there isn't an appropriate alternative to `"life"`, we'll let the user decide which operation they want to use instead. CellProfiler will complain when someone loads a pipeline/project with `"life"`, forcing them to reconfigure their pipeline. All you need to do, this time, is upgrade the `variable_revision_number`.

```python
if variable_revision_number == 5:
    # Removed operation "life".
    variable_revision_number = 6

return setting_values, variable_revision_number, from_matlab
```

cellprofiler/modules/morph.py
```diff
@@ -603,6 +595,15 @@ def upgrade_settings(self, setting_values,
 
             variable_revision_number = 5
 
+        if variable_revision_number == 5:
+            # Removed "life" operation
+            logger.warn("""
```
mcquin: Python doesn't strip leading or trailing whitespace from `"""` messages, so the resulting log message looks like:

```
# previous load or start cp warning

            Morph's 'Life' option has been removed, this pipeline might
            not be compatible with the current version of CellProfiler.

```

Two approaches I know to prevent unnecessary newlines and whitespace in messages are:

1. Use `\` within `"""`s to escape line breaks and shove the text all the way left:
    ```python
         logger.warn("""\
    Morph's 'Life' option has been removed, this pipeline might \
    not be compatible with the current version of CellProfiler.""")
     ```

2. Use multiple single-quoted strings, which Python will automatically concatenate:

    ```python
     logger.warn(
         "Morph's 'Life' option has been removed, this pipeline might "
         "not be compatible with the current version of CellProfiler."
     )
    ```

AetherUnbound: I could have sworn that I had found something online that stated that python multi-line strings didn't preserve that extra whitespace...perhaps that might have been for `argparse`. I'll fix that!

PR #3418, opened 2017-12-13
Read TIF stacks with skimage.external.tifffile

### LINKED ISSUE #3394: 3D Slice Viewer fails if exactly 3 or 4 slices
Hat tip to @tischi in [the forum](http://forum.cellprofiler.org/t/loading-data-into-cellprofiler-3d/5372);

If you try to use the 3D slice viewer with only 3 or 4 slices, you get a view that looks like the below; if you have 2 or 5-10 (and presumably everything after) it works fine.  Happens to him on Mac (at least with a 3 image stack) and me on PC.  Pipeline and images to verify here.
[3DViewer.zip](https://github.com/CellProfiler/CellProfiler/files/1516114/3DViewer.zip)

image

mcquin:
Resolves #3394

`skimage.io.imread` swaps axes on data it assumes to be multichannel.
CellProfiler only supports single-channel TIF stacks to represent
three-dimensional data. Avoid confusing stacks with only 3 or 4
planes with multichannel data by using `skimage.external.tifffile` to
load three-dimensional data.

### DIFF
--- a/cellprofiler/modules/loadimages.py
+++ b/cellprofiler/modules/loadimages.py
@@ -110,7 +110,7 @@ from cellprofiler.measurement import \
     C_FILE_NAME, C_PATH_NAME, C_URL, C_OBJECTS_FILE_NAME, C_OBJECTS_PATH_NAME, \
     C_OBJECTS_URL
 import numpy
-import skimage.io
+import skimage.external.tifffile
 
 '''STK TIFF Tag UIC1 - for MetaMorph internal use'''
 UIC1_TAG = 33628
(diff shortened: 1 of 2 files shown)


AetherUnbound:
Awesome! Does CellProfiler support other image types beyond tiffs?

mcquin:
@AetherUnbound 

> Awesome! Does CellProfiler support other image types beyond tiffs?

Yes, CellProfiler uses OME's BioFormats to load these [supported formats](https://docs.openmicroscopy.org/bio-formats/5.7.1/supported-formats.html). However, this only applies to two-dimensional image data. Three-dimensional data is currently restricted to the TIF file format.

AetherUnbound:
Another question (this shouldn't hold up the merge): Are there any plans/desires to allow CellProfiler to process multi-channel volumetric images?

mcquin:
> Are there any plans/desires to allow CellProfiler to process multi-channel volumetric images?

Yes. It's not on the roadmap yet, though it has come up in conversations as something we'd like to add support for in the future.

PR #3420, opened 2017-12-13
Display multichannel images as RGBA, truncating channels if necessary.

### LINKED ISSUE #2783: Module displays fail in master with >3 channel images
Trying to update ColorToGray to make a prettier display for >3 channel images, I found out those don't work at all in master:
```
Failed to run module ColorToGray
  File "C:\Users\bcimini\GitHub\Fork\CellProfiler\cellprofiler\gui\pipelinecontroller.py", line 2906, in do_step
    module.display(workspace, fig)
  File "C:\Users\bcimini\GitHub\Fork\CellProfiler\cellprofiler\modules\colortogray.py", line 316, in display
    self.display_split(workspace, figure)
  File "C:\Users\bcimini\GitHub\Fork\CellProfiler\cellprofiler\modules\colortogray.py", line 379, in display_split
    title="Original image")
  File "C:\Users\bcimini\GitHub\Fork\CellProfiler\cellprofiler\gui\figure.py", line 216, in wrapper
    return fn(*args, **kwargs)
  File "C:\Users\bcimini\GitHub\Fork\CellProfiler\cellprofiler\gui\figure.py", line 1302, in subplot_imshow
    subplot.imshow(self.normalize_image(image, **kwargs))
  File "C:\Python27\lib\site-packages\matplotlib\__init__.py", line 1819, in inner
    return func(ax, *args, **kwargs)
  File "C:\Python27\lib\site-packages\matplotlib\axes\_axes.py", line 4922, in imshow
    im.set_data(X)
  File "C:\Python27\lib\site-packages\matplotlib\image.py", line 453, in set_data
    raise TypeError("Invalid dimensions for image data")
TypeError: Invalid dimensions for image data
```

Sample image taken from [this forum post](http://forum.cellprofiler.org/t/can-you-use-cellprofiler-to-analyze-whole-tissue-samples-6-color-images/4711/5?u=bcimini)

mcquin:
Resolves #2783 

Only RGB(A) images can be displayed in CellProfiler. When asked to display multichannel data that is more than 4 channels, CellProfiler will display the first 3 channels as an RGB image and emit a warning that channels are being dropped from display.

I chose to display the first 3 channels in this case because we certainly don't want the 4th channel to be interpreted as the alpha channel. However, this means that all 4-channel data is interpreted as the 4th channel representing alpha.

Additionally, I've edited the titles of the original images in the display for **ColorToGray** to reflect that the image is displayed "as RGBA".


cellprofiler/gui/figure.py
```diff
@@ -1493,6 +1493,15 @@ def subplot_imshow_color(self, x, y, image, title=None, normalize=False, rgb_mas
         if rgb_mask is None:
             rgb_mask = [1, 1, 1]
 
+        # Truncate multichannel data that is not RGB(A) (5+ channel data) and display it as RGB.
```
AetherUnbound: So just to clarify - CellProfiler will accept >3 channel 2D images and process, it simply won't display all the channels as RGB?

mcquin: Correct. The first 3 channels will be displayed as RGB. Does this seem reasonable?

AetherUnbound: I think it does, especially with the warning message.

cellprofiler/gui/figure.py
```diff
@@ -1493,6 +1493,15 @@ def subplot_imshow_color(self, x, y, image, title=None, normalize=False, rgb_mas
...
+        if image.shape[2] > 4:
+            logger.warn(
+                "Multichannel display is only supported for RGB(A) (3- or 4-channel) data."
+                " Input image has {:d} channels. Only the first 3 channels will be displayed.".format(image.shape[-1])
```
AetherUnbound: Perhaps this could be `Only the first 3 channels will be displayed as RGB`?

mcquin: Good suggestion! I'll edit the message.

cellprofiler/gui/figure.py
mcquin: Cool. Let's see if @bethac07 agrees.

mcquin:
Thanks for the feedback, @bethac07 :)

> I think RGB(A) images are going to be much much rarer than 4-channel stacks; my preference would be to truncate (and note the truncation) for anything bigger than 3 channels, not 4.

Great point. There's only one case I can think of where RGBA is important (semi-transparent label overlays). I'll figure out how to make RGB the default display for 4+ channel images.

> Printing the message to the console isn't ideal as most users don't think to look there. However I can't think of a clean way to warn people though outside of identifying EVERY module which might display a multichannel image and putting a check for >3 channels in to change something in the module display window (ie change "ColorImage (as RGB)" to "ColorImage (first 3 channels as RGB)", which seems cumbersome and not ideal.

Do you recommend removing `(as RGB)` from those display titles? The alternative message `(first 3 channels as RGB)` is more clear but too verbose to fit in the already crowded module display.

> I think at a minimum we should add something to the help. In the help for Help->Using Module Display Windows -> How To Use The Image Tools there's a section on selecting which channels to show- perhaps we should add a note there that only the first three channels are shown. If anyone has a better idea, I'm all ears.

:+1: I'll add some information there.

PR #3429, opened 2017-12-21
Update module templates to reflect current style

### LINKED ISSUE #2289: Update module templates to reflect current preferred style

mcquin:
Resolves #2289 

A high-level overview of the changes made to templates:

- PEP 08/consistent formatting
- Remove import aliases and use absolute imports
- Include sections and section description in top-level module documentation
- Ensure templates are functional (i.e., they can be run as plugins)

### DIFF
--- a/cellprofiler/modules/plugins/imagetemplate.py
+++ b/cellprofiler/modules/plugins/imagetemplate.py
@@ -36,8 +36,8 @@ as `reStructuredText <http://docutils.sourceforge.net/rst.html>`_.
 
 Note whether or not this module supports 3D image data and respects masks.
 A module which respects masks applies an image's mask and operates only on
-the unmasked data. Update the table below to indicate which image processing
-features this module supports.
+the image data not obscured by the mask. Update the table below to indicate 
+which image processing features this module supports.
 
 |
 
--- a/cellprofiler/modules/plugins/imagetemplate.py
+++ b/cellprofiler/modules/plugins/imagetemplate.py
@@ -159,7 +159,7 @@ class ImageTemplate(cellprofiler.module.ImageProcessing):
 
     ###############################################
     #
-    # create_settings is where you declare the user interface elements
+    # "create_settings" is where you declare the user interface elements
     # (the "settings") which the user will use to customize your module.
     #
     # You can look at other modules and in cellprofiler.settings for
--- a/cellprofiler/modules/plugins/imagetemplate.py
+++ b/cellprofiler/modules/plugins/imagetemplate.py
@@ -111,7 +111,6 @@ wherever possible, include a link to the original work. For example,
    (`link <http://dx.doi.org/10.1016/1047-3203(90)90014-M>`__)
 """
 
-###################################
 #
 # Constants
 #
@@ -119,14 +118,12 @@ wherever possible, include a link to the original work. For example,
 # constants if they will appear more than once in your program. That way,
 # if someone wants to change the text, that text will change everywhere.
 # Also, you can't misspell it by accident.
-###################################
-
+#
 GRADIENT_MAGNITUDE = "Gradient magnitude"
 GRADIENT_DIRECTION_X = "Gradient direction - X"
 GRADIENT_DIRECTION_Y = "Gradient direction - Y"
 
 
-###################################
 #
 # The module class.
 #
@@ -149,9 +146,7 @@ GRADIENT_DIRECTION_Y = "Gradient direction - Y"
 #    should inherit from this class. These are modules that take objects as
 #    input and output new objects.
 #
-###################################
 class ImageTemplate(cellprofiler.module.ImageProcessing):
-    ###############################################
     #
     # The module starts by declaring the name that's used for display,
     # the category under which it is stored and the variable revision
@@ -161,12 +156,10 @@ class ImageTemplate(cellprofiler.module.ImageProcessing):
     # This module's category is "Image Processing" which is defined
     # by its superclass.
     #
-    ###############################################
     module_name = "ImageTemplate"
 
     variable_revision_number = 1
 
-    ###############################################
     #
     # "create_settings" is where you declare the user interface elements
     # (the "settings") which the user will use to customize your module.
@@ -174,7 +167,6 @@ class ImageTemplate(cellprofiler.module.ImageProcessing):
     # You can look at other modules and in cellprofiler.settings for
     # settings you can use.
     #
-    ################################################
     def create_settings(self):
         #
         # The superclass (cellprofiler.module.ImageProcessing) defines two
--- a/cellprofiler/modules/plugins/imagetemplate.py
+++ b/cellprofiler/modules/plugins/imagetemplate.py
@@ -1,3 +1,5 @@
+# coding=utf-8
+
 #################################
 #
 # Imports from useful Python libraries
@@ -126,16 +128,20 @@ GRADIENT_DIRECTION_Y = "Gradient direction - Y"
 
 ###################################
 #
-# The module class
+# The module class.
 #
-# Your module should "inherit" from cellprofiler.cpmodule.CPModule.
-# This means that your module will use the methods from CPModule unless
-# you re-implement them. You can let CPModule do most of the work and
-# implement only what you need.
+# Your module should "inherit" from cellprofiler.module.Module, or a
+# subclass of cellprofiler.module.Module. This module inherits from
+# cellprofiler.module.ImageProcessing, which is the base class for
+# image processing modules. Image processing modules take an image as
+# input and output an image.
+#
+# This module will use the methods from cellprofiler.module.ImageProcessing
+# unless you re-implement them. You can let cellprofiler.module.ImageProcessing
+# do most of the work and implement only what you need.
 #
 ###################################
-
-class ImageTemplate(cellprofiler.module.Module):
+class ImageTemplate(cellprofiler.module.ImageProcessing):
     ###############################################
     #
     # The module starts by declaring the name that's used for display,
@@ -143,9 +149,12 @@ class ImageTemplate(cellprofiler.module.Module):
     # number which can be used to provide backwards compatibility if
     # you add user-interface functionality later.
     #
+    # This module's category is "Image Processing" which is defined
+    # by its superclass.
+    #
     ###############################################
     module_name = "ImageTemplate"
-    category = "Image Processing"
+
     variable_revision_number = 1
 
     ###############################################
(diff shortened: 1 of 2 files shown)


cellprofiler/modules/plugins/imagetemplate.py
```diff
@@ -1,36 +1,115 @@
...
+produces a second image for downstream processing. You can use this as
+a starting point for your own module: rename this file and put it in your
+plugins directory.
+
+The text you see here will be displayed as the help for your module, formatted
+as `reStructuredText <http://docutils.sourceforge.net/rst.html>`_.
+
+Note whether or not this module supports 3D image data and respects masks.
+A module which respects masks applies an image's mask and operates only on
+the unmasked data. Update the table below to indicate which image processing
```
AetherUnbound: This seems confusing to me. It applies the mask but operates on the _unmasked_ data? (I'm probably misunderstanding how this works)

cellprofiler/modules/plugins/imagetemplate.py
```diff
@@ -1,36 +1,115 @@
...
+^^^^^^^^^^^^^^^
+
+Include implementation details or notes here. Additionally provide any 
+other background information about this module, including definitions
+or adopted conventions. Information which may be too specific to fit into
+the general description should be provided here.
+
+Omit this section if there is no technical information to mention.
+
+References
```
AetherUnbound: Thanks for all this great reference material :)

cellprofiler/modules/plugins/imagetemplate.py
```diff
@@ -49,125 +128,146 @@
...
-            0 to .5 = decreasing with increasing X, .5 to 1 = increasing
-            with increasing X).</li>
-            <li><i>%(GRADIENT_DIRECTION_Y)s</i> to get the relative
-            contribution of the gradient in the Y direction.</li></ul>
-            """ % globals()
+        self.gradient_choice = cellprofiler.setting.Choice(
+            text="Gradient choice:",
+            # The choice takes a list of possibilities. The first one
+            # is the default - the one the user will typically choose.
+            choices=[GRADIENT_MAGNITUDE, GRADIENT_DIRECTION_X, GRADIENT_DIRECTION_Y],
```
AetherUnbound: It might also be good to add a `value` keyword here to showcase default options.

cellprofiler/modules/plugins/imagetemplate.py
```diff
@@ -176,128 +276,110 @@ def create_settings(self):
...
-        image_set = workspace.image_set
-        #
-        # Get the input image object. We want a grayscale image here.
-        # The image set will convert a color image to a grayscale one
-        # and warn the user.
+        # The superclass's "run" method handles retreiving the input image
+        # and saving the output image. Module-specific behavior is defined
+        # by setting "self.function", defined in this module. "self.function"
+        # is called after retrieving the input image and before saving
+        # the output image.
```
AetherUnbound: I know that in some cases one needs access to other parts of `self` that aren't available in `self.function`. Do you think it would be beneficial to add that here or is this template supposed to be a fairly simple template?

cellprofiler/modules/plugins/imagetemplate.py
```diff
@@ -176,128 +276,110 @@ def create_settings(self):
...
+        power2 = numpy.sqrt((fft * fft.conjugate()).real)
+        mode = numpy.argwhere(power2 == power2.max())[0]
+        scale = numpy.sqrt(numpy.sum((mode + .5) ** 2))
+
+    gradient_magnitude = scipy.ndimage.gaussian_gradient_magnitude(pixels, scale)
+
+    if gradient_choice == GRADIENT_MAGNITUDE:
+        gradient_image = gradient_magnitude
+    else:
+        # Numpy uses i and j instead of x and y. The x axis is 1
```
AetherUnbound: I'm confused by this comment, where do `i` and `j` come up in this example?

cellprofiler/modules/plugins/measurementtemplate.py
```diff
@@ -1,59 +1,147 @@
...
+
+-  *MeasurementName*: A brief description of the measurement.
+-  *MeasurementName*: A brief description of the measurement.
+
+**Measurement category:**
+
+-  *MeasurementName*: A brief description of the measurement.
+-  *MeasurementName*: A brief description of the measurement.
+
+This section can be omitted if no measurements are made by this module.
```
AetherUnbound: I don't think this belongs in this section, as I assume measurement modules make measurements (perhaps that's an improper assumption).

cellprofiler/modules/plugins/measurementtemplate.py
```diff
@@ -64,14 +152,14 @@
...
-# you re-implement them. You can let CPModule do most of the work and
+# Your module should "inherit" from cellprofiler.module.Module.
+# This means that your module will use the methods from Module unless
+# you re-implement them. You can let Module do most of the work and
 # implement only what you need.
 #
 ###################################
 
-class MeasurementTemplate(cpm.Module):
+class MeasurementTemplate(cellprofiler.module.Module):
```
AetherUnbound: Would it make sense to have a `MeasurementProcessing` class to inherit from?

cellprofiler/modules/plugins/measurementtemplate.py
```diff
@@ -148,8 +244,11 @@ def create_settings(self):
...
-        return [self.input_image_name, self.input_object_name,
-                self.radial_degree]
+        return [
+            self.input_image_name,
+            self.input_object_name,
+            self.radial_degree
+        ]
 
     #
     # CellProfiler calls "run" on each image set in your pipeline.
```
AetherUnbound: I noticed you changed these lines in the other template. Should they be changed here too?

cellprofiler/modules/plugins/measurementtemplate.py
```diff
@@ -257,39 +358,38 @@ def run(self, workspace):
...
                 # Do the same with -m
-                #
                 feature = self.get_measurement_name(n, -m)
-                meas.add_measurement(input_object_name, feature, zi)
+                measurements.add_measurement(input_object_name, feature, zi)
             else:
                 # For zero, the total is the sum of real and imaginary parts
-                meas.add_measurement(input_object_name, feature, zr + zi)
-            #
+                measurements.add_measurement(input_object_name, feature, zr + zi)
```
AetherUnbound: It looks like you're adding several values for a feature with the same name. Do you mind describing this behavior?

cellprofiler/modules/plugins/measurementtemplate.py
```diff
@@ -257,39 +358,38 @@ def run(self, workspace):
...
             statistics.append([feature, zmean, zmedian, zsd])
 
     ################################
     #
-    # DISPLAY
+    # "display" lets you use matplotlib to display your results.
     #
-    def display(self, workspace, figure=None):
+    ################################
+    def display(self, workspace, figure):
```
AetherUnbound: It seems like this has to be overridden for all measurement modules. Does it make sense to standardize this processing/displaying like it's been done in the `ImageProcessing` and `ObjectProcessing` modules, or are the implementations too specific?

cellprofiler/modules/plugins/measurementtemplate.py
```diff
@@ -347,22 +448,25 @@ def measure_zernike(self, pixels, labels, indexes, centers, radius, n, m):
...
+        radius = numpy.hstack([[1], radius])
+
         #
         # Now get one array that's the y coordinate of each pixel and one
         # that's the x coordinate. This might look stupid and wasteful,
         # but these "arrays" are never actually realized and made into
         # real memory.
         #
-        y, x = np.mgrid[0:labels.shape[0], 0:labels.shape[1]]
+        y, x = numpy.mgrid[0.0:labels.shape[0], 0.0:labels.shape[1]]
```
AetherUnbound: Why the `0.0`?

cellprofiler/modules/plugins/measurementtemplate.py
```diff
@@ -454,8 +556,7 @@ def get_measurement_name(self, n, m):
     # of measurement columns produced. The second is more informal and
     # tells CellProfiler how to categorize its measurements.
     #
-    #
-    # get_measurement_columns gets the measurements for use in the database
+    # "get_measurement_columns" gets the measurements for use in the database
     # or in a spreadsheet. Some modules need this because they
     # might make measurements of measurements and need those names.
     #
```
AetherUnbound: I can't add the comment on the exact line, but line 570 references `cpmeas` which should be expanded to the full name.

cellprofiler/modules/plugins/measurementtemplate.py
```diff
@@ -474,62 +575,57 @@ def get_measurement_columns(self, pipeline):
...
         For the Zernikes, the scales are of the form, N2M2 or N2MM2 for
         negative azimuthal degree
         '''
-        if image_name in self.get_measurement_images(
-                pipeline, object_name, category, measurement):
-            return [("N%dM%d" % (n, m)) if m >= 0 else
-                    ("N%dMM%d" % (n, -m)) for n, m in
-                    self.get_zernike_indexes(True)]
+        if image_name in self.get_measurement_images(pipeline, object_name, category, measurement):
+            return [("N%dM%d" % (n, m)) if m >= 0 else ("N%dMM%d" % (n, -m)) for n, m in self.get_zernike_indexes(True)]
```
AetherUnbound: This could probably be split up for readability.

cellprofiler/modules/plugins/imagetemplate.py
mcquin: Right. A mask obstructs unwanted pixels. It's a boolean matrix where `False` represents "masked" pixels (pixels we don't want to consider) and `True` represents "unmasked" pixels (pixels we want to consider). I'll clarify this.

AetherUnbound: AH! I was thinking about masks the wrong way. I thought "masked" data was the data you cared about.

cellprofiler/modules/plugins/imagetemplate.py
mcquin: Good question. This is intended to be a very simple template. Accessing other methods or instance variables is beyond the scope of this example.

I'm working on a guide for writing modules. I expect to include information on writing a module with a function that needs state (`self`).

cellprofiler/modules/plugins/imagetemplate.py
mcquin: Oof. I'll clarify.

cellprofiler/modules/plugins/measurementtemplate.py
mcquin: Hmm... I'm not sure. We'd need to define the commonalities between measurement modules (beyond "they make measurements"). Currently, I'm not sure what abstractions would be useful. I'm open to ideas.

AetherUnbound: I haven't written any measurement modules, so I'm not sure what could be refactored either :) I'll keep an open mind if I work on them in the future!

cellprofiler/modules/plugins/measurementtemplate.py
mcquin: Fun trick: it makes the data have floating-point values, rather than integers. It's needed for a `+=`'s below, which requires matching datatypes.

cellprofiler/modules/plugins/imagetemplate.py
```diff
@@ -372,8 +375,12 @@ def gradient_image(pixels, gradient_choice, automatic_smoothing, scale):
...
         gradient_image = gradient_magnitude
     else:
-        # Numpy uses i and j instead of x and y. The x axis is 1
-        # and the y axis is 0
+        # Image data is indexed by rows and columns, with a given point located at
+        # position (row, column). Here, x represents the column coordinate (at index 1)
+        # and y represents the row coordinate (at index 0).
+        #
+        # You can learn more about image coordinate systems here:
+        # http://scikit-image.org/docs/dev/user_guide/numpy_images.html#coordinate-conventions
```
AetherUnbound: :raised_hands:

cellprofiler/modules/plugins/measurementtemplate.py
mcquin: Would you like me to clarify why `zr`, `zi` are arrays, and how that works with respect to saving per-object measurements in CellProfiler?

cellprofiler/modules/plugins/measurementtemplate.py
mcquin: Great question. A quick look shows me that most measurement modules only display `workspace.statistics` (there are a few exceptions). That could be made the default behavior in a measurement module superclass.

cellprofiler/modules/plugins/measurementtemplate.py
mcquin: I've added a comment explaining this.

cellprofiler/modules/plugins/measurementtemplate.py
mcquin: There's some `np.`s in there, too.

cellprofiler/modules/plugins/measurementtemplate.py
mcquin: 👍 Done https://github.com/CellProfiler/CellProfiler/pull/3429/commits/2c7c53c7e83a0b7bbac9fba6a02963b58c1accbb

cellprofiler/modules/plugins/measurementtemplate.py
AetherUnbound: I just realized that `feature` is changing for every measurement you add. Nope, this should be good :)

PR #3434, opened 2017-12-26
Re-arrange Tile help settings to correspond with settings order

### LINKED ISSUE #3292: Documentation - Tile, arrange settings in help nicely
Right now, _Automatically calculate number of rows?_ (and columns) is at the end of the help, but it's a question shown towards the beginning of the GUI, so it is a little confusing when looking at the help. 

As a more general question, is there a reason that the help sections for individual settings are not automatically shown in the order they are in the GUI? It'd be awesome if that just happened magically, but if not we could have someone systematically go through and arrange them. I've noted a couple of cases that were obvious from just reading the help but we could go through them all if needed.

AetherUnbound:
Resolves #3292

### DIFF
--- a/cellprofiler/modules/tile.py
+++ b/cellprofiler/modules/tile.py
@@ -268,8 +268,6 @@ create a grid that has roughly the same number of rows and columns.
                   self.place_first, self.tile_style,
                   self.meander]
 
-        for additional in self.additional_images:
-            result += [additional.input_image_name]
         return result
 
     def is_aggregation_module(self):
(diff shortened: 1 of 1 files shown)


cellprofiler/modules/tile.py
```diff
@@ -261,6 +261,17 @@ def visible_settings(self):
...
         return result
 
+    def help_settings(self):
+        result = [self.input_image, self.output_image, self.tile_method,
+                  self.wants_automatic_rows, self.rows,
+                  self.wants_automatic_columns, self.columns,
+                  self.place_first, self.tile_style,
+                  self.meander]
+
+        for additional in self.additional_images:
```
mcquin: We can omit this `for` loop. It will display text like:

```
Select an additional image to tile 
Select an additional image to tile? 
```

as many times as there are additional images.

PR #3449, opened 2018-01-08
Improve visualization for low-contrast images.

### LINKED ISSUE #3396: Improve visualization for low-contrast images
A couple different CP power users/friends of the lab have now reported so on [the forum](http://forum.cellprofiler.org/t/cp3-log-normalized-gives-black-image/5288/3).

image

mcquin:
Resolves #3396.

Adds gamma adjustment/power law transformation. This feature can
be enabled via the figure menu or in the preferences dialog.

Additionally, allows configuring the value of gamma and the scale
of the log transformation in the preferences dialog, or when
the normalization option is selected from the figure menu.

Examples:
screen shot 2018-01-08 at 1 43 50 pm

screen shot 2018-01-08 at 1 44 54 pm

### DIFF
--- a/cellprofiler/gui/preferencesdlg.py
+++ b/cellprofiler/gui/preferencesdlg.py
@@ -16,6 +16,7 @@ FILEBROWSE = "FileBrowse"
 FONT = "Font"
 COLOR = "Color"
 CHOICE = "Choice"
+TEXT = "Text"
 
 
 class IntegerPreference(object):
@@ -289,7 +290,7 @@ class PreferencesDlg(wx.Dialog):
                 ["Intensity normalization factor",
                  cellprofiler.preferences.get_normalization_factor,
                  cellprofiler.preferences.set_normalization_factor,
-                 "Text",
+                 TEXT,
                  cellprofiler.preferences.NORMALIZATION_FACTOR_HELP],
                 ["CellProfiler plugins directory",
                  cellprofiler.preferences.get_plugin_directory,
--- a/cellprofiler/gui/preferencesdlg.py
+++ b/cellprofiler/gui/preferencesdlg.py
@@ -119,8 +119,7 @@ class PreferencesDlg(wx.Dialog):
                 current = getter()
                 if current is None:
                     current = ""
-                ctl = wx.TextCtrl(scrollpanel, -1, current,
-                                  validator=validator)
+                ctl = wx.TextCtrl(parent=scrollpanel, id=-1, value=current, validator=validator)
                 min_height = ctl.GetMinHeight()
                 min_width = ctl.GetTextExtent("Make sure the window can display this")[0]
                 ctl.SetMinSize((min_width, min_height))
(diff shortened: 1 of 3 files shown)


cellprofiler/gui/preferencesdlg.py
```diff
@@ -282,9 +281,16 @@ def get_preferences(self):
...
-                  cellprofiler.preferences.INTENSITY_MODE_LOG],
+                 [cellprofiler.preferences.INTENSITY_MODE_RAW,
+                  cellprofiler.preferences.INTENSITY_MODE_NORMAL,
+                  cellprofiler.preferences.INTENSITY_MODE_LOG,
+                  cellprofiler.preferences.INTENSITY_MODE_GAMMA],
                  cellprofiler.preferences.INTENSITY_MODE_HELP],
+                ["Intensity normalization factor",
+                 cellprofiler.preferences.get_normalization_factor,
+                 cellprofiler.preferences.set_normalization_factor,
+                 "Text",
```
AetherUnbound: Should this be a different constant like `DIRBROWSE` or `CHOICE`, or is it fine to stay "Text"?

cellprofiler/preferences.py
```diff
@@ -625,6 +626,24 @@ def resources_root_directory():
 directly from the server).\
 """
 
+NORMALIZATION_FACTOR_HELP = """\
+Sets the normalization factor for intensity normalization methods:
+
+-  *{INTENSITY_MODE_LOG}*: Set the gain applied to the each pixel in the image during log normalization. Pixels are
+   transformed according to the formula O = gain * log(1 + I), where I is the pixel intensity.
+-  *{INTENSITY_MODE_GAMMA}*: Set the value of gamma. Pixels are transformed according to the formula O = I ** gamma,
```
AetherUnbound: Could the equation portion of this docstring be in backtics for readability?

cellprofiler/preferences.py
```diff
@@ -1840,6 +1859,21 @@ def set_choose_image_set_frame_size(w, h):
     config_write(CHOOSE_IMAGE_SET_FRAME_SIZE, "%d,%d" % (w, h))
 
 
+__normalization_factor = "1.0"
+
+
+def get_normalization_factor():
```
AetherUnbound: It might be a good idea to refactor the preferences here to be a class with `classmethod` properties (that seems like it should be possible at least). It also looks like there are a lot of other opportunities for refactoring in `preferences.py` :smile:

cellprofiler/preferences.py
```diff
@@ -625,6 +626,24 @@ def resources_root_directory():
 directly from the server).\
 """
 
+NORMALIZATION_FACTOR_HELP = """\
+Sets the normalization factor for intensity normalization methods:
+
+-  *{INTENSITY_MODE_LOG}*: Set the gain applied to the each pixel in the image during log normalization. Pixels are
```
bethac07: Can we add a sentence here, similar to the one below, indicating higher values lead to brighter images?  Just for the equation-averse.  Otherwise, I love it!

cellprofiler/gui/preferencesdlg.py
mcquin: I'll create a constant, for consistency.

cellprofiler/preferences.py
mcquin: So many opportunities! These can be tackled later.

cellprofiler/preferences.py
mcquin: Great idea! I'll add that.

Issue #1022, opened 2014-01-17
Database deadlock


LeeKamentsky:
From Rob Ralston:
The problem with database locking has occurred only with the Windows version; the CentOS version has not had this trouble.
Today I downloaded the new release candidate that Lee Kamentsky sent out and tested the problem pipeline again. 

The problem was reproducible with the release candidate and actually was worse than with the previous beta version. I have attached two Word documents with screen shots and the error output. I also exported .csv versions of selected columns from Per_Image tables from the MySQL output (attached). There are 87 image pairs (numbered 1-87) the tables show that some images could be accessed and processed (I did not include the data columns), but there was extensive skipping (e.g., images processed were 1, 5,6,7,9,13 ..., images 2,3,4,8,10,11,12 ... were skipped do to DB locking). The images were in a DropBox folder, so I did a second run with images in a folder on the desktop, but had the same problem.

I ran a related pipeline on a larger image set, 311 image pairs including the 87 pairs mentioned above, with output to .csv files. This pipeline used the legacy image loading via a .csv metadata file and output the data into .csv files rather than to SQL. This pipeline ran very nicely (I really really appreciate your work on the multithreading!) with no problems. Testing a subset of the data showed the results were the same as obtained previously with r11710 on Win7 or r11052 on Linux.

I am running CP2.1 release candidate using an i5 processor with 8GB RAM on Windows 7 home premium OEM. The MySQL version is 5.5.28 winx64. This is a dual boot machine: I run CP2.0 r11052 on Linux Mint 13 on the same machine and also CP2.1 beta on CentOS 6.5 (via a SDHC card as mentioned previously). All of the Linux programs, including your CentOS version can write to the MySQL database (linux version in this case) without any problems. When I ran CP r11710 on the Win7 side, it also wrote data to MySQL without problems. 

I will check the SHOW PROCESSLIST command as you suggest and try a few other tests to see if I can isolate the specific problem.

Thanks again for your work on CellProfiler and making it available across multiple platforms. I really like the program.

Rob Ralston

PR #3456, opened 2018-01-16
Use `READ COMMITTED` isolation in place of coarse grained lock

chris-allan:
In a highly concurent batch processing environment there can easily be
100s of instances of CellProfiler.  With a decently complicated
pipeline and images collected with large frame (2K x 2K+) cameras there
can also easily be several 100 objects per image.  Under these
conditions MySQL lock timeouts will occur when using the InnoDB database
engine which is the default as of MySQL 5.7.

The coarse grained lock was instituted in 26c3841 as a fix for #1022.
While that issue does not detail the specifics it can be assumed that
the deadlocks mentioned were related to gap locking:

  https://dev.mysql.com/doc/refman/5.7/en/innodb-locking.html#innodb-gap-locks

The propensity for InnoDB to be affected by deadlocking problems is well
known.  So much so that there is a page dedicated to it in the MySQL
manual:

  https://dev.mysql.com/doc/refman/5.7/en/innodb-deadlocks.html

The ExportToDatabase module really executes no interdependent SELECT
statements in each corresponding transaction.  So, solving deadlocking
issues can be done much more elegently and concurrently by selecting an
appropriate level of transaction isolation for MySQL.  These levels are
detailed here:

  https://dev.mysql.com/doc/refman/5.7/en/innodb-transaction-isolation-levels.html

The level `READ COMMITTED` has been specifically chosen due to the
lack of interdependency in CellProfiler batch processing workflows and
the fact that gap locking is *disabled* for this isolation level.

### DIFF
--- a/cellprofiler/modules/exporttodatabase.py
+++ b/cellprofiler/modules/exporttodatabase.py
@@ -301,6 +301,11 @@ def connect_mysql(host, user, pw, db):
     '''Creates and returns a db connection and cursor.'''
     connection = MySQLdb.connect(host=host, user=user, passwd=pw, db=db)
     cursor = SSCursor(connection)
+
+    rv = cursor.execute('SET TRANSACTION ISOLATION LEVEL READ COMMITTED')
+    logger.info('Set MySQL transaction isolation to "READ COMMITTED": %r' % rv)
+    cursor.execute('BEGIN')
+
     #
     # Use utf-8 encoding for strings
     #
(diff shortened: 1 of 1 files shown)


chris-allan:
There were several days of work and lots of cluster time spent that went into researching and testing this change by @emilroz. During which the changes made by @0x00b1, @mcquin, @bethac07, and @AetherUnbound to the readability, testability, and general cleanliness of this module since 2.2.0 were essential. Thanks a lot!

chris-allan:
An example of the exceptions you are likely to see without this change when running CellProfiler under the conditions outlined in the PR description:

 * http://forum.cellprofiler.org/t/mysql-exporttodatabase-error-when-running-multiple-jobs-on-cluster/6205

Specifically:

```
...
OperationalError: (1205, ‘Lock wait timeout exceeded; try restarting transaction’)
...
```

PR #3483, opened 2018-02-05
Fix image resource paths

### LINKED ISSUE #3458: [Ubuntu 16.04] Odd behavior when determining relative image paths
I'm getting inconsistent behavior between my work machine and my laptop when attempting to load documentation images. Things seem to work fine as-is on my work machine, but on my desktop none of the images display properly. 

They _do_ display correctly when I modify L241 of `_help.py` to be
```
    return "./data/images/{}".format(filename)
```
instead of (current master)
```
    return "../images/{}".format(filename)
```

This was discovered when troubleshooting (and may be related to) #3454

mcquin:
Resolves #3458

### DIFF
--- a/cellprofiler/modules/_help.py
+++ b/cellprofiler/modules/_help.py
@@ -4,7 +4,7 @@ import os.path
 
 import pkg_resources
 
-import cellprofiler.setting
+import cellprofiler.gui.help.content
 from cellprofiler.preferences import ABSOLUTE_FOLDER_NAME, DEFAULT_INPUT_SUBFOLDER_NAME, DEFAULT_OUTPUT_SUBFOLDER_NAME
 
 DEFINITION_OBJECT = """\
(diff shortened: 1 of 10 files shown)


cellprofiler/modules/_help.py
```diff
@@ -220,29 +217,10 @@
...
-    )) == os.path.join("cellprofiler","data", "images", filename):
-        return os.path.relpath(pkg_resources.resource_filename(
-            "cellprofiler",
-            os.path.join("data", "images", filename)
-        ))
 
-    # Otherwise, we are probably building the documentation in sphinx.
-    # The path separator used by sphinx is "/" on all platforms.
-    return "../images/{}".format(filename)
+# TODO: Move these too??
```
AetherUnbound: Where would a better place for this be? It seems to be a common help resource, so it makes sense to me to leave it here.

mcquin: These images are only used in the context of modules, so I decided to leave these variables here. I will remove the TODO comment.

PR #3479, opened 2018-02-03
link DOIs to new, secure resolver

katrinleinweber:
See #3478 :-)   I used Atom's project-wide search-and-replace, so I guess I caught all them DOIs.

Issue #3478, opened 2018-02-03
update DOI links to new resolver 


katrinleinweber:
The DOI foundation [updated their recommendation](https://www.doi.org/doi_handbook/3_Resolution.html#3.8) from `…dx.doi.org/` to `https://doi.org/`.

PR #3487, opened 2018-02-06
Follow-up to #3478

katrinleinweber:
Hello! I think that PR was a bit too quickly merged, because [I wanted revert an intermediate work-around](https://github.com/CellProfiler/CellProfiler/pull/3479#discussion_r166345745). So, this can revert 94f215df because patch d83bd941 was included in the mean-time. Cheers :-)


mcquin:
My apologies, I didn't see that comment on the commit. Thanks for the fix!

katrinleinweber:
You're welcome :-)

PR #3490, opened 2018-02-07
Remove unused, yet created, wxIDs.

### LINKED ISSUE #3461: C++ wx assertion errors pop up randomly during analysis runs
I've encountered this once or twice myself, but here are [some](http://forum.cellprofiler.org/t/cellprofiler-3-windows-10-large-dataset/5521/3) from [the](http://forum.cellprofiler.org/t/erro-message-c-assertion/5491) forum [recently](http://forum.cellprofiler.org/t/c-assertion-error-cp-3-0/5554); in only case did someone capture things from the terminal after (bottom snippet) but I've transcribed as best I can in the top snippet.  It's popped up in Windows 7, Mac, and Windows 10; the common thread seems to be that the pipelines run fine on smaller batches but hit this error on larger ones.  For at least 2 users running the pipeline with the display enabled ("eyes open") vs closed seemed to make a difference, one user said it did not and he had to decrease his batch size.  

I hadn't reported it before because it's an intermittent issue that apparently only crops up on large batch sizes, which I assumed would make it hard to debug and troubleshoot, but I realized it should probably be documented and hopefully you guys can do your magic and find it. 

image
image
image

```
C++ assertion "(itemid >=0 && itemid <SHRT_MAX) ||  (itemid>=wxID_AUTO_LOWEST && itemid <= wxID_AUTO_HIGHEST)" 
failed at ..\..\src\common\menucmn.cpp(260) in wxMenuItemBase::wxMenuItemBase(): invalid item value
```

```
File “CellProfiler\cellprofiler\gui\cpframe.py”, line 483, in OnClose
File “CellProfiler\cellprofiler\gui\pipelinecontroller.py”, line 1333, in check_close
File “site-packages\wx-3.0-msw\wx_controls.py”, line 193, in init
wx._core.PyAssertionError: C++ assertion “id == wxID_ANY || (id >= 0 && id < 32767) || (id >= wxID_AUTO_LOWEST && id <= wxID_AUTO_HIGHEST)” failed at …\src\common\wincmn.cpp(372) in wxWindowBase::CreateBase(): invalid id value
```

mcquin:
Resolves #3461 

These IDs were not used but were being created every time the module was displayed.
They were not being cleaned up, and so caused an overflow error when allocated new IDs.

### DIFF
--- a/cellprofiler/gui/namesubscriber.py
+++ b/cellprofiler/gui/namesubscriber.py
@@ -42,7 +42,6 @@ class NameSubscriberComboBox(wx.Panel):
         if choices is None:
             choices = []
         self.orig_choices = choices
-        self.IDs = [wx.NewId() for c in choices]
         sizer = wx.BoxSizer(wx.HORIZONTAL)
         self.combo_dlg = wx.ComboBox(
                 self, choices=[choice[0] for choice in choices],
@@ -65,7 +64,6 @@ class NameSubscriberComboBox(wx.Panel):
         for child in self.combo_dlg.Children:
             # Mac implements read_only combobox as a choice in a child
             child.Bind(wx.EVT_RIGHT_DOWN, self.right_menu)
-        self.Bind(wx.EVT_MENU, self.menu_selected)
         self.callbacks = []
 
     def choice_made(self, evt):
@@ -97,12 +95,8 @@ class NameSubscriberComboBox(wx.Panel):
         menu = wx.Menu()
         all_menu = wx.Menu()
 
-        unsorted_choices = [
-            (name, annotation, num, is_input_module, identifier)
-            for (name, annotation, num, is_input_module), identifier in
-            zip(self.orig_choices, self.IDs)]
         fn_key = lambda x: (x[2], x)
-        choices_sorted_by_num = sorted(unsorted_choices, key=fn_key)
+        choices_sorted_by_num = sorted(self.orig_choices, key=fn_key)
         for name, annotation, num, is_input_module, choiceid in \
                 choices_sorted_by_num:
             all_menu.Append(choiceid, "filler")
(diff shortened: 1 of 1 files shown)


caticoa3:
I'm experiencing the same error mentioned in issue #3461. I installed CellProfiler version 3.1.3 and reinstalled today, before writting this message to make sure I had the latest of the master branch. From what I see above the correction has been merged into master. I'm by no means a expert at github so maybe perhaps Im wrong and its not included in the master yet. 
I used Conda to install following the instructions provided on the wiki. I am working on Linux (Ubuntu 16.04). My computer hsa a 3.4GHz quad core AMD A10, 8GB of ram and 8GB swap.  

I'm using CellProfiler to correcting uneven illumination of labeled cells and their nuclei.

First, I have about 148 errors from duplicate images or missing channels. 
eg, Error for image set, channel=6.22.08 / MAP2 / P / 9b / 2 / , metadata=CY3_or_CY5: Duplicate entries
Error for image set, channel=6.22.08 / MAP2 / P / 9b / 3 / , metadata=CY3_or_CY5: Duplicate entries
Error for image set, channel=6.22.08 / RIP /  / 18a / 3 / , metadata=DAPI: Missing from channel   

I understand the errors are occuring to avoid creating identical or incomplete image sets. So, I ignore these and the process continues. 

I then have are about 530 images sets that begin processing. Then, another error occurs:
broadcast error.
I think this is occuring because some of my images are probably not the same dimensions. I'll need to check on this. Therefore, I cancel processing for all image sets with this errors.
After canceling intemitently images due to this error the process is about 480 sets into processing, I get the error mentioned in #3461  

cellprofiler error
traceback 2018-05-04
Kernel process accumulate/increase in swap as images set are processed. Most of the swap is consumed by the time the error mentioned in #3461 occurs.
htop at time of error 3641 2018-05-04

I selected "Cancel" for each image that experienced the #3461 error, until the end of the batch. After that the CellProfiler would not exit post-processing.

Thanks for reading

PR #3495, opened 2018-02-09
Remove unused ImageSetCache

### LINKED ISSUE #3493: Remove measurements' ImageSetCache from stable
It's unused and may be incompatible with the latest versions of h5py. From https://github.com/CellProfiler/CellProfiler/pull/3455#issuecomment-357985250:

> The image set cache is broken with versions of h5py >= 2.7.0. It is implied in #2229 that while there are tests targetting it, the class isn't itself actually used anywhere. Lacking more information I have pinned the h5py version at < 2.7.0.

Additionally, the ImageSetCache was removed in 3.0.0 in #2229.

This should be done before releasing 2.2.2.

mcquin:
Resolves #3493


mcquin:
> Anything else barring the 2.2.2 release?

Not that I can think of!

PR #3496, opened 2018-02-11
Include example pipeline as a package resource.

### LINKED ISSUE #803: Load a simple pipeline URL on introduction page fails when using a proxy server
We use a proxy server for our internet access which causes the "Load a simple pipeline" link on the "Welcome to CellProfiler" window to fail (see screenshot). I'm on OSX 10.8.4 and CP 2.1 rev 7a43182.

image

### LINKED ISSUE #3007: Remove internet access requirement when loading example pipeline from Welcome screen
Is there a good reason why we don't just package the three images + example pipeline when releasing CP? It ensures a match between version of CP and pipeline version, and works even without internet access. It would allow us to tear out a fair bit of code I'd imagine.

I'm not sure why we set it up this way: possibly because we thought it'd be nice to be able to update the example all the time... but it turns out we never make substantive changes to it anymore other than updates to version changes.

mcquin:
Resolves #3007 
Resolves #803


AetherUnbound:
Since `setup.py` pulls everything in from `data/` two levels deep, the examples folder should be pulled when setup runs. However, when I try loading the pipeline from the welcome screen I still get the images being sourced from `githubusercontent`.

mcquin:
Nice catch. I'll ensure the images can be fetched without an internet
connection as well.

PR #3500, opened 2018-02-15
Add explicit volumetric support to CalculateMath

### LINKED ISSUE #3499: CalculateMath should support 3D
It just does math on measurements that are per-image or per-object, it shouldn't be anything 2D or 3D specific in it... I think?

bethac07:
Resolves #3499


bethac07:
Resolves #3499

PR #3514, opened 2018-03-06
Backport s3 URL support to stable

mcquin:


cellprofiler/modules/images.py
```diff
@@ -256,6 +256,9 @@ def prepare_run(self, workspace):
                 expression, ifcls)
             file_array = env.make_object_array(len(file_list), scls)
             for i, url in enumerate(file_list):
+                if url.startswith("s3:"):
+                    url = url.replace(" ", "+")
```
AetherUnbound: Why does this have to get removed and then replaced?

mcquin: Great question. Essentially, `" "` characters cause URI validation to fail. s3 links don't work when you %-encode whitespace within the resource name. s3 uses `+` to encode whitespace (and %-encodes `+` if it appears in the bucket or resource name), so it's safe to use here.

PR #3515, opened 2018-03-07
UTF fixes

### LINKED ISSUE #2538: remove `cellprofiler.utilities.utf16encode`

AetherUnbound:
Resolves #2538 
Pipeline files will now be written out as version `4`, which will be in strictly string-escaped utf-16 format (rather than the homebrewed utf-16 conversion). 

Unit tests were adjusted to reflect these changes.

### DIFF
--- a/cellprofiler/pipeline.py
+++ b/cellprofiler/pipeline.py
@@ -138,7 +138,7 @@ FMT_MATLAB = "Matlab"
 FMT_NATIVE = "Native"
 
 '''The current pipeline file format version'''
-NATIVE_VERSION = 3
+NATIVE_VERSION = 4
 
 '''The version of the image plane descriptor section'''
 IMAGE_PLANE_DESCRIPTOR_VERSION = 1
@@ -880,6 +880,7 @@ class Pipeline(object):
             raise NotImplementedError('Invalid header: "%s"' % header)
         version = NATIVE_VERSION
         from_matlab = False
+        do_deprecated_utf16_decode = False
         do_utf16_decode = False
         has_image_plane_details = False
         git_hash = None
@@ -898,7 +899,9 @@ class Pipeline(object):
                 version = int(value)
                 if version > NATIVE_VERSION:
                     raise ValueError("Pipeline file version is {}.\nCellProfiler can only read version {} or less.\nPlease upgrade to the latest version of CellProfiler.".format(version, NATIVE_VERSION))
-                elif version > 1:
+                elif version > 1 and version < 4:
+                    do_deprecated_utf16_decode = True
+                elif version >= 4:
                     do_utf16_decode = True
             elif kwd in (H_SVN_REVISION, H_DATE_REVISION):
                 pipeline_version = int(value)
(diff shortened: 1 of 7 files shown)


cellprofiler/pipeline.py
```diff
@@ -1026,6 +1031,8 @@ def rl():
                 attribute_strings = attribute_string[1:-1].split('|')
                 variable_revision_number = None
                 # make batch_state decodable from text pipelines
+                # NOTE: These variables are **necessary**, even though they aren't used anywhere obvious
+                # Removing them **will*
```
popiel: This comment looks incomplete.

tests/test_pipeline.py
```diff
@@ -663,28 +663,39 @@ def test_13_03_save_long_measurements(self):
...
         fd = cStringIO.StringIO()
         pipeline.savetxt(fd, save_image_plane_details=False)
         result = fd.getvalue()
         lines = result.split("\n")
-        self.assertEqual(len(lines), 10)
+        self.assertEqual(len(lines), 11)
+        text, value = lines[-3].split(":")
+        #
+        # unicode encoding:
+        #     backslash: \\ (BOM encoding)
```
popiel: The BOM is separate from the backslash, and should be presented on its own line in this comment.

tests/test_pipeline.py
```diff
@@ -663,28 +663,39 @@ def test_13_03_save_long_measurements(self):
...
+        #     unicode character: \u
```
popiel: I think you mean \u2211 here.  BTW, that's the n-ary summation character, if you want to name it (as you did for the backslash above).

0x00b1:
@AetherUnbound Amazing! The problem with a number of tests is the use of a string rather than byte literal coupled with the use of `encode` rather than `decode`. I believe this can be fixed by prepending a `b` to the `value` string and replacing `encode` with `decode`, e.g.

```Python
>>> value = b"\xff\xfe/\x00t\x00m\x00p\x00/\x00c\x00p\x00_\x00e\x00x\x00a\x00m\x00p\x00l\x00e\x00i\x00m\x00a\x00g\x00e\x00s\x00A\x00K\x00V\x00C\x00N\x00Y\x00"
>>> value.decode("utf-8")
"/tmp/cp_exampleimagesAKVCNY"
```

cellprofiler/preferences.py
```diff
@@ -206,7 +205,7 @@ def config_read(key):
         value = None
     if value is not None:
         try:
-            value = utf16decode(value)
+            value = cellprofiler.utilities.utf16encode.utf16decode(value)
```
0x00b1: Could this be replaced with the corresponding Python method?

```Python
value = value.encode("utf-16")
```

cellprofiler/preferences.py
```diff
@@ -228,7 +227,7 @@ def config_write(key, value):
         shutup = wx.LogNull()
     __cached_values[key] = value
     if value is not None:
-        value = utf16encode(value)
+        value = value.encode('utf-16')
```
0x00b1: Could this be replaced with the corresponding Python method?

```Python
value = value.decode("utf-16")
```

cellprofiler/preferences.py
AetherUnbound: I think so! I can try that

cellprofiler/preferences.py
AetherUnbound: I think this is supposed to be encoded so that it's written out in byte format here...although I may be swapping encode/decode

AetherUnbound:
I'll give your suggestion a shot WRT the unit tests!

tests/test_pipeline.py
AetherUnbound: How is the BOM separate from the two backslashes here?

Issue #1992, opened 2016-04-29
Can't select Illumination Function or Objects NamesAndTypes set to All Images 


dlogan:
Master on Mac
Open a blank pipeline.
In NamesAndTypes, leave "Assign a name to:" = "All images"
You can see that there are only three image types to choose from:
![image](https://cloud.githubusercontent.com/assets/1225595/14923321/8ac7c7a4-0e0a-11e6-9376-46a17f11cf03.png)

Now in NamesAndTypes, change "Assign a name to:" = "Grayscale image"
You can see that there are now five image types to choose from:
![image](https://cloud.githubusercontent.com/assets/1225595/14923330/99ce12bc-0e0a-11e6-9fde-c901d6e85ac6.png)

Maybe this is a feature, but I can't think why.  Probably should show all 5.

braymp:
I think the idea there was to omit more esoteric image types from the simple "I just have one channel" view.

dlogan:
Hmm, I see.  I think it's preferable to show them all -- the default is Grayscale anyway which aligns with the simplest use-case.  I was writing up instructions for folks using ilastik and they might be loading label matrices as inputs, and I could imagine someone wanting to load only Objects.  

I'd prefer to do this sort of simplifying when (if?) we implement the Simple/Advanced Mode, but am ok with ignoring this if others don't agree.

AnneCarpenter:
I think it makes no sense to choose "Illumination Function" if you're choosing all images. I don't see why Objects would be intentionally excluded from the list, though. 

So, for simplicity, it sounds like we are all supportive of putting all 5 options in.

DavidStirling:
I've been looking into this (better late than never!), but it seems that it's not trivial to allow using 'Objects' when not matching with rules. "Assign to everything" mode works very differently and would need a substantial rewrite to be compatible with using objects as well. Since this can easily be worked around by creating a generic rule, I've just added a note to the help text explaining this.

PR #3522, opened 2018-03-19
Allow pre-labeled images to be used in ConvertImageToObjects

AetherUnbound:
If the objects from a pipeline were saved, there was no way to easily import them into a different pipeline as objects. This extends `ConvertImageToObjects` to (logically) convert an integer labeled array into objects. This conversion respects previous labels - it doesn't force to connected but separately labeled objects into a single, shared label.

### DIFF
--- a/cellprofiler/modules/convertimagetoobjects.py
+++ b/cellprofiler/modules/convertimagetoobjects.py
@@ -5,6 +5,7 @@ import skimage.measure
 import cellprofiler.module
 import cellprofiler.setting
 
+
 HELP_BINARY_IMAGE = """\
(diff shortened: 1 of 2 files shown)


bethac07:
>If the objects from a pipeline were saved, there was no way to easily import them into a different pipeline as objects. 

This is not true.  You can designate them as "Objects" in NamesAndTypes.  Unless this is doing something different this functionality is redundant.

AetherUnbound:
That's incredibly good to know! It seems that option is not exposed by default though, only when you filter on image matching certain rules.

AetherUnbound:
This has also been mentioned in #1992, and it might be removable once #3523 gets resolved

PR #3528, opened 2018-03-27
Make the display for labeled images show label index rather than RGB value

### LINKED ISSUE #3521: 3D labeled objects (segmentations) show RGB values not intensity value
3D labeled objects (and perhaps all objects?) show RGB values for segmentations, not intensity values. This can be frustrating when trying to determine if objects before and after a step have the same labels. 

Left side shows intensity:
screenshot_2018-03-19_11-01-20

Right side shows RGB:
screenshot_2018-03-19_11-01-35

AetherUnbound:
Resolves #3521 

Previously, modules that used the reasonably named `subplot_imshow_labels` function would ultimately display the labels as RGB values rather than intensity values. This is supremely unhelpful in most cases where one wants to know the indexed label given to a certain blob (e.g. RelateObjects). 

This PR adds a custom colormap to `subplot_imshow_labels` when a background image is not supplied that allows the background of the labeled image to remain black while still showing the intensity values of the given objects. I had to play around with different colormaps to find the right fit - if there were 200+ objects in an image, adjacent objects would be essentially indistinguishable with the default colormap. I ultimately found the [prism](https://matplotlib.org/examples/color/colormaps_reference.html) colormap with a _higher_ frequency gave that "random" coloring that was desired without having to randomize the actual array in RGB space. 

Examples **before**:
screenshot_2018-03-26_18-16-16
screenshot_2018-03-26_18-17-21

Examples **after**:
screenshot_2018-03-26_18-18-45
screenshot_2018-03-26_18-19-56


AnneCarpenter:
If I understand correctly, you are providing an additional option for a colormap, or does it replace the current colormap? 

I don't even remember if there are multiple colormap options allowed in the first place! But your description makes sense to me - definitely we want to be able to distinguish adjacent objects well. Sounds like you've tested the things I would've suggested -  making sure it works for a wide range of object numbers. 

I think the new colors are tasty-looking. It's like a whole new look for CP!

cellprofiler/gui/figure.py
```diff
@@ -1588,19 +1588,29 @@ def subplot_imshow_labels(
...
+            )
+            colormap = None
+        else:
+            # Mask the original labels
+            label_image = numpy.ma.masked_where(image == 0, image)
+            # Create a custom colormap
+            # Based off of the prism color map (with higher frequency):
+            # https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/_cm.py#L53-L57
+            new_prism = {
+                'red': lambda x: 0.75 * numpy.sin((x * 42 + 0.25) * numpy.pi) + 0.67,
```
0x00b1: A minor detail, but `0.67`?

AetherUnbound: I literally copied this block straight from matplotlib's source (linked above) and adjusted the frequency `20.9 --> 42` to make it a bit higher. What would you suggest changing this to?

AetherUnbound:
@AnneCarpenter I'm actually modifying the default _label_ colormap. This will only affect modules that try to show the segmentations in the display via the function `subplot_imshow_label`. All other functions are unaffected by this change.

For me (and @derekthirstrup) it seemed odd and frustrating that what should have been displayed as segmentation labels was displayed as RGB values. `subplot_imshow_label` calls `subplot_imshow` under the hood, which _does_ use the user-defined colormap if one was not defined. So all modules that don't use `subplot_imshow_label` will continue to use the user defined colormaps, but those that do will now have this overridden colormap. That said, even though we might be overriding the user's colormap, I think it makes sense in this case to do so because I'm doing some other logic to make sure the image background is black.

AetherUnbound:
@bethac07 @AnneCarpenter @0x00b1 
I believe I've fixed this appropriately! Several improvements:

- User defined colormaps work correctly for objects
- Object plots now show intensity correctly, rather than random RGB values, or floats for an integer intensity value
- Object labels are randomized similar to how they were before (so adjacently labeled objects won't usually be the same color), without losing the object label information

**Before**
2D with default colormap
2dbeforecmap
2D with "cool" colormap
2dbeforecmap_cool
3D with default colormap
3dbeforecmap

**After**
2D with default colormap
2daftercmap
2D with "cool" colormap
2daftercmap_cool
3D with default colormap
3daftercmap
3D with "cool" colormap
3daftercmap_cool

cellprofiler/gui/figure.py
```diff
@@ -1593,19 +1604,35 @@ def subplot_imshow_labels(
...
+            label_image = cellprofiler.object.overlay_labels(
+                labels=image,
+                opacity=opacity,
+                pixel_data=background_image,
+                max_label=max_label,
+                seed=seed
+            )
+            colormap = None
+        else:
+            # Mask the original labels
```
0x00b1: We could consider breaking this into a function to emphasize the underlying logic.

cellprofiler/modules/convertimagetoobjects.py
```diff
@@ -138,7 +138,7 @@ def display(self, workspace, figure):
             y=0
         )
 
-        figure.subplot_imshow(
+        figure.subplot_imshow_labels(
```
0x00b1: We should (eventually) consider converging to one common name for this type of data. In addition to labels, we also use masks and segmented. 🤷‍♀️

PR #3533, opened 2018-03-28
Release v3.1.3

AetherUnbound:


CONTRIBUTING.md
```diff
@@ -166,3 +166,18 @@ $ python setup.py test
 ```
 
 You can run individual tests by providing the option `-a PATH_TO_TEST_FILE`.
+
```
bethac07: We decided we wouldn't do this here, yes, but in the distribution repo?

cellprofiler/__init__.py
```diff
@@ -1,3 +1,3 @@
 __test__ = False
 
-__version__ = "3.0.0"
+__version__ = "3.1.3"
```
0x00b1: 😂

cellprofiler/modules/flagimage.py
```diff
@@ -744,7 +744,7 @@ def upgrade_settings(self, setting_values, variable_revision_number,
                 wants_maximum = cps.YES
             if new_or_append == "Append existing flag":
                 logger.warning(
-                        "CellProfiler 3.0 can't combine flags from multiple FlagImageForQC modules imported from version 1.0 and 2.0")
+                        "CellProfiler 3 can't combine flags from multiple FlagImageForQC modules imported from version 1.0 and 2.0")
```
0x00b1: Good idea!

CONTRIBUTING.md
AetherUnbound: I wasn't clear on where we landed, @0x00b1 ?

0x00b1: I don't have a preference.

0x00b1: Maybe we should leave this in CONTRIBUTING and add it to distribution too?

PR #3536, opened 2018-03-29
JSON pipelines (WIP)

0x00b1:


cellprofiler/pipeline.py
```diff
@@ -1117,6 +1117,82 @@ def encode_txt(self, s):
...
+        for module in self.__modules:
+            module_dictionary = {
+                "name": module.module_name,
+                "settings": {},
+                "batch_state": str(getattr(module, "batch_state")),
+                "enabled": getattr(module, "enabled"),
+                "module_num": getattr(module, "module_num"),
+                "notes": getattr(module, "notes"),
+                "show_window": getattr(module, "show_window"),
+                "svn_version": getattr(module, "svn_version"),
```
0x00b1: We can remove this since it’s no longer used.

cellprofiler/pipeline.py
```diff
@@ -1117,6 +1117,82 @@ def encode_txt(self, s):
...
+
+        for module in self.__modules:
+            module_dictionary = {
+                "name": module.module_name,
+                "settings": {},
+                "batch_state": str(getattr(module, "batch_state")),
+                "enabled": getattr(module, "enabled"),
+                "module_num": getattr(module, "module_num"),
+                "notes": getattr(module, "notes"),
+                "show_window": getattr(module, "show_window"),
```
0x00b1: We should rename this to `display`.

cellprofiler/pipeline.py
```diff
@@ -1117,6 +1117,87 @@ def encode_txt(self, s):
...
+    def to_dictionary(self):
+        pipeline_dictionary = {
+            "cookie": COOKIE,
+            "modules": [],
+            "version": cellprofiler.__version__
+        }
+
+        for module in self.__modules:
+            module_dictionary = {
+                "batch_state": str(getattr(module, "batch_state")),
```
AetherUnbound: These should probably have defaults, otherwise they'd throw `AttributeErrors`

cellprofiler/pipeline.py
```diff
@@ -1117,6 +1117,87 @@ def encode_txt(self, s):
...
+                "settings": {},
+                "version": getattr(module, "variable_revision_number")
+            }
+
+            for setting in module.settings():
+                # FIXME: Why does CellProfiler have hidden settings?
+                if not setting.text:
+                    continue
+
+                v = self.encode_txt(setting.unicode_value.encode('utf-8'))
```
AetherUnbound: The thing about encoding the values as utf-8 means that the actual pipeline file isn't really human readable. I was thinking - wouldn't it be possible to save the file as utf-8 and just keep these as unicode values?

cellprofiler/pipeline.py
```diff
@@ -1117,6 +1117,87 @@ def encode_txt(self, s):
...
+
+        return pipeline_dictionary
+
+    def export_json(self, pathname):
+        """
+
+        """
+        dictionary = self.to_dictionary()
+
+        with open(pathname, "w") as stream:
```
AetherUnbound: RE above, this can be a `codecs.open` with `utf-8`, then we don't have to encode the variables and they can still be readable!

Issue #3544, opened 2018-04-05
SaveCroppedObjects should change the file type to save disk space


karhohs:
Currently SaveCroppedObjects saves each object mask as a 8-bit TIFF. The storage overhead to create cropped objects can be orders of magnitude greater than the original size (approximately the number of objects in an image X the size of the image) of the image set being processed. I have run out of disk space several times due to this demand. Preferably, the object masks would be saved as a binary image.

bethac07:
A single binary image? You can already do this by doing "ConvertObjectsToImage" and "SaveImages"; I'm not sure we want to fully duplicate that functionality within SaveCroppedObjects.

karhohs:
Hi @bethac07! I think the behavior with the TIFF files is specific to the SaveCroppedObjects module. An example of the issue is if I wanted to create masks and my image set was 100MB, I'd end up with 10GB worth of masks.

bethac07:
Resolved by #3545

PR #3545, opened 2018-04-05
PNG compression solves the storage issue

karhohs:
Re: #3544 It seems like saving masks as 8-bit PNG instead of 8-bit TIFF reduces the space required a satisfactory amount; it is a one-line fix. FWIW, I did not see a way to save a binary image with skimage, but did find an issue that alludes to this https://github.com/scikit-image/scikit-image/issues/1623.

I've been converting the TIFF images into PNG images after CellProfiler creates the crops/masks, so this would also save a step from a practical perspective.

### DIFF
--- a/cellprofiler/modules/savecroppedobjects.py
+++ b/cellprofiler/modules/savecroppedobjects.py
@@ -78,8 +78,8 @@ class SaveCroppedObjects(cellprofiler.module.Module):
 
             filename = os.path.join(
                 directory,
-                "{}_{:04d}_{}.png".format(self.objects_name.value, label, int(time.time()))
-            )
+                "{}_{}.png".format(self.objects_name.value, label)
+                )
 
             skimage.io.imsave(filename, skimage.img_as_ubyte(mask))
 
--- a/cellprofiler/modules/savecroppedobjects.py
+++ b/cellprofiler/modules/savecroppedobjects.py
@@ -78,7 +78,7 @@ class SaveCroppedObjects(cellprofiler.module.Module):
 
             filename = os.path.join(
                 directory,
-                "{}_{:04d}_{}.tiff".format(self.objects_name.value, label, int(time.time()))
+                "{}_{:04d}_{}.png".format(self.objects_name.value, label, int(time.time()))
             )
 
             skimage.io.imsave(filename, skimage.img_as_ubyte(mask))
--- a/cellprofiler/modules/savecroppedobjects.py
+++ b/cellprofiler/modules/savecroppedobjects.py
@@ -8,10 +8,9 @@ SaveCroppedObjects
 the value 255. All other pixels (i.e., background pixels and pixels corresponding to other objects) are assigned the
 value 0. The dimensions of each image are the same as the original image.
 
-The filename for an exported image is formatted as "{object name}_{label index}_{timestamp}.tiff", where *object name*
+The filename for an exported image is formatted as "{object name}_{label index}.tiff", where *object name*
 is the name of the exported objects, *label index* is the integer label of the object exported in the image (starting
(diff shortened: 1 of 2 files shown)


karhohs:
Looks like I need to update the test file. Would someone be willing to offer some guidance here?

bethac07:
1) https://travis-ci.org/CellProfiler/CellProfiler/builds/362779313#L2639  Change the expected file name for the associated test file

bethac07:
Nevermind, the last one is simple- you can't save 3D volumetric images as PNGs

karhohs:
In the end, it looks like 8-bit files are the way to go in terms of compatibility with TIFF and PNG. Offering TIFF will work for both 2D and 3D masks, and PNG is available to make the module more general. Compressing the TIFF is very similar to PNG compression in terms of storage savings.

cellprofiler/modules/savecroppedobjects.py
```diff
@@ -51,6 +50,17 @@ def create_settings(self):
             value=cellprofiler.setting.DEFAULT_OUTPUT_FOLDER_NAME
         )
 
+        self.file_format = cellprofiler.setting.Choice(
+            "Saved file format",
+            [
+                "png",
```
AetherUnbound: Usually these kinds of string constants for options are put at the top of the module
E.g. https://github.com/CellProfiler/CellProfiler/blob/master/cellprofiler/modules/labelimages.py#L53
This makes it so that if the strings ever change, they only have to be changed in one place

cellprofiler/modules/savecroppedobjects.py
```diff
@@ -76,12 +86,21 @@ def run(self, workspace):
         for label in unique_labels:
             mask = labels == label
 
-            filename = os.path.join(
-                directory,
-                "{}_{:04d}_{}.tiff".format(self.objects_name.value, label, int(time.time()))
-            )
+            if self.file_format.value == "png":
```
AetherUnbound: This strings would also be changed to reference the aforementioned constant

cellprofiler/modules/savecroppedobjects.py
```diff
@@ -76,12 +86,21 @@ def run(self, workspace):
...
+                skimage.io.imsave(filename, skimage.img_as_ubyte(mask))
+
+            elif self.file_format.value == "tiff":
+                filename = os.path.join(
+                    directory,
+                    "{}_{}.tiff".format(self.objects_name.value, label)
+                    )
 
-            skimage.io.imsave(filename, skimage.img_as_ubyte(mask))
+                skimage.io.imsave(filename, skimage.img_as_ubyte(mask), compress=6)
```
AetherUnbound: Would it make sense to allow users to specify this compression level if the `tiff` option was selected?

0x00b1: @karhohs asked me about that and I said it was probably worth avoiding since, in my experience, TIFF libraries have varying levels of support for compression. @AetherUnbound what do you think?

AetherUnbound: That's fair. Also, I can't imagine a time myself where I'd want _less_ compression, if anything I'd want more. 6 is good :)

bethac07:
I think #3549 has a valid point, before we pull this do we want to address it?

cellprofiler/modules/savecroppedobjects.py
karhohs: 6 was more or less an arbitrary choice. I tried a few different levels, but the results were all similar. I this is likely due to the binary and sparse nature of the masks. A little compression goes a long ways.

karhohs:
commit *aee2743* implements the suggestions made by @AetherUnbound

PR #3550, opened 2018-04-17
Add ability to add scatterplots to 3D module displays

AetherUnbound:
Previously, scatterplots couldn't be added to modules that could be used in 3D pipelines because said pipelines use "Gridspec" to display data, not regular subplots (as `subplot_scatter` was expecting). This PR extends the `subplot_scatter` function to allow for displaying a single scatterplot within a 3D module's display. 

Example:
screenshot_2018-04-17_09-47-39

### DIFF
--- a/cellprofiler/gui/figure.py
+++ b/cellprofiler/gui/figure.py
@@ -696,11 +696,16 @@ class Figure(wx.Frame):
             else:  # self.dimensions == 3
                 axes = event.inaxes
 
-                axes_image = axes.get_images()[0]
+                axes_image_list = axes.get_images()
 
-                fields += ["Z: {}".format(axes.get_label())]
+                if len(axes_image_list):
+                    axes_image = axes_image_list[0]
 
-                im = axes_image.get_array().data
+                    fields += ["Z: {}".format(axes.get_label())]
+
+                    im = axes_image.get_array().data
+                else:
+                    im = None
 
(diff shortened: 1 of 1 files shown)


cellprofiler/gui/figure.py
```diff
@@ -1918,6 +1935,9 @@ def subplot_scatter(self, x, y,
         axes.set_xscale(xscale)
         axes.set_yscale(yscale)
 
+        if self.dimensions == 3:
```
0x00b1: We should probably just do this in the future and avoid `matplotlib.gridspec` altogether!

0x00b1:
@AetherUnbound LGTM but Travis is failing.

0x00b1:
And I have no idea how that failure is related to this PR. 🤔

cellprofiler/gui/figure.py
AetherUnbound: Wait the `gridspec` is pretty neat haha, I'd say we move everything over to it!

AetherUnbound:
I'm not sure how this error is related either...investigating

AetherUnbound:
This has to do with a pytest version mismatch (no idea why this was ever working??). This works on pytest v3.3.2 but not v3.0.5

AetherUnbound:
Passing unit tests! @0x00b1

PR #3566, opened 2018-05-11
Fix caching section for Travis CI

iriberri:
It looks like the `cache` key was missing in the YAML and this caused some parse issues in Travis CI - hope this helps!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ addons:
       - libhdf5-serial-dev
       - python-mysqldb
       - python-pip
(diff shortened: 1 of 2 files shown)


iriberri:
Oops - just noticed Travis didn't ran a build because I had my own account blocked. Closing and reopening to trigger it again :)

AetherUnbound:
It looks like this build is failing on the most recent version of numpy. I'll work on this, it may just mean pinning the numpy version for now.

Issue #3571, opened 2018-05-15
Add bounding box (and potentially other regionprops) to MeasureObjectSizeShape


karhohs:
Note, this is a 2D request.

It would be convenient to have bounding box coordinates included in the measurements and features created by the MeasureObjectSizeShape module. In a particular use case, bounding boxes output from a deep learning model are being compared to a CellProfiler segmentation pipeline; it would be helpful if CellProfiler produced bounding box coordinates to make this comparison straightforward.

Fortunately, skimage.measure.regionprops already includes bounding box information, so it should be possible to include this in CellProfiler output, too.

Looking at measureobjectsizeshape.py the 2D measurements appear to be made after an if else statement: https://github.com/CellProfiler/CellProfiler/blob/cd06ed3706485bccf9ca28901c6ad304523f79bc/cellprofiler/modules/measureobjectsizeshape.py#L344

Currently, the following measurements are made:
* Area
* Center X
* Center Y
* Center Z
* Compactness
* Eccentricity
* Extent
* Perimeter
* Solidity
* Form Factor
* Euler Number
* Major Axis Length
* Maximum Radius
* Mean Radius
* Median Radius
* Min Feret Diameter
* Minor Axis Length
* Max Feret Diameter
* Orientation
* Zernike Numbers

There are a lot measurements created by skimage region props:
* Area
* Bbox, the bounding box (min_row, min_col, max_row, max_col)
* Bbox Area
* Centroid
* Convex Area
* Convex Image
* Coordinate List
* Eccentricity
* Equivalent Diameter
* Euler Number
* Extent
* Filled Area
* Filled Image
* Image binary thumbnail the same size as the bounding box
* inertia tensor
* inertia tensor eigenvalues
* intensity image
* label
* local_centroid
* major axis length
* max intensity
* mean intensity
* min intensity
* minor axis length
* moments (central, hu, normalized)
* orientation
* perimeter
* solidity
* weighted moments based on intesnity

If we add bounding box coordinates, are there any other regionprops worth adding? Many of the skimage regionprops would not fit into the category of SizeShape (or into a table) and many already exist in CellProfiler. Are the more exotic measurements, such as the intertia tensor eigenvalues, of any use?

PR #3573, opened 2018-05-22
Bounding box coordinates 

jhung0:


karhohs:
This would solve #3571

cellprofiler/modules/measureobjectsizeshape.py
```diff
@@ -112,6 +112,8 @@
    angles). The minimum and maximum Feret diameters are the smallest and
    largest possible diameters, rotating the calipers along all possible
    angles.
+-  *BoundingBoxMinX, BoundingBoxMinY, BoundingBoxMaxX, BoundingBoxMaxY:* *(2D only)* The
+   bounding box coordinates for an object. X corresponds to rows and Y corresponds to columns.
```
bethac07: "X corresponds to rows and Y corresponds to columns."- this is vague.  Are you saying it's opposite to what we'd expect (ie X corresponds to position up-and-down and Y corresponds to position left-and-right)?  If so, we should change how the variables are assigned to correspond to the "expected" definitions of X and Y.  If not, we should clarify or omit this sentence.

0x00b1:
Bump @jhung0

cellprofiler/modules/measureobjectsizeshape.py
bethac07: Confirmed that the coordinates are currently backwards relative to meanings of X and Y elsewhere in CP

cellprofiler/modules/measureobjectsizeshape.py
```diff
@@ -367,6 +373,19 @@ def run_on_objects(self, object_name, workspace):
...
                 nobjects = numpy.max(objects.indices)
+
+            bounding_box_min_x = []
+            bounding_box_min_y = []
+            bounding_box_max_x = []
+            bounding_box_max_y = []
+
+            for label, index in objects.get_labels():
+                props = skimage.measure.regionprops(label)
+                bounding_box_min_x += [int(prop.bbox[0]) for prop in props]
```
bethac07: I think all you need to do here is set bounding_box_min_x to bbox[1], min_y to bbox[0], max_x to bbox[3], and max_y to bbox[2]

Issue #3531, opened 2018-03-27
Resize fails when image isn't between -1 and 1


bethac07:
In 3.0.0 and 3.1.1, if you try to do a resize module on an image that's not between -1 and 1, you get the error below.  This happens with all interpolation methods and resizing factors I tried (both up and down).  This is crucial for moving the Cell Painting illumination correction pipeline (or really any illumination correction pipeline done on large images) to 3.X.

```
Failed to run module Resize
Traceback (most recent call last):
  File "CellProfiler\cellprofiler\gui\pipelinecontroller.py", line 2887, in do_step
  File "CellProfiler\cellprofiler\pipeline.py", line 2022, in run_module
  File "CellProfiler\cellprofiler\modules\resize.py", line 258, in run
  File "CellProfiler\cellprofiler\modules\resize.py", line 336, in apply_resize
  File "site-packages\skimage\transform\_warps.py", line 135, in resize
  File "site-packages\skimage\transform\_warps.py", line 722, in warp
  File "site-packages\skimage\transform\_warps.py", line 535, in _convert_warp_input
  File "site-packages\skimage\util\dtype.py", line 336, in img_as_float
  File "site-packages\skimage\util\dtype.py", line 236, in convert
ValueError: Images of type float must be between -1 and 1.
```

Files and tip via [the forum](http://forum.cellprofiler.org/t/problem-with-the-illumination/5850)
[ResizeBug.zip](https://github.com/CellProfiler/CellProfiler/files/1854112/ResizeBug.zip)

AetherUnbound:
I feel like this may be an issue with the illumination correction module more than it is a module with resize, because resize is expecting float values between -1 and 1 (which I think CellProfiler "should" keep all images in that format internally) but the illumination pipeline provided creates an image with `1` and `2` as values. I'm still investigating this though.

bethac07:
Yeah, someone just posted a bug in another module in the forum the other day re: other modules being cranky with values outside of the -1>x>1 range.  The cases where you have values outside that range are rare (such as illumination correction), but important and CP should be able to handle them.

votti:
Hi there,
[...]

PR #3577, opened 2018-06-03
Issues/3531

votti:
This fixes the issue #3531 
It is addressed by rescaling the intensities to -1/1 during the function call to skimage.transform.resize, by dividing by the maximum absolute value of the image. The result is scaled back to the original values by multiplying by the same value.

A test has been added to test for this potential issue.

### DIFF
--- a/cellprofiler/modules/resize.py
+++ b/cellprofiler/modules/resize.py
@@ -317,20 +317,24 @@ resized with the same settings as the first image."""))
         new_shape = self.resized_shape(image, workspace)
 
         order = self.spline_order()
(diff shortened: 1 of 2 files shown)


votti:
"Ready to pull- the failing tests are all due to skimage 0.14 (see #3580)"
Does also apply here. As far as I can judge the failing tests are not related to any code I touched.

bethac07:
Closing temporarily to retrigger Travis

0x00b1:
Thanks, @votti! This is fantastic. :)

PR #3601, opened 2018-07-07
Allow an arbitrary number of channels to be added in ColorToGray

### LINKED ISSUE #3598: Support more channels in `colortogray`
Our images can be stacks of more than 20 channels, but the current channel selection settings in the`colortogray` module only supports up to 20 channels (hardcoded: https://github.com/CellProfiler/CellProfiler/blob/613247de8b9227b17f9b2eb31b43ad86682aee7d/cellprofiler/modules/colortogray.py#L210).

Thus I would suggest two options:
A) Either increase the hard coded max channel number to e.g. 100 channels
B) Change the widget from `cps.Choice` to `cps.Integer`

I am happy to write a pull request for either option.

AetherUnbound:
Resolves #3598 

I've modified ColorToGray to allow an arbitrary number of channels to be added to the module. I might try and add a test to split more than 20 images as was the cased described in #3598. 

@bethac07 and @0x00b1, I'm worried that the way I've changed the module may work going forward and pass all the tests, but won't work for pipelines that involved more than the 4 originally defined channels. Do you guys have a pipeline with more than 4 channels that you could test to see if this still works?


AetherUnbound:
I have absolutely no clue why the build is failing :(

PR #3619, opened 2018-08-06
Issue/3598

votti:
We are dealing routinely with 30+ channel images. As discussed in the issue #3598 , it thus would sometimes be convenient to easily extract single planes to grayscale using `colortogray`. The previous solution for this solution was elegant and supported that many channels. Unfortunately in practice it is very tedious, as e.g. selecting a channel with index 20 would require the addition of 20 channels.

I thus propose to change the channel_choice widget to a `cps.Integer`. This is a little bit less verbose, the user will put in just a channel number. The hint that Channel1 often is 'Red' is thus a bit less obvious as before, when the user selected 'Red: 1'. However I would argue a user that needs to split using the 'Channel' and not the 'RGB' option, anyway often should be aware what channel number corresponds to what color. Additionally the hint that channel 1 is red would still be available in the help. Thus I find the solution of using an Integer to indicate a channel number a veritable alternative, that is more convenient if higher plane numbers should be selected.

### DIFF
--- a/cellprofiler/modules/colortogray.py
+++ b/cellprofiler/modules/colortogray.py
@@ -528,14 +528,14 @@ Select the name of the output grayscale image."""))
                               setting_values[13:])
             variable_revision_number = 3
 
-        #
-        # Standardize the channel choices
-        #
-        setting_values = list(setting_values)
-        nchannels = int(setting_values[SLOT_CHANNEL_COUNT])
-        for i in range(nchannels):
-            idx = SLOT_FIXED_COUNT + SLOT_CHANNEL_CHOICE + i * SLOTS_PER_CHANNEL
-            channel_idx = self.get_channel_idx_from_choice(setting_values[idx])
-            setting_values[idx] = channel_idx+1
-
+        if variable_revision_number < 4:
+            #
+            # Standardize the channel choices
+            #
+            setting_values = list(setting_values)
+            nchannels = int(setting_values[SLOT_CHANNEL_COUNT])
+            for i in range(nchannels):
+                idx = SLOT_FIXED_COUNT + SLOT_CHANNEL_CHOICE + i * SLOTS_PER_CHANNEL
+                channel_idx = self.get_channel_idx_from_choice(setting_values[idx])
+                setting_values[idx] = channel_idx+1
         return setting_values, variable_revision_number, from_matlab
--- a/cellprofiler/modules/colortogray.py
+++ b/cellprofiler/modules/colortogray.py
@@ -219,11 +219,11 @@ Enter a name for the resulting grayscale image coming from the value.""")
             doc="""\
 *(Used only when splitting images)*
 
-This setting chooses a channel to be processed. For example, *Red: 1*
+This setting chooses a channel to be processed. For example, *1*
 is the first
 channel in a .TIF or the red channel in a traditional image file.
-*Green: 2* and *Blue: 3* are the second and third channels of a TIF or
-the green and blue channels in other formats. *Alpha: 4* is the
+*2* and *3* are the second and third channels of a TIF or
+the green and blue channels in other formats. *4* is the
 transparency channel for image formats that support transparency and is
 channel # 4 for a .TIF file. **ColorToGray** will fail to process an
 image if you select a channel that is not supported by that image, for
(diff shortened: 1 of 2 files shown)


votti:
I noticed that I mixed up `channel_name` and `channel_choice` in my commit messages which might have made them confusing - I did a rebase to reword them.

cellprofiler/modules/colortogray.py
```diff
@@ -208,30 +208,22 @@ def create_settings(self):
...
         group.can_remove = can_remove
-        if len(self.channels) == len(self.channel_names):
-            ch = next(self._get_next_channel())
-            self.channel_names.append(ch)
-        group.append("channel_choice", cps.Choice(
+        group.append("channel_choice", cps.Integer(
             text="Channel number",
-            choices=self.channel_names,
-            value=self.channel_names[len(self.channels)],
+            value=len(self.channels)+1,
```
0x00b1: Add a space between operators, e.g.

```Python
value=len(self.channels) + 1
```

0x00b1:
@votti This is very close. I believe you need to update the `upgrade_settings` method. 

Let me know if you need help!

cellprofiler/modules/colortogray.py
votti: Thanks for the hint! I will check the formatting once more.

cellprofiler/modules/colortogray.py
```diff
@@ -537,6 +536,6 @@ def upgrade_settings(self,
         for i in range(nchannels):
             idx = SLOT_FIXED_COUNT + SLOT_CHANNEL_CHOICE + i * SLOTS_PER_CHANNEL
             channel_idx = self.get_channel_idx_from_choice(setting_values[idx])
-            setting_values[idx] = self.channel_names[channel_idx]
+            setting_values[idx] = channel_idx+1
```
votti: I adapted the `upgrade_settings` by changing this line as well as the `get_channel_idx_from_choice` function.
This seemed to work fine for the example settings in the `test_colortogray` and some manual examples.

I noticed that the `upgrade_settings` for this module was a bit special, as it runs the `# Standardize the channel choices` ignoring the actual `variable_revision_number` - is there actually a particular reason for that? Or should I better also put this section in an `if` clause dependent on the  `variable_revision_number`?

AetherUnbound: The purpose of the `upgrade_settings` is for allowing previous versions of `.cppipe` files to still be loaded into new version of cellprofiler. I'm not sure that this block is needed anymore at all, but I would test two things:
1) Export a `.cppipe` file with a bunch of channels in the master branch, then check to make sure that the pipeline loads properly in your modified branch
2) Perform the same steps as above, but remove that whole standardization block entirely! See what it changes :) It may not be needed at all

Thanks again for your work here!

votti:
Hi team,
I addressed the issues you raised:
- corrected the formatting
- added a test for old pipelines
- changed the  `upgrade_settings` to do nothing if the settings come from the latest version.

Also I noticed that I did not bump the version number yet.
Please let me know if there is anything missing!

cellprofiler/modules/colortogray.py
```diff
@@ -528,14 +528,14 @@ def upgrade_settings(self,
...
-        # Standardize the channel choices
-        #
-        setting_values = list(setting_values)
-        nchannels = int(setting_values[SLOT_CHANNEL_COUNT])
-        for i in range(nchannels):
-            idx = SLOT_FIXED_COUNT + SLOT_CHANNEL_CHOICE + i * SLOTS_PER_CHANNEL
-            channel_idx = self.get_channel_idx_from_choice(setting_values[idx])
-            setting_values[idx] = channel_idx+1
-
+        if variable_revision_number < 4:
```
AetherUnbound: Great! I think this is exactly what was needed. I believe you'll also need to set `variable_revision_number` to 4 at the end of this block now (see line 529)

tests/modules/test_colortogray.py
```diff
@@ -319,3 +319,84 @@ def callback(caller, event):
     assert module.channels[0].image_name == "RedChannel1"
     assert module.channels[1].image_name == "GreenChannel2"
     assert module.channels[2].image_name == "BlueChannel3"
+
+def test_2_6_load_v3():
```
AetherUnbound: Thank you for adding this test!

AetherUnbound:
@votti This PR is great! Thanks for your work on this :) I think there's only 1 more change that needs to be made, and then it's ready. As soon as that change is made I'll merge it in :+1:

0x00b1:
LGTM!

PR #3639, opened 2018-09-09
Added "Open Folder" option to mouse pointer context menu in file drop zone

VolkerH:
When running Cellprofiler on Linux (particularly remotely through X forwarding) one may not have a file browser supporting dragging and dropping files/folder into the image import area widget. The existing context menu only offers the option to add files individually, which is impractical if there are many images in a deep folder structure.

This pull request adds an "Open Folder" context menu item that enables selection of an individual folder, making it possible to add folders when drag and drop is not possible.


VolkerH:
context_menu

cellprofiler/gui/pipelinecontroller.py
```diff
@@ -133,6 +133,8 @@ def __init__(self, workspace, frame):
         wx.EVT_MENU(frame, cpframe.ID_EDIT_DELETE, self.on_remove_module)
         wx.EVT_MENU(frame, cpframe.ID_EDIT_DUPLICATE, self.on_duplicate_module)
 
+        wx.EVT_MENU(frame, cpframe.ID_EDIT_BROWSE_FOR_FOLDER,
+                    self.on_pathlist_browse_folder)
         wx.EVT_MENU(frame, cpframe.ID_EDIT_BROWSE_FOR_FILES,
                     self.on_pathlist_browse)
```
AetherUnbound: I would actually modify the original `on_pathlist_browse` to be `on_pathlist_browse_images`, just to make a better distinction from `on_pathlist_browse_folder`.

cellprofiler/gui/pipelinecontroller.py
```diff
@@ -1537,6 +1551,7 @@ def get_pathlist_file_context_menu(self, paths):
                 (self.PATHLIST_CMD_REMOVE, self.PATHLIST_CMD_REMOVE),
                 (self.PATHLIST_CMD_REFRESH, self.PATHLIST_TEXT_REFRESH),
                 (self.PATHLIST_CMD_BROWSE, self.PATHLIST_CMD_BROWSE),
```
AetherUnbound: Same here for the `PATHLIST_CMD_BROWSE` command

cellprofiler/gui/pipelinecontroller.py
```diff
@@ -1560,6 +1575,15 @@ def on_pathlist_file_command(self, paths, cmd):
                 self.on_pathlist_browse(
                     None,
                     default_dir=path)
+        elif cmd == self.PATHLIST_CMD_BROWSE_FOLDER:
```
AetherUnbound: I think in this case and the case below, it might be better to have the following:
```
        elif cmd in [self.PATHLIST_CMD_BROWSE, self.PATHLIST_CMD_BROWSE_FOLDER]:
            browse_function = self.on_pathlist_browse_folder if cmd == self.PATHLIST_CMD_BROWSE_FOLDER else self.on_pathlist_browse
            if len(paths) == 0 or not paths[0].startswith("file:"):
                browse_function(None)
            else:
                path = urllib.url2pathname(paths[0][5:])
                path = os.path.split(path)[0]
                browse_function(None, default_dir=path)
```
That way we only have the `urllib` parsing occurring once and don't have repeated code.

VolkerH:
Ok, I made the suggested changes. (with the slight difference that I wrote the conditional assignment to `browse_function` as an explicit `if: else:` statement to break the long line.

AetherUnbound:
Hmm, Travis is being weird about the build. We'll give it a little bit but otherwise I'll merge it in :smiley:

VolkerH:
Thanks for the quick merge!

Issue #2059, opened 2016-06-02
Remove `done-file` command-line option


0x00B1:
This can be replaced by returning correct exit codes.

bethac07:
When this gets integrated if we're still using the Distributed-CellProfiler we're/I'm going to need you to have a look at that code, since it relies on checking for the existence and contents of the done file.  Just a forewarning :)

0x00B1:
@bethac07 You can make the change since CellProfiler now returns the appropriate code.

bethac07:
@0x00B1 
1)On the stable?
2)I don't know how and therefore will still probably need someone to code review it once I try to figure it out to make sure I didn't break anything

PR #3640, opened 2018-09-12
Bubble exit code up from headless run for proper system reporting

AetherUnbound:
The error code that was returned from run_pipeline_headless was not being captured by the calling process. Additionally, an exception raised in a pipeline wasn't being caught as a failure inside the run_pipeline_headless block.

This was preventing us from capturing the exit code appropriately when cellprofiler failed on a run, since it would just emit a `0`.

### DIFF
--- a/cellprofiler/__main__.py
+++ b/cellprofiler/__main__.py
@@ -57,7 +57,7 @@ def main(args=None):
         cellprofiler.preferences.set_headless()
         cellprofiler.worker.aw_parse_args()
         cellprofiler.worker.main()
-        sys.exit(exit_code)
+        return exit_code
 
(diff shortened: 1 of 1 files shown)


AetherUnbound:
This is the first step for #2059 it looks like

cellprofiler/__main__.py
```diff
@@ -130,6 +130,7 @@ def main(args=None):
     if options.write_schema_and_exit:
         write_schema(options.pipeline_filename)
 
+    exit_code = 0
```
bengerman13: AFAICT - this is unneeded. We set `exit_code = 0` on line 51

PR #3641, opened 2018-09-14
Demote the update exception from warning to debug

AetherUnbound:
This message is often part of peoples' error messages on
GitHub/the forum, and it's a bit misleading. Since this site
is no longer maintained, we shouldn't let users know about this
error anyway.


bengerman13:
Should the script even try to do anything if the site is not maintained anymore?

AetherUnbound:
That's a fair point, it might be worth removing this code entirely.

bengerman13:
Does the script still do anything useful _ever_? Should it be removed, or turned into a big no-op that emits a deprecation warning?

AetherUnbound:
While I want to agree, I want to touch the stable branch as little as possible.

PR #3643, opened 2018-09-14
WIP: DO NOT MERGE -- flake8 across ALL repos

cclauss:
@AetherUnbound The flake8 tests can help in the port to Python 3 but it is useful to make sure they run cleanly on Python 2 as well as Python 3.  This script runs common tests across all __CellProfiler__ Python repos.

<sub>https://travis-ci.org/CellProfiler/CellProfiler/builds/428840598?utm_source=github_status&utm_medium=notification</sub>

__E901,E999,F821,F822,F823__ are the "_showstopper_" flake8 issues that can halt the runtime with a SyntaxError, NameError, etc. Most other flake8 issues are merely "style violations" -- useful for readability but they do not effect runtime safety.
* F821: undefined name `name`
* F822: undefined name `name` in `__all__`
* F823: local variable name referenced before assignment
* E901: SyntaxError or IndentationError
* E999: SyntaxError -- failed to compile a file into an Abstract Syntax Tree

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,42 @@
-addons:
-  apt:
-    packages:
-      - libhdf5-serial-dev
-      - python-mysqldb
-      - python-pip
-cache:
-  apt: true
-  directories: $HOME/.cache/pip
-dist: trusty
-env:
-  global:
-    - LC_ALL="en_US.UTF-8" CP_MYSQL_TEST_HOST="127.0.0.1" CP_MYSQL_TEST_USER="root" CP_MYSQL_TEST_PASSWORD=""
-  matrix:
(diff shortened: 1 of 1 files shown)


AetherUnbound:
@cclauss this is really slick, thanks for taking the time to go through this!

cclauss:
If you want we could set up another _temporary_ repo called __CellProfiler-flake8-tests__ (or similar) with the same Travis CI file used above.  For each repo, you could get these five flake8 tests to be green on Python 2 and then I could follow up with getting the flake8 tests to be green on Python 3 as well.  Being syntax compatible with both Python 2 and Python 3 is NOT a complete port but it is a solid step in the right direction.

PR #3652, opened 2018-09-30
Python 3 fixes for basestring, cmp(), reload(), unicode(), xrange()

cclauss:
Adds the popular [__six__](http://six.readthedocs.io) module for compatibility between Python 2 and Python 3.

I could uses some guidance on how to handle the conversion of [__unicode(x, "utf-8")__.](https://travis-ci.org/CellProfiler/CellProfiler/jobs/435299908#L969)..

For a later PR, we should look at __six.StringIO__ vs __six.moves.cStringIO__.

### DIFF
--- a/cellprofiler/gui/metadatactrl.py
+++ b/cellprofiler/gui/metadatactrl.py
@@ -7,7 +7,7 @@ import cellprofiler.preferences
 import wx
 import wx.lib.masked
 
-from six import text_type, unichr
+import six
 
 __choice_ids = []
 
@@ -128,7 +128,7 @@ class MetadataControl(wx.PyControl):
                 if value[index] in ('g', '?'):
                     state = STATE_PRE
                 else:
-                    self.__tokens.append(text_type(value[index]))
+                    self.__tokens.append(six.text_type(value[index]))
                     state = STATE_INITIAL
             elif state == STATE_PRE:
                 if value[index] != '<':
@@ -386,7 +386,7 @@ class MetadataControl(wx.PyControl):
 
     def on_char(self, event):
         self.delete_selection()
-        c = unichr(event.GetUnicodeKey())
+        c = six.unichr(event.GetUnicodeKey())
         self.__tokens.insert(self.__cursor_pos, c)
         self.move_cursor_pos(self.__cursor_pos + 1)
         self.on_token_change()
--- a/cellprofiler/gui/metadatactrl.py
+++ b/cellprofiler/gui/metadatactrl.py
@@ -7,6 +7,8 @@ import cellprofiler.preferences
 import wx
 import wx.lib.masked
 
+from six import text_type, unichr
+
 __choice_ids = []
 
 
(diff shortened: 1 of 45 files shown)


cellprofiler/gui/metadatactrl.py
```diff
@@ -7,6 +7,8 @@
 import wx
 import wx.lib.masked
 
+from six import text_type, unichr
```
AetherUnbound: Is `unichr` needed here?

cellprofiler/gui/pipelinecontroller.py
```diff
@@ -51,6 +51,13 @@
 import wx.lib.buttons
 import wx.lib.mixins.listctrl
 from functools import reduce
+from six import string_types, text_type
+
+try:
+    cmp             # Python 2
+except NameError:
+    def cmp(a, b):  # Python 3
```
AetherUnbound: This is defined wherever `cmp` is needed, I think it might be good to go into `cellprofiler.utilities` (I can move it there)

cclauss:
The only caution that I would have on fully qualifying is that it is slower.  There is an [__avoid the dots__](https://wiki.python.org/moin/PythonSpeed/PerformanceTips#Avoiding_dots...) performance tip that might be worth paying attention to on performance sensitive code that is executed repeatedly.

cclauss:
```python
import timeit
print(timeit.timeit("s = six.text_type('string')", "import six"))  # 0.17...
print(timeit.timeit("s = text_type('string')", "import six ; from six import text_type"))  # 0.12...
```

AetherUnbound:
Thanks for the input! Honestly, with a codebase this large, I prefer the readability of fully qualified imports over the speed gained with the `from` syntax.

AetherUnbound:
It looks like our `centresome` package is the next in line to get the python 3 treatment (since the build is failing on that now) - good thing there's been a lot of work there! I'll work on getting that merged in and deployed.

cclauss:
https://github.com/CellProfiler/centrosome/pull/86 might help too.

cclauss:
https://github.com/CellProfiler/centrosome/pull/86/files#diff-6bd152122a88772ca6beef760652315bR134

0x00b1:
I think we should replace instances of `cStringIO` with `StringIO`.

cclauss:
Six gives us [six.StringIO](https://six.readthedocs.io/#six.StringIO) and and [__six.moves.cStringIO__](https://six.readthedocs.io/#module-six.moves) so the high performance way to go is __from six.moves.cStringIO import StringIO__ while the slower,safer way to go is __from six import StringIO__.

AetherUnbound:
@cclauss I'd like to merge this, we can work on the next set of python3 related changes once I deploy a new version of centrosome in another branch. Anything else you think we need here?

cclauss:
Let's go for it.  Did you see this comment on version mismatch? https://github.com/CellProfiler/CellProfiler/commit/cc7573aca83bca227c96d6088fe5379c46fa157c#r30721946  Perhaps it is intentional.

AetherUnbound:
It was not - I was attempting to change the deploy mechanism and finally got it all working on 1.5.2 (even though 1.5.1 is effectively the same code). Thanks for noticing!

PR #3664, opened 2018-10-09
Set numpy to be higher than the travis default (1.13)

AetherUnbound:
I don't want to merge this until we have fixes for the failing `unpinned_dependency.txt` builds


0x00b1:
Thanks, @AetherUnbound!

PR #2831, opened 2017-08-12
Changed the CropObjects module adding the functionality to save the cropped object from the original image.

SariNusier:
Follow up on PR #2790 

As requested, kept the option to save masks while adding the option to save the objects as individual images cropped from the original image.

PR #3671, opened 2018-10-18
SaveCroppedObjects, redux

### LINKED ISSUE #2255: Create a new module: ExportImagesPerObject

bethac07:
Resolves #2255 

Taking the best of @SariNusier 's PR #2831  and putting it on top of the current expanded functionalities of SaveCroppedObjects.

### DIFF
--- a/cellprofiler/modules/savecroppedobjects.py
+++ b/cellprofiler/modules/savecroppedobjects.py
@@ -159,7 +159,8 @@ class SaveCroppedObjects(cellprofiler.module.Module):
         result = [
             self.export_option,
             self.objects_name,
-            self.directory
+            self.directory,
+            self.file_format
         ]
 
         if self.export_option.value == SAVE_PER_OBJECT:
--- a/cellprofiler/modules/savecroppedobjects.py
+++ b/cellprofiler/modules/savecroppedobjects.py
@@ -139,7 +139,7 @@ class SaveCroppedObjects(cellprofiler.module.Module):
 
                 skimage.io.imsave(filename, skimage.img_as_ubyte(mask), compress=6)
 
-        filenames.append(filename)
+            filenames.append(filename)
 
         if self.show_window:
             workspace.display_data.filenames = filenames
--- a/cellprofiler/modules/savecroppedobjects.py
+++ b/cellprofiler/modules/savecroppedobjects.py
@@ -25,6 +25,7 @@ YES          NO           YES
 import numpy
 import os.path
 import skimage.io
+import skimage.measure
 import time
 
 import cellprofiler.module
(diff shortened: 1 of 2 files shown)


cellprofiler/modules/savecroppedobjects.py
```diff
@@ -25,27 +25,54 @@
...
+            ],
+            doc="""
+            Choose the way you want the per-object crops to be exported.
+            <p>The choices are:<br>
+            <ul><li><i>{SAVE_PER_OBJECT}</i>: Save a per-object crop from the original image based on the object's
+            bounding box.</li>
+            <li><i>{SAVE_MASK}</i>: Export a per-object mask.</li>
+            </ul></p>
+            """.format(**{
+                "SAVE_PER_OBJECT": SAVE_PER_OBJECT,
```
AetherUnbound: I know this is how other modules do it, but for IDE ease do you mind replacing this with the following?
```
.format(
    SAVE_PER_OBJECT=SAVE_PER_OBJECT,
    SAVE_MASK=SAVE_MASK
)
```

cellprofiler/modules/savecroppedobjects.py
```diff
@@ -86,7 +113,15 @@ def run(self, workspace):
...
         for label in unique_labels:
-            mask = labels == label
+            if self.export_option == SAVE_MASK:
+                mask = labels == label
+
+            elif self.export_option == SAVE_PER_OBJECT:
+                mask_in = labels == label
+                images = workspace.image_set
+                x = images.get_image(self.image_name.value)
+                properties=skimage.measure.regionprops(mask_in.astype(int),intensity_image=x.pixel_data)
```
AetherUnbound: There should be spaces around the assignment and after the comma to follow pep8

cellprofiler/modules/savecroppedobjects.py
```diff
@@ -87,7 +87,10 @@ def create_settings(self):
             ],
             value=O_TIFF,
             doc="""\
-**%(O_PNG)** files do not support 3D. **%(O_TIFF)** files use zlib compression level 6."""
+**{O_PNG}** files do not support 3D. **{O_TIFF}** files use zlib compression level 6.""".format(
```
AetherUnbound: Nice catch!

bethac07: Wouldn't have caught it (and other documentation problems too- some of the added documentation was in html, not rst) if you hadn't asked me to change the formatting thing, so good on you!

PR #3676, opened 2018-10-31
Add ability to select measurements in MeasureColocalization

bethac07:


cellprofiler/modules/measurecolocalization.py
```diff
@@ -897,53 +1038,25 @@ def get_measurement_images(self, pipeline, object_name, category,
...
+            raise NotImplementedError("Automatic upgrade for this module is not supported in CellProfiler 3.")
+
         if variable_revision_number == 2:
             image_count = int(setting_values[0])
             idx_thr = image_count + 2
             setting_values = \
                 setting_values[:idx_thr] + ["15.0"] + setting_values[idx_thr:]
             variable_revision_number = 3
 
+        if variable_revision_number == 3:
```
AetherUnbound: FWIW I don't think adding a bunch of values on the end of the settings list doesn't need a settings revision. It _should_ just add the defaults (in this case `True`) for all the new settings. I'd try it - setting revision 4 may not be necessary!

bethac07: I can and will check that, but my personal gut feeling is that since we've added a bunch of settings even though we've kept compatibility we should bump the version anyway.  @0x00b1 what do you think there?

0x00b1: @AetherUnbound I agree. However, it really depends on how you view the `variable_revision_number`. I think from a user perspective it should be upgraded (but you're right, it shouldn't need to be updated). I suggest @bethac07 (or someone else) comes up with a policy around when to update the `variable_revision_number` that we start to adhere to in future changes. What do you think?

AetherUnbound: I agree with @bethac07 here - from what I understand it's pretty much only for internal use - so if we don't have to change it then I don't think we should.

Issue #3677, opened 2018-11-01
Objects created by RelateObjects do not initialize with a Count


bethac07:
Objects created by Watershed and the Identify modules initialize with an object count and a per-image count; it does not appear this happens in RelateObjects.  It doesn't seem to fully add to the object set at all, since it has no measurements.
If you therefore attempt to add measurements to an object created by RelateObjects, an error is thrown when you get to ExportToSpreadsheet and/or ExportToDatabase, because it can't write out the per-image count (presumably it would then throw an error when it tries to write out the object count in the Object spreadsheet too, but it doesn't get that far).
Pipeline to reproduce below, based on the WelcomeScreen images.

![image](https://user-images.githubusercontent.com/6721515/47859869-9d2bf680-ddc5-11e8-85b5-1828cce74739.png)

[RelateObjectsError.cppipe.zip](https://github.com/CellProfiler/CellProfiler/files/2538956/RelateObjectsError.cppipe.zip)

bethac07:
I'm not really sure, but comparing between IPO and Relate I have a sneaking suspicion it's this- https://github.com/CellProfiler/CellProfiler/blob/master/cellprofiler/modules/identifyprimaryobjects.py#L1067

AetherUnbound:
I think you're right @bethac07! What a simple thing for me to overlook when I added the child objects option. I'll add that into relate and see if that fixes it. Thanks!

AetherUnbound:
Damn, unfortunately I don't think that's it because I'm adding the objects here :( https://github.com/CellProfiler/CellProfiler/blob/master/cellprofiler/modules/relateobjects.py#L403-L411

AetherUnbound:
Update: I _might_ be missing this! https://github.com/CellProfiler/CellProfiler/blob/master/cellprofiler/module.py#L1328 (this isn't present in RelateObjects)

bethac07:
Fixed in 319

PR #3680, opened 2018-11-08
#3677 Add default measurements for RelateObjects

AetherUnbound:
Since we're overriding the "run" method, we also have to make sure that `add_measurements` is called to add the default measurements for an ObjectProcessing module. 

@bethac07 do you mind making sure this works? It looks like it works on my machine (the run is successful and a CSV is made).

### DIFF
--- a/cellprofiler/modules/relateobjects.py
+++ b/cellprofiler/modules/relateobjects.py
@@ -410,6 +410,8 @@ parents or children of the parent object."""
 
         objects.add_objects(y, y_name)
 
+        self.add_measurements(workspace)
+
         if self.show_window:
             workspace.display_data.parent_labels = parents.segmented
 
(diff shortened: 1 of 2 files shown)


bethac07:
It no longer errors out, but something is still not right- when I try to measure the objects made by RelateObjects, most of them come up with all measurements as 0 and nan

image

AetherUnbound:
Dang, thanks for finding that Beth. I'll keep looking into it!

AetherUnbound:
image
Hey, that looks promising! Lemme commit this, do you mind giving this a shot @bethac07?

AetherUnbound:
Crap, this build is failing because pytest 4.0 was just relased. I'll get a PR out for that.

bethac07:
Something is still not right.  I made a pipeline where the Parent is Nuclei, the Child is IdentifyPrimaryObjects, and the resulting objects are RelateObjects.  Theoretically, since RelateObjects are just IdentifyPrimaryObjects inside Nuclei, the counts for the two sets of Children objects should be the same, but they're massively different.  

image

AetherUnbound:
Dang, do you mind sharing that pipeline?

bethac07:
Use the Welcome Screen Pipeline images
[Issues3677.cpproj.zip](https://github.com/CellProfiler/CellProfiler/files/2596920/Issues3677.cpproj.zip)

bethac07:
Closing in favor of #3799

Issue #3541, opened 2018-04-02
DisplayDataOnImage np error


bethac07:
When trying to run the pipeline and image sets below (H/T the [forum](http://forum.cellprofiler.org/t/cp3-3-0-error-with-overlayoutlines-and-displaydataonimage/5929/2)) in 3.0.0 or 3.1.3, the following error is encountered.  From my testing it doesn't matter which image you choose to display data on, or even if you say you want no background image at all.  The pipeline works just fine in 2.2, and the relevant line hasn't been changed in 2 years according to the module blame, so I assume this is something to do with changes to numpy.  At that, we've reached my end of troubleshooting ability.

```
Failed to run module DisplayDataOnImage
Traceback (most recent call last):
  File "cellprofiler\gui\pipelinecontroller.py", line 2888, in do_step
    self.__pipeline.run_module(module, workspace)
  File "cellprofiler\pipeline.py", line 2031, in run_module
    module.run(workspace)
  File "cellprofiler\modules\displaydataonimage.py", line 296, in run
    mask = ~(np.isnan(values) | np.isnan(x) | np.isnan(y))
ValueError: operands could not be broadcast together with shapes (1029,) (1023,)
```

[DisplayDataOnImage.zip](https://github.com/CellProfiler/CellProfiler/files/1869689/DisplayDataOnImage.zip)

AetherUnbound:
This looks to be a mismatch between the number of objects (1029), and the number of calculated centers (1023), so I think this issue might be further upstream from the DisplayDataOnImage module

AetherUnbound:
I've updated the forum post regarding this fix (travis is still failing due to the cache error we're addressing).

bethac07:
So digging into this, you were 100 percent right about the issue being further upstream.  In this pipeline, primary objects can be of any size, and there are 1041.   The number of secondary objects SHOULD equal the number of primary objects, and the number of indices do, but even when I turn filtering off in ISO it tells me there should be 1041 labels but it only has 1035 centers. 

It appears to only happen when you turn on the hole filling setting- when hole filling is off, you start with 1098 labels (before you do a filtering step), then after filtering end up with 1041; when you have hole filling on, you have 1098 labels, then after hole filling you somehow only have 1086, which after filtering gives you 1035.  The bug is therefore likely somewhere in https://github.com/CellProfiler/centrosome/blob/master/centrosome/cpmorphology.py#L33 .  My guess is that this is also what's causing #3645

[...]

PR #3690, opened 2018-11-26
revert changes to object count method

bethac07:


AetherUnbound:
Has this been confirmed to resolve the indexing issue?

AetherUnbound:
I think this will cause a regression on #3541. This is totally fine (I think my change caused many more problems than it solved), I just want to make note of it here!

bethac07:
It does seem to fix the issue, and it does indeed re-break 3541.  I'll reopen that.

AetherUnbound:
LGTM though! Again, my naïveté with this codebase I think caused more damage than good in that PR lol. I will go through and find another solution to that particular issue!

Issue #3620, opened 2018-08-07
Feature request: SaveImage should support Hdf5


votti:
In our workflow we like to pass multicolor image stacks between CellProfiler and Ilastik (https://github.com/BodenmillerGroup/ImcSegmentationPipeline).
In order to allow our things to work with vanilla CP3 without any additional dependencies, I noticed that saving color stacks to the 'h5' format works reliably. Compared to normally saved tiffs with the `SaveImage` module, this has the advantage that Ilastik does recognize it as `yxc` right away and not as `tyx`.
Although it is possible to change `tyx` to `cyx` within Ilastik, it has to be manually done (which is inconvenient for beginners) and also gives problems when batch processing.

I modified your `SaveImages` module to do this export to hdf5: https://github.com/BodenmillerGroup/ImcPluginsCP/blob/develop-cp3/plugins/saveimages_h5.py

Based on the example I found at: https://github.com/ilastik/ilastik/blob/master/bin/combine_channels_as_h5.py

However I was wondering if it could make sense to just add this option to your `SaveImage` module.
If this would be an option for you, I would be happy to take time to write a corresponding pull request.

AetherUnbound:
I'm personally not against this, though I'd love to hear @0x00b1's opinion! I'd say go ahead and submit a PR (as it seems you have most of the implementation already) :smile: If you'd like to wait for Allen too that'd be fine.

I'll go ahead and assign it to you :+1:

PR #3694, opened 2018-12-05
Adds Ilastik compatible hdf5 image saving

votti:
This implements saving images in an Ilastik compatible Hdf5 format as discussed in issue: #3620

It is based on the following script, which I was directed to when I asked once about the format specifications:
https://github.com/ilastik/ilastik/blob/master/bin/combine_channels_as_h5.py
Currently saving as Uint8, Uint16 and float is supported. Volumetric data should also work.

Currently, I only did some manual testing but I will look into adding some proper tests soon.


bethac07:
Thanks so much for this, sorry for the delay in pulling!

PR #3696, opened 2018-12-05
Make FilterObjects respect masks

### LINKED ISSUE #3695: FilterObjects doesn't respect masks (but it's supposed to)
When you use the option in FilterObjects to "remove objects touching image or mask borders", it removes images touching image borders but not mask borders - if you compare the top two images to the bottom one, they should all be identical, but it's clearly not.  Pipeline below, to be used with the ExampleFly images

image

[Issues3695.cppipe.zip](https://github.com/CellProfiler/CellProfiler/files/2648912/Issues3695.cppipe.zip)

bethac07:
Resolves #3695

### DIFF
--- a/cellprofiler/modules/filterobjects.py
+++ b/cellprofiler/modules/filterobjects.py
@@ -878,7 +878,15 @@ value will be retained.""".format(**{
         '''
         labels = src_objects.segmented
 
-        interior_pixels = scipy.ndimage.binary_erosion(numpy.ones_like(labels))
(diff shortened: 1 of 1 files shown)


AetherUnbound:
Are there any tests that can be added for this new code path?

bethac07:
At least I should probably try to figure why the previous test `test_09_02_discard_mask_objects` failed...

bethac07:
I'll fix the tests (and made an issue to do so), but I want to make sure this gets pulled in before

PR #3704, opened 2018-12-13
add JVM environment fix

bethac07:
This is ugly, and I'm very sorry about that, but it has to be called before bioformats is imported because bioformats checks if it can spin up javabridge upon import.  It also works to add this in the init if you find it less gross to have it there.

### DIFF
--- a/cellprofiler/__main__.py
+++ b/cellprofiler/__main__.py
@@ -1,13 +1,14 @@
 from __future__ import print_function
 
 import os
-import os.path
 import sys
 if sys.platform.startswith("win") and hasattr(sys, 'frozen'):
-    # Allow us to package java with windows builds
+    # For Windows builds, if JAVA_HOME is not already set, use the copy of Java packaged with CP
+    # We specify this location by setting 'CP_JAVA_HOME' at install.
+    # JAVA_HOME must be set before bioformats import.
     if 'JAVA_HOME' not in os.environ:
-        if 'CP_JAVAHOME' in os.environ:
-            os.environ['JAVA_HOME'] = os.environ['CP_JAVAHOME']
+        if 'CP_JAVA_HOME' in os.environ:
+            os.environ['JAVA_HOME'] = os.environ['CP_JAVA_HOME']
 
 import bioformats.formatreader
 import ctypes
@@ -28,6 +29,7 @@ import logging.config
 import matplotlib
 import numpy
 import optparse
+import os.path
 import pkg_resources
 import re
 import site
--- a/cellprofiler/__main__.py
+++ b/cellprofiler/__main__.py
@@ -1,14 +1,14 @@
 from __future__ import print_function
 
 import os
+import os.path
 import sys
 if sys.platform.startswith("win") and hasattr(sys, 'frozen'):
     # For Windows builds, if JAVA_HOME is not already set, use the copy of Java packaged with CP
     # We specify this location by setting 'CP_JAVA_HOME' at install.
     # JAVA_HOME must be set before bioformats import.
-    if 'JAVA_HOME' not in os.environ:
-        if 'CP_JAVA_HOME' in os.environ:
-            os.environ['JAVA_HOME'] = os.environ['CP_JAVA_HOME']
+    if 'JAVA_HOME' not in os.environ and if 'CP_JAVA_HOME' in os.environ:
+        os.environ['JAVA_HOME'] = os.environ['CP_JAVA_HOME']
 
 import bioformats.formatreader
 import ctypes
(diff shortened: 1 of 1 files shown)


cellprofiler/__main__.py
```diff
@@ -1,4 +1,14 @@
 from __future__ import print_function
+
+import os
+import os.path
```
AetherUnbound: It looks like `os.path` isn't used, this can probably be removed

cellprofiler/__main__.py
```diff
@@ -1,4 +1,14 @@
...
+import sys
+if sys.platform.startswith("win") and hasattr(sys, 'frozen'):
+    # Allow us to package java with windows builds
+    if 'JAVA_HOME' not in os.environ:
+        if 'CP_JAVAHOME' in os.environ:
```
AetherUnbound: This should probably be `CP_JAVA_HOME` for consistency

cellprofiler/__main__.py
```diff
@@ -1,4 +1,14 @@
 from __future__ import print_function
+
+import os
+import os.path
+import sys
+if sys.platform.startswith("win") and hasattr(sys, 'frozen'):
+    # Allow us to package java with windows builds
+    if 'JAVA_HOME' not in os.environ:
```
AetherUnbound: This _could_ be the following, which might have some better clarity:
```
# Set JAVA_HOME to be CP_JAVA_HOME if it is available, otherwise set to an empty string
# (os.environ does not accept None as a valid environment variable)
os.environ['JAVA_HOME'] = os.environ.get('JAVA_HOME') or os.environ.get('CP_JAVA_HOME') or ''
```

bethac07: I don't think we want to do the empty string setting because I don't think we want to actually set JAVA_HOME in case we need Java to do its normal thing to find copies of itself.  Just for maximum safety.

cellprofiler/__main__.py
bethac07: You're right, that's a leftover from a previous attempt to fix this.  Previously both os and os.path WERE set though- do we want to maintain that (and move os.path down into the rest of the imports) or do we want to just import os?

cellprofiler/__main__.py
AetherUnbound: Okay, then I'd just have this logic be `if 'JAVA_HOME' not in os.environ and 'CP_JAVA_HOME` in os.environ: <assigment>`

cellprofiler/__main__.py
bethac07: K

cellprofiler/__main__.py
bethac07: I can add more clarity to the comment though.

cellprofiler/__main__.py
AetherUnbound: Oh oops I didn't realize it was used below! No I'd leave it up here then.

cellprofiler/__main__.py
bethac07: inside the sys.platform.startswith, or outside it?

cellprofiler/__main__.py
```diff
@@ -1,14 +1,14 @@
...
+import os.path
 import sys
 if sys.platform.startswith("win") and hasattr(sys, 'frozen'):
     # For Windows builds, if JAVA_HOME is not already set, use the copy of Java packaged with CP
     # We specify this location by setting 'CP_JAVA_HOME' at install.
     # JAVA_HOME must be set before bioformats import.
-    if 'JAVA_HOME' not in os.environ:
-        if 'CP_JAVA_HOME' in os.environ:
-            os.environ['JAVA_HOME'] = os.environ['CP_JAVA_HOME']
+    if 'JAVA_HOME' not in os.environ and if 'CP_JAVA_HOME' in os.environ:
```
AetherUnbound: Uhh, this may throw a syntax error. I think it should be 
```
if 'JAVA_HOME' not in os.environ and 'CP_JAVA_HOME' in os.environ:
```

bethac07: lolyup

cellprofiler/__main__.py
AetherUnbound: Oh sorry, no your current implementation looks good!

cellprofiler/__main__.py
jni: Why? importing `os` gives one access to `os.path`.

0x00b1: @jni Unfortunately a lot of tools don't pick up on relative imports without explicit declaration. 🤷‍♂️

PR #3720, opened 2019-02-14
Use ImageIO by default when reading and writing images

0x00b1:
Hooray!


AetherUnbound:
Uh oh, looks like some build problems :(

PR #3757, opened 2019-05-30
WIP: Wxpython4 drag drop modules fix

jeffsyang:


jeffsyang:
I forgot to sync our fork so i'll need to do that first before sending another pr

PR #3758, opened 2019-05-30
Drag and drop reordering for pipeline modules

jeffsyang:
@bethac07 @0x00b1


cellprofiler/gui/pipelinelistview.py
```diff
@@ -1041,7 +1041,7 @@ def OnData(self, x, y, action):
                 self.data_object.GetReceivedFormat().GetType()
                 == self.pipeline_data_object.GetFormat().GetType()
             ):
-                pipeline_data = self.pipeline_data_object.GetDataHere()
+                pipeline_data = self.pipeline_data_object.GetData().tobytes().replace(b'\00', '')
```
AetherUnbound: I notice you're replacing a byte sequence `b'\00'` with a string - is `pipeline_data` supposed to be a string here or a `bytes`?

jeffsyang: i believe it should be a string. if that's the case, I'll just wrap it in a `str()`?

AetherUnbound: I think it's worth it for clarity!

0x00b1:
Thanks @jeffsyang. Please rebase in the future! I manually added the line in #5fe8dc768.

PR #3763, opened 2019-06-06
unresolved reference since no python 2 buffer() in python 3

jeffsyang:
removed python2 code

@0x00b1 @bethac07

### DIFF
--- a/cellprofiler/utilities/hdf5_dict.py
+++ b/cellprofiler/utilities/hdf5_dict.py
@@ -25,8 +25,6 @@ install_aliases()
 import six.moves.urllib.parse
 
 
-buffer = memoryview
-
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


cellprofiler/utilities/hdf5_dict.py
```diff
@@ -25,10 +25,7 @@
 import six.moves.urllib.parse
 
 
-try:
-    buffer  # Python 2
-except NameError:  # Python 3
-    buffer = memoryview
+buffer = memoryview
```
AetherUnbound: This variable existed to provide a common interface for these keywords - any instance of `buffer` in these files should just be changed to `memoryview` and this assignment should be removed entirely

jeffsyang: will do

PR #3764, opened 2019-06-10
WIP: read in text mode instead of binary mode

jeffsyang:
@0x00b1 @AetherUnbound @bethac07


AetherUnbound:
What's the reason for this change? Not complaining! Just want to know :)

jeffsyang:
> What's the reason for this change? Not complaining! Just want to know :)

Please correct me if I'm mistaken, but it appears that string and byte functions aren't as flexible in python3 as compared to python2.7. Noticing that there was a move to remove `.encode()`/`.decode()` functions, I thought that it'll be easier to read cppipe files as strings, in contrast to bytes, ultimately steering away from the need to encode and decode later on.

PR #3765, opened 2019-06-10
updating to extract exception message string correctly

jeffsyang:
@bethac07 @0x00b1 @AetherUnbound


jeffsyang:
i'll make this change in the PR https://github.com/CellProfiler/CellProfiler/pull/3764 to keeps things a bit simpler

PR #3766, opened 2019-06-11
Python3 load example cppipe continuation

jeffsyang:

hey all, i understand that i had reincorporated some of the en/decode stuff but unless we remove the \x__ sequences in the example cppipe, i don't think it's possible to load it. Please let me know if i'm missing anything.
-JS

Update: I've also tested loading the pipelines in repo https://github.com/CellProfiler/examples

### DIFF
--- a/cellprofiler/gui/pipelinelistview.py
+++ b/cellprofiler/gui/pipelinelistview.py
@@ -142,7 +142,7 @@ class PipelineListView(object):
         self.outputs_panel = wx.Panel(panel)
         outputs_sizer.Add(self.outputs_panel, 1, wx.EXPAND)
         self.outputs_panel.SetSizer(wx.BoxSizer())
-        self.outputs_panel.SetBackgroundStyle(wx.BG_STYLE_COLOUR)
+        self.outputs_panel.SetBackgroundStyle(wx.BG_STYLE_PAINT)
         self.outputs_button = wx.Button(
             self.outputs_panel, label="View output settings", style=wx.BU_EXACTFIT
         )
--- a/cellprofiler/pipeline.py
+++ b/cellprofiler/pipeline.py
@@ -1147,9 +1147,8 @@ class Pipeline(object):
                 #
                 # Set up the module
                 #
-                module_name = module_name
                 module = self.instantiate_module(module_name)
-                module.module_num = module_number
+                module.set_module_num(module_number)
                 #
                 # Decode the attributes. These are turned into strings using
                 # repr, so True -> 'True', etc. They are then encoded using
@@ -1174,12 +1173,11 @@ class Pipeline(object):
                     if len(a.split(":")) != 2:
                         raise ValueError("Invalid attribute string: %s" % a)
                     attribute, value = a.split(":")
-                    value = value
+                    if attribute in skip_attributes:
+                        continue
                     value = eval(value)
                     if attribute == "variable_revision_number":
                         variable_revision_number = value
-                    elif attribute in skip_attributes:
-                        pass
                     else:
                         setattr(module, attribute, value)
                 if variable_revision_number is None:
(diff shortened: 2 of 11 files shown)


cellprofiler/pipeline.py
```diff
@@ -1174,12 +1175,13 @@ def rl():
                     if len(a.split(":")) != 2:
                         raise ValueError("Invalid attribute string: %s" % a)
                     attribute, value = a.split(":")
-                    value = value
-                    value = eval(value)
+                    if attribute in skip_attributes:
+                        continue
+                    # En/decode needed to read example cppipe format
+                    # TODO: remove en/decode when example cppipe no longer has \x__ characters
```
AetherUnbound: Sorry, this is a bug I introduced in trying to carve out some of the home-brewed UTF-16 encoding/decoding :) 
I don't think this should stop your PR from being merged, but I think it might be worth trying this with an example pipeline that _doesn't_ have those escape characters. This would be a new "version" of the pipeline file (you can see in this file I think how the different versions are handled) but I think the quicker we can get away from that the better!

jeffsyang: @AetherUnbound @bethac07 @0x00b1 would you happen to know where I could get such a pipeline? If not, I could always edit one manually

bethac07: Most of the ones here should work https://github.com/CellProfiler/examples

cellprofiler/gui/pipelinelistview.py
```diff
@@ -142,7 +142,7 @@ def __init__(self, panel, frame):
         self.outputs_panel = wx.Panel(panel)
         outputs_sizer.Add(self.outputs_panel, 1, wx.EXPAND)
         self.outputs_panel.SetSizer(wx.BoxSizer())
-        self.outputs_panel.SetBackgroundStyle(wx.BG_STYLE_COLOUR)
+        self.outputs_panel.SetBackgroundStyle(wx.BG_STYLE_PAINT)
```
jeffsyang: @bethac07 @0x00b1 @AetherUnbound it seems like BG_STYLE_COLOUR is no longer supported even though it's still listed in the documentation (https://wxpython.org/Phoenix/docs/html/wx.BackgroundStyle.enumeration.html); could you please let me know which background style is preferred?

cellprofiler/pipeline.py
jeffsyang: @AetherUnbound After creating a "cleaned" version of ExampleFly.cppipe (opened in python interpreter, ran `file.encode().decode('unicode_escape')`, and saved it as another cppipe file), this branch's Cellprofiler appears to read it successfully (v3.1.8 is able to read the cleaned version as well). That said, I had to hack around to get images to load (doesn't appear to be an issue with the cppipe file but the threading which populates the pipelinecontroller's pathlist).

Hack Steps:
1. loaded the welcome screen example pipeline for the images
2. deleted a couple modules to make sure next step works
3. file->imported cleaned cppipe
4. checked that deleted modules reappeared
5. 'Start test mode'
6. was able to step through a few modules (some still need fixing)

bethac07: What about the other example pipelines?

AetherUnbound: Glad the read in worked! Next step is to get it to write out without that garbage encoding stuff!

jeffsyang: @bethac07 haven't gotten to around that yet but it's the next thing on my to-do list
@AetherUnbound good catch, that slipped my mind actually

thanks for the feedback!  👍

jeffsyang: @bethac07 after importing all the cppipe files in the example repo, a common issue that came up was how python 3 treats int divided (/) by int. It resolves into a float whereas // resolves the result as an int. I have incorporated the fixes in Commit https://github.com/CellProfiler/CellProfiler/pull/3766/commits/1c257ae0b6bebf590621822d16cc67f10befb3b8 in this branch.

jeffsyang: @AetherUnbound I believe it no longer writes out the cppipe with the encoding stuff. Please take a look at PR https://github.com/CellProfiler/CellProfiler/pull/3768.

PR #3768, opened 2019-06-14
Export cppipe as text with no encoding

jeffsyang:
when exporting cppipe, it will no longer encode :|[]\ into \x__ sequences
also bumped up cppipe format versioning

This PR is meant to work in conjunction with PR https://github.com/CellProfiler/CellProfiler/pull/3766 but was separated to make it easier to scrap if incorrect.

@0x00b1 @AetherUnbound @bethac07

### DIFF
--- a/cellprofiler/pipeline.py
+++ b/cellprofiler/pipeline.py
@@ -153,7 +153,7 @@ FMT_MATLAB = "Matlab"
 FMT_NATIVE = "Native"
 
 """The current pipeline file format version"""
-NATIVE_VERSION = 4
+NATIVE_VERSION = 5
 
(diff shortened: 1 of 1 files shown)


cellprofiler/pipeline.py
```diff
@@ -1238,20 +1239,6 @@ def save(self, fd_or_filename, format=FMT_NATIVE, save_image_plane_details=True)
         else:
             raise NotImplementedError("Unknown pipeline file format: %s" % format)
 
-    def encode_txt(self, s):
```
AetherUnbound: I am SO glad to see these go! Way to go @jeffsyang :tada:

PR #3769, opened 2019-06-18
[WIP] Find a more memory-efficient way to handle relating objects

bethac07:
I do not claim that this is the correct way to do this, and I am testing its performance right now on a case where the normal functionality barfs. Just spaghetti at the wall to see what works rn.

### DIFF
--- a/cellprofiler/object.py
+++ b/cellprofiler/object.py
@@ -4,6 +4,7 @@ import matplotlib.cm
 import numpy
 import scipy.ndimage
 import scipy.sparse
+import scipy.stats
 import skimage.color
 
 import cellprofiler.preferences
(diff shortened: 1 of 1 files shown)


cellprofiler/object.py
```diff
@@ -308,10 +309,22 @@ def relate_children(self, children):
...
+		parent_list=[]
+		parent_labels=self.segmented
+		parent_labels=numpy.where(parent_labels==0, numpy.NaN, parent_labels) #ignore the background
+		child_labels=children.segmented
+		for eachchild in range(1,numpy.max(child_labels)+1):
+		    child_hist.append(int(scipy.stats.mode((numpy.where(child_labels==eachchild, 1, numpy.NaN)) * parent_labels, axis=None, nan_policy='omit')[0][0]))
+		for eachparent in range(1,numpy.max(child_labels)+1):
+                    parent_list.append(child_hist.count(eachparent))
+		return numpy.array(child_hist), numpy.array(parent_list)		
+	    except:
```
AetherUnbound: I'd highly recommend catching a specific exception here when you're able to determine if this is more efficient :)

bethac07: For sure!  Just waiting to see if a) it'll build and b) it'll run without dying on the dataset that currently dies when I try to run it.  I would try to add some list comprehensions, etc too. 

_A priori_ it seems logical to me that just checking what parent values are present in the child would be more memory efficient (if not necessarily clock-faster) than computing the whole coo matrix, but my instincts for this are 100% a WIP- do you have any sense on if you think this might work? I should know in the next couple hours I guess.

AetherUnbound: I totally agree - I think that's a better heuristic, and then excluding those children from the coo matrix computation

bethac07:
This is doing ok memory wise, but super non performant.  Boo.

PR #3772, opened 2019-06-24
Image Loading

jeffsyang:

### DIFF
--- a/cellprofiler/gui/pathlist.py
+++ b/cellprofiler/gui/pathlist.py
@@ -114,7 +114,7 @@ class PathListCtrl(wx.ScrolledWindow):
             self.SetFont(self.DROP_FILES_AND_FOLDERS_FONT)
             self.drop_files_and_folders_text_extent = self.GetFullTextExtent(
                 self.DROP_FILES_AND_FOLDERS_HERE
-            )
+            )[:2]
         except:
             logger.warn(
                 'Failed to get text extend for "%s" message'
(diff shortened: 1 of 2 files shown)


cellprofiler/gui/pathlist.py
```diff
@@ -114,7 +114,7 @@ def __init__(self, *args, **kwargs):
             self.SetFont(self.DROP_FILES_AND_FOLDERS_FONT)
             self.drop_files_and_folders_text_extent = self.GetFullTextExtent(
                 self.DROP_FILES_AND_FOLDERS_HERE
-            )
+            )[:2]
```
AetherUnbound: Did the return type for this change? This seems suspect

cellprofiler/gui/pipelinecontroller.py
```diff
@@ -2112,36 +2112,30 @@ def add_paths_to_pathlist(self, filenames):
...
+                dlg.SetSize(min_width, h)
+
             queue = six.moves.queue.Queue()
             interrupt = [False]
             message = ["Initializing"]
 
             def fn(filenames=filenames, interrupt=None, message=None, queue=queue):
-                if message is None:
-                    message = message
-                if interrupt is None:
```
AetherUnbound: Are we asserting that interrupt will now exist? It seems possible for it to be `None` in which case `interrupt[0]` will fail.

cellprofiler/gui/pathlist.py
jeffsyang: tbh, i'm not entirely sure how to interpret the documentation at https://wxpython.org/Phoenix/docs/html/wx.GraphicsContext.html#wx.GraphicsContext.GetFullTextExtent but after stepping through the code, it returned 4 values, hence the splice. Please let me know if a better method comes to mind.

cellprofiler/gui/pipelinecontroller.py
jeffsyang: hm... i might have made a mistake here; i'll need to run some tests and figure out the details regarding threading and sharing variables.

PR #3790, opened 2019-07-08
Python3 load images on windows

jeffsyang:
@0x00b1 @bethac07 @AetherUnbound 

On windows, Config.Exist() appears to be sufficient but introduced a quick hack to get it working in the meantime.

Issue #3781, opened 2019-07-01
In Python3 branch, visualization of image sets failed


bethac07:
Image sets are correctly made and can be run in test mode, but no columns appear and you get the following stack trace:

```
"Traceback (most recent call last):
  File ""c:\users\administrator\appdata\local\programs\python\python37\lib\site-packages\wx\core.py"", line 2228, in Notify
    self.notify()
  File ""c:\users\administrator\appdata\local\programs\python\python37\lib\site-packages\wx\core.py"", line 3384, in Notify
    self.result = self.callable(*self.args, **self.kwargs)
  File ""c:\users\administrator\documents\github\cellprofiler\cellprofiler\cellprofiler\gui\moduleview.py"", line 2434, in do_reset
    self.set_selection(self.__module.module_num)
  File ""c:\users\administrator\documents\github\cellprofiler\cellprofiler\cellprofiler\gui\moduleview.py"", line 656, in set_selection
    self.__frame.reset_imageset_ctrl(refresh_image_set=False)
  File ""c:\users\administrator\documents\github\cellprofiler\cellprofiler\cellprofiler\gui\cpframe.py"", line 443, in reset_imageset_ctrl
    self.__imageset_ctrl.recompute()
  File ""c:\users\administrator\documents\github\cellprofiler\cellprofiler\cellprofiler\gui\imagesetctrl.py"", line 1024, in recompute
    n_rows_added, n_columns_added = self.table.recompute()
  File ""c:\users\administrator\documents\github\cellprofiler\cellprofiler\cellprofiler\gui\imagesetctrl.py"", line 96, in recompute
    self.columns = self.get_columns()
  File ""c:\users\administrator\documents\github\cellprofiler\cellprofiler\cellprofiler\gui\imagesetctrl.py"", line 205, in get_columns
    columns.sort(cmp=ordering_fn)
TypeError: 'cmp' is an invalid keyword argument for sort()"
```

![image](https://user-images.githubusercontent.com/6721515/60456091-57e6ef80-9c06-11e9-9046-d324bdc4ec1b.png)

PR #3792, opened 2019-07-08
Python3 issue 3781

jeffsyang:

For Issue https://github.com/CellProfiler/CellProfiler/issues/3781

Also added on top of PR https://github.com/CellProfiler/CellProfiler/pull/3790

### DIFF
--- a/cellprofiler/gui/imagesetctrl.py
+++ b/cellprofiler/gui/imagesetctrl.py
@@ -4,6 +4,7 @@
 
 
 import re
+import functools
 
 import numpy
 import six
@@ -202,7 +203,7 @@ class ImageSetGridTable(wx.grid.GridTableBase):
                 COL_ORDER.index(a.column_type), COL_ORDER.index(b.column_type)
             )
 
-        columns.sort(cmp=ordering_fn)
+        columns = sorted(columns, key=functools.cmp_to_key(ordering_fn))
 
         return columns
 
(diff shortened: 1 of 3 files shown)


cellprofiler/gui/imagesetctrl.py
```diff
@@ -202,7 +203,7 @@ def ordering_fn(a, b):
                 COL_ORDER.index(a.column_type), COL_ORDER.index(b.column_type)
             )
 
-        columns.sort(cmp=ordering_fn)
+        columns = sorted(columns, key=functools.cmp_to_key(ordering_fn))
```
AetherUnbound: Maybe outside the scope of this PR, but should `ordering_fn` be rewritten so that `cmp_to_key` is no longer needed?

jeffsyang: I think that's a good idea considering python 3.7+ doesn't appear to support the `cmp` argument anymore. i'm thinking it'll be easier if i make the change in this pr.

AetherUnbound: Sounds good, feel free to ping me when you have that ready

jeffsyang: hey, so, after looking at the source, I'm starting to think it'll be better if we continue to use `cmp_to_key()` since it's appears to be in a "refactored" form already. What do you think? (Source Snippet: http://code.activestate.com/recipes/576653-convert-a-cmp-function-to-a-key-function/)

0x00b1: @AetherUnbound and @jeffsyang Yeah. You want to use `sorted` with the `key` parameter. This works.

Issue #3784, opened 2019-07-01
In Python 3 branch, project loading fails


bethac07:
```
"Traceback (most recent call last):
  File ""c:\users\administrator\documents\github\cellprofiler\cellprofiler\cellprofiler\gui\pipelinecontroller.py"", line 687, in __on_open_workspace
    self.do_open_workspace(path)
  File ""c:\users\administrator\documents\github\cellprofiler\cellprofiler\cellprofiler\gui\pipelinecontroller.py"", line 798, in do_open_workspace
    self.__workspace.load(filename, load_pipeline)
  File ""c:\users\administrator\documents\github\cellprofiler\cellprofiler\cellprofiler\workspace.py"", line 409, in load
    self.measurements.hdf5_dict.hdf5_file
  File ""c:\users\administrator\documents\github\cellprofiler\cellprofiler\cellprofiler\utilities\hdf5_dict.py"", line 1027, in __init__
    assert g.attrs.get(A_CLASS, None) == CLASS_FILELIST_GROUP
AssertionError"
```

PR #3795, opened 2019-07-11
Python3 issue 3784

jeffsyang:
@bethac07 @0x00b1 @AetherUnbound 

For Issue https://github.com/CellProfiler/CellProfiler/issues/3784
Fix is ugly on this one so please let me know if a better idea comes to mind.

Btw, this PR doesn't yet fix the whole CP 3.1.8 cpproj loading bug, just the one addressed in the ticket. The problem remaining is to load the images from the cpproj file.

Added on top of PR https://github.com/CellProfiler/CellProfiler/pull/3790

### DIFF
--- a/cellprofiler/preferences.py
+++ b/cellprofiler/preferences.py
@@ -257,8 +257,10 @@ def config_exists(key):
         return True
     if not get_config().Exists(key):
         return False
-    if get_config().GetEntryType(key) == 1:
-        return get_config().Read(key) is not None
+    # FIXME: Issue reported at https://github.com/wxWidgets/Phoenix/issues/1292; quick hack for it to work for now
+    if sys.platform != "win32":
+        if get_config().GetEntryType(key) == 1:
+            return get_config().Read(key) is not None
     return True
 
 
--- a/cellprofiler/preferences.py
+++ b/cellprofiler/preferences.py
@@ -257,9 +257,22 @@ def config_exists(key):
         return True
     if not get_config().Exists(key):
         return False
-    # FIXME: Issue reported at https://github.com/wxWidgets/Phoenix/issues/1292; quick hack for it to work for now
+    # FIXME: Issue reported at https://github.com/wxWidgets/Phoenix/issues/1292, use respective .Read()
     if sys.platform != "win32":
         if get_config().GetEntryType(key) == 1:
+            return get_config().Read(key) is not None
+    else:
+        # Bool keys
+        if key in [SHOW_SAMPLING, TELEMETRY, TELEMETRY_PROMPT, STARTUPBLURB]:
+            return get_config().ReadBool(key) is not None
+        # Int keys
+        elif key in [SKIPVERSION, OMERO_PORT, MAX_WORKERS, JVM_HEAP_MB]:
+            return get_config().ReadInt(key) is not None
+        # Double keys
+        elif key in [TITLE_FONT_SIZE, TABLE_FONT_SIZE, PIXEL_SIZE]:
+            return get_config.ReadDouble(key) is not None
+        # String keys
+        else:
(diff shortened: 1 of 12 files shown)


cellprofiler/utilities/hdf5_dict.py
```diff
@@ -1024,6 +1024,12 @@ def __init__(self, hdf5_file, lock=None, filelist_name=DEFAULT_GROUP):
         g = self.hdf5_file.require_group(FILE_LIST_GROUP)
         if filelist_name in g:
             g = g[filelist_name]
+            # In Python 2, some hdf5 vals are stored as bytes. Convert to Python 3 unicode string
+            if g.attrs.get(A_CLASS, None):
+                try:
+                    g.attrs[A_CLASS] = g.attrs[A_CLASS].decode()
+                except AttributeError:
```
bethac07: What is this case handling?

cellprofiler/preferences.py
```diff
@@ -257,8 +257,10 @@ def config_exists(key):
         return True
     if not get_config().Exists(key):
         return False
-    if get_config().GetEntryType(key) == 1:
-        return get_config().Read(key) is not None
+    # FIXME: Issue reported at https://github.com/wxWidgets/Phoenix/issues/1292; quick hack for it to work for now
```
bethac07: I think we don't want to evade checking entirely on Windows, just in case we "forget"/deprioritize coming back to this; can you add an "else" that checks the text written to the config and handles it appropriately? For almost any other preference I wouldn't bother, but it's the "Can we track your usage" preference, so because of GDPR etc we don't want to ignore the user's stated intent.

cellprofiler/utilities/hdf5_dict.py
jeffsyang: when loading CP 3.1.8 cpproj files, `g.attrs[A_CLASS]` will be a `bytes` object, and because it's being compared to a `str`, it's causing the `assert` below to fail.

cellprofiler/preferences.py
jeffsyang: Sure; just a head's up, the fix I'm thinking isn't very pretty. The pseudo-code is something on the lines of

> else:
> => if key == some_bool_key:
> ===> return get_config().ReadBool(key) is not None
> => elif key == some_int_key>:
> ===> return get_config().ReadInt(key) is not None
> => etc...

Would this be alright for the time being?

cellprofiler/utilities/hdf5_dict.py
jeffsyang: I ended up scrapping this change and added an equality checker at `cellprofiler.utilities.legacy.equals()` to replace the `==` when comparing between bytes and strings. Thoughts?

0x00b1:
@jeffsyang I tried this and it didn't work with the following pipeline. Is this supposed to be a work-in-progress? If so, can you please change the PR type.

[ExampleFly.cppipe.zip](https://github.com/CellProfiler/CellProfiler/files/3458056/ExampleFly.cppipe.zip)

jeffsyang:
> @jeffsyang I tried this and it didn't work with the following pipeline. Is this supposed to be a work-in-progress? If so, can you please change the PR type.
> 
> [ExampleFly.cppipe.zip](https://github.com/CellProfiler/CellProfiler/files/3458056/ExampleFly.cppipe.zip)

@0x00b1 I made a few changes and was able to load cppipe files versions 3, 4, and 5. I didn't test with earlier versions than that though.

cellprofiler/preferences.py
jeffsyang: @bethac07 Just a head's up that with the new changes, on Windows, it will use the respective `.Read(key)` to check if the key exists.

PR #3799, opened 2019-08-07
Remake RelateObjects

bethac07:


bethac07:
Even though I'm now explicitly adding the measurements in L445 and L446, they still aren't being added to the workspace as far as I can tell.  SUPER frustrating.

bethac07:
So currently, despite the fact that I'm explicitly adding measurements (see 1), objects are being added to the workspace (see 2) but not the measurements (see 3).  Does this help trace it down at all?  


1) image

2) 
image


3) 
image

bethac07:
For anyone dying about this later, the secret was to properly add things to `get_measurement_columns`, `get_measurements`, and `get_categories`

PR #3808, opened 2019-08-14
test_createbatchfiles

jeffsyang:
had to modify pipeline.load() again to match tests
tested on ubuntu and windows for cppipe versions 3, 4, and 5


0x00b1:
LGTM!

PR #3810, opened 2019-08-14
Matplotlib requires fd in bytes for tests

jeffsyang:
fixed matplotlib error using BytesIO; although matplotlib 3.1.0 documentation mentions it taking a str, it is actually bytes

removed matlab v1 test and others

### DIFF
--- a/tests/modules/test_definegrid.py
+++ b/tests/modules/test_definegrid.py
@@ -29,52 +29,52 @@ OBJECTS_NAME = "objects"
 
 
 class TestDefineGrid(unittest.TestCase):
-    def test_01_02_load_v1(self):
-        data = (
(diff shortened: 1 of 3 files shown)


tests/modules/test_definegrid.py
```diff
@@ -29,53 +29,6 @@
 
 
 class TestDefineGrid(unittest.TestCase):
-    def test_01_02_load_v1(self):
```
AetherUnbound: What's the reason for removing this test?

jeffsyang: There was some difficulty fixing the load_v1 tests so after asking, Allen suggested removing them

bethac07: Makes sense then! If you guys can keep those conversations in the ip-cellprofiler channel, we will have access to them :)

PR #3811, opened 2019-08-14
Set module num via setter

jeffsyang:
common fix throughout the tests

this branch is meant for test that are fixed only by this change. tests that depend on this and other changes will be put on another PR


jeffsyang:
although haven't gone through all the tests yet, still ok to merge

PR #3832, opened 2019-09-03
Fix RelateObjects & Tests

bethac07:


cellprofiler/modules/relateobjects.py
```diff
@@ -828,17 +828,40 @@ def get_child_measurement_columns(self, pipeline):
...
+    def get_saved_child_measurement_columns(self, pipeline):
+        """Return measurements for saved child objects"""
         if self.wants_child_objects_saved:
             columns = super(RelateObjects, self).get_measurement_columns(
-                pipeline,
-                additional_objects=[
-                    (self.y_name.value, self.output_child_objects_name.value)
-                ],
-            )
+                pipeline,additional_objects=[(self.y_name.value, self.output_child_objects_name.value)])
```
AetherUnbound: Nit: newline after `pipeline,`

cellprofiler/modules/relateobjects.py
```diff
@@ -828,17 +828,40 @@ def get_child_measurement_columns(self, pipeline):
...
+            columns_to_return = []
+            for eachcolumn in columns:
```
AetherUnbound: Nit: maybe just `column` instead of `eachcolumn`?

cellprofiler/modules/relateobjects.py
```diff
@@ -828,17 +828,40 @@ def get_child_measurement_columns(self, pipeline):
...
+                if eachcolumn[0] == self.output_child_objects_name.value:
```
AetherUnbound: Should this `if..elseif` be two conditions in an `or` statement, since both result in the same operation?

cellprofiler/modules/relateobjects.py
```diff
@@ -828,17 +828,40 @@ def get_child_measurement_columns(self, pipeline):
...
+                    columns_to_return.append(eachcolumn)
+                elif self.output_child_objects_name.value in eachcolumn[1]:
+                    columns_to_return.append(eachcolumn)
+            return columns_to_return
         else:
-            columns = super(RelateObjects, self).get_measurement_columns(pipeline)
+            return []
```
AetherUnbound: You could potentially initialize this before the `if` statement as an empty list, and then have a single logic block within the `if` that adds the columns to that list. Then the function can unconditionally return the list, which will sometimes have columns in it (if that makes sense).

AetherUnbound:
Nice! 🙌

bethac07:
(Let me know if I didn't fix right 😄  )

bethac07:
Rebase!

PR #3845, opened 2019-09-26
Fix broken link

DavidStirling:
Fixed broken link in module tutorial


bethac07:
@0x00b1 is https ok, or does it need to be http? Otherwise LGTM

PR #3849, opened 2019-10-10
Revert to saved confirmation, fix module window placement, help updates

DavidStirling:
Per Nasim's suggestion - the "Revert to saved" button is right next to the standard save buttons, so a yes/no confirmation dialog would avoid frustration when people mis-click (as just happened). I've tried to add one!


DavidStirling:
Please excuse me using this to try fixing minor annoyances!

PR #3850, opened 2019-10-16
Create code of conduct

AnneCarpenter:
Does this look good to everyone?


bethac07:
LGTM

0x00b1:
Thanks, @AnneCarpenter. Would you mind changing the filename to `CONDUCT` (no file extension)?

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,77 @@
...
+when an individual is representing the project or its community. Examples of
+representing a project or community include using an official project e-mail
+address, posting via an official social media account, or acting as an appointed
+representative at an online or offline event. Representation of a project may be
+further defined and clarified by project maintainers.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported by contacting anne@broadinstitute.org or any other project team member
```
0x00b1: I recommend using your personal e-mail address.

AnneCarpenter:
> Thanks, @AnneCarpenter. Would you mind changing the filename to `CONDUCT` (no file extension)?

I don't mind terribly! But my understanding is that CODE_OF_CONDUCT.md is the standard for Github; at least it's what you get when you choose the default. And for a newbie, the longer form is more immediately recognizable.

CODE_OF_CONDUCT.md
AnneCarpenter: I believe I have changed it in the right place!

0x00b1:
We try to adopt POSIX-like style rather than GitHub recommendations (e.g. “LICENSE”).

0x00b1:
(For similar reasons it’d be great if you could also add 80 column line breaks.)

AnneCarpenter:
All set w the filename change. Don't know how to check/fix 80 column line breaks, it looked like it already is that, can you help?

PR #3856, opened 2019-10-23
Fix WX4 GUI bugs

DavidStirling:
Have been working through the spreadsheet of issues found during the previous round of CP4 testing.

### DIFF
--- a/cellprofiler/gui/cpframe.py
+++ b/cellprofiler/gui/cpframe.py
@@ -857,6 +857,7 @@ class CPFrame(wx.Frame):
         # ID_HELP_MODULE is used in _both_ button contexts and menu contexts,
         # so it needs event bindings for either type
         self.Bind(wx.EVT_MENU, self.__on_help_module, id=ID_HELP_MODULE)
+        self.Bind(wx.EVT_BUTTON, self.__on_help_module, id=ID_HELP_MODULE)
 
         self.Bind(wx.EVT_MENU, self.about, id=ID_HELP_ABOUT)
         self.Bind(wx.EVT_MENU, self.__on_preferences, id=ID_OPTIONS_PREFERENCES)
--- a/cellprofiler/gui/figure.py
+++ b/cellprofiler/gui/figure.py
@@ -655,7 +655,7 @@ class Figure(wx.Frame):
             if evt.xdata is not None:
                 x0 = min(self.mouse_down[0], evt.xdata)
                 x1 = max(self.mouse_down[0], evt.xdata)
-            if evt.ydata != None:
+            if evt.ydata is not None:
                 y0 = min(self.mouse_down[1], evt.ydata)
                 y1 = max(self.mouse_down[1], evt.ydata)
 
--- a/cellprofiler/gui/figure.py
+++ b/cellprofiler/gui/figure.py
@@ -652,7 +652,7 @@ class Figure(wx.Frame):
         else:
             x0, x1 = self.mouse_down[0], self.mouse_down[0]
             y0, y1 = self.mouse_down[1], self.mouse_down[1]
-            if evt.xdata != None:
+            if evt.xdata is not None:
                 x0 = min(self.mouse_down[0], evt.xdata)
                 x1 = max(self.mouse_down[0], evt.xdata)
             if evt.ydata != None:
--- a/cellprofiler/gui/figure.py
+++ b/cellprofiler/gui/figure.py
@@ -650,14 +650,10 @@ class Figure(wx.Frame):
             y0 = evt.ydata
             y1 = evt.ydata
         else:
-            x0, x1 = self.mouse_down[0], self.mouse_down[0]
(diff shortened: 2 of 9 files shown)


cellprofiler/gui/cpframe.py
```diff
@@ -857,6 +857,7 @@ def __add_menu(self):
         # ID_HELP_MODULE is used in _both_ button contexts and menu contexts,
         # so it needs event bindings for either type
         self.Bind(wx.EVT_MENU, self.__on_help_module, id=ID_HELP_MODULE)
+        self.Bind(wx.EVT_BUTTON, self.__on_help_module, id=ID_HELP_MODULE)
```
AetherUnbound: I'm interested about this change - why was the `EVT_BUTTON` binding needed here for help as well?

cellprofiler/gui/figure.py
```diff
@@ -650,10 +650,14 @@ def on_mouse_move(self, evt):
             y0 = evt.ydata
             y1 = evt.ydata
         else:
-            x0 = min(self.mouse_down[0], evt.xdata)
-            x1 = max(self.mouse_down[0], evt.xdata)
-            y0 = min(self.mouse_down[1], evt.ydata)
-            y1 = max(self.mouse_down[1], evt.ydata)
+            x0, x1 = self.mouse_down[0], self.mouse_down[0]
+            y0, y1 = self.mouse_down[1], self.mouse_down[1]
+            if evt.xdata != None:
```
AetherUnbound: I think the proper comparison here might be `if evt.xdata is not None`, same with below
```suggestion
            if evt.xdata is not None:
```

cellprofiler/gui/figure.py
```diff
@@ -650,10 +650,14 @@ def on_mouse_move(self, evt):
...
+                x0 = min(self.mouse_down[0], evt.xdata)
+                x1 = max(self.mouse_down[0], evt.xdata)
+            if evt.ydata != None:
```
AetherUnbound:
```suggestion
            if evt.ydata is not None:
```

cellprofiler/gui/figure.py
```diff
@@ -650,10 +650,14 @@ def on_mouse_move(self, evt):
             y0 = evt.ydata
             y1 = evt.ydata
         else:
-            x0 = min(self.mouse_down[0], evt.xdata)
-            x1 = max(self.mouse_down[0], evt.xdata)
-            y0 = min(self.mouse_down[1], evt.ydata)
-            y1 = max(self.mouse_down[1], evt.ydata)
+            x0, x1 = self.mouse_down[0], self.mouse_down[0]
```
AetherUnbound: Another thought - could this be done like so?

```
x = self.mouse_down[0]
x0 = min(x, evt.xdata or x)
x1 = max(x, evt.xdata or x)
```
The `or` would then default to `x` if it was `None`

cellprofiler/gui/pipelinecontroller.py
```diff
@@ -4176,6 +4187,7 @@ def __init__(self, file_callback_fn, text_callback_fn):
 
     def OnDropFiles(self, x, y, filenames):
         self.file_callback_fn(x, y, filenames)
+        return True
 
     def OnDropText(self, x, y, text):
         self.text_callback_fn(x, y, text)
```
AetherUnbound: Should this also return `True`?

cellprofiler/gui/cpframe.py
DavidStirling: It looks like the help module functions were rewritten not too long ago. All of them are present as links in the menu, but this module is also called by the module help button within the main GUI pane. Without a button bind the help button does not work.

cellprofiler/gui/figure.py
DavidStirling: Yep, thanks!

cellprofiler/gui/figure.py
DavidStirling: You may be right there. I expect there are a few ways to achieve this, so I wasn't sure which would be the best solution.

cellprofiler/gui/pipelinecontroller.py
DavidStirling: You mean OnDropText? I suspected it might, but I'm not sure how to properly trigger that function to test it.

0x00b1:
Thanks for reviewing @AetherUnbound!

cellprofiler/gui/pipelinecontroller.py
0x00b1: @AetherUnbound Yeah. I think this is correct: https://wxpython.org/Phoenix/docs/html/wx.TextDropTarget.html#wx.TextDropTarget.OnDropText

@DavidStirling please return `True` from `OnDropText`.

0x00b1:
LGTM. A minor issue with `OnDropText`.

PR #3869, opened 2019-11-12
A couple more CP4 fixes

DavidStirling:
Been working on a few more issues from the spreadsheet

### DIFF
--- a/cellprofiler/gui/editobjectsdlg.py
+++ b/cellprofiler/gui/editobjectsdlg.py
@@ -301,7 +301,7 @@ class EditObjectsDialog(wx.Dialog):
         #
         self.orig_axes = self.figure.add_subplot(1, 1, 1)
         self.orig_axes.set_zorder(1)  # preferentially select on click.
-        self.orig_axes._adjustable = "box-forced"
+        self.orig_axes._adjustable = "box"
         self.orig_axes.set_title(
             self.title,
             fontname=cellprofiler.preferences.get_title_font_name(),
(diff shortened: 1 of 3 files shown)


cellprofiler/gui/pipelinecontroller.py
```diff
@@ -2531,7 +2531,7 @@ def duplicate_modules(self, modules):
             module_num = len(self.__pipeline.modules()) + 1
         for m in modules:
             module = self.__pipeline.instantiate_module(m.module_name)
-            module.module_num = module_num
+            module.set_module_num(module_num)
```
DavidStirling: It looks like this behaves differently in Python 3.
From module.py - `module_num = property(get_module_num, set_module_num)`
In Python 2 it looks like trying to assign module_num works fine, but in Python 3 it triggers the get_ command which raises an exception (as the module_number we're trying to set is still the default of -1?). Calling the set_ command directly seems to get it working and allows you to duplicate modules.

setup.py
```diff
@@ -63,7 +63,7 @@ def find_resources(directory, subdirectory):
         "joblib>=0.13",
         "mahotas>=1.4",
         "matplotlib>=2.0.0, !=2.1.0",
-        "mysqlclient==1.4.2",
+        "mysqlclient==1.4.5",
```
DavidStirling: Only this version seems to be working on Python 3.7+. 1.4.4 installs but can't be imported and 1.4.2 didn't work at all.

cellprofiler/gui/editobjectsdlg.py
```diff
@@ -301,7 +301,7 @@ def print_figure(self, *args, **kwargs):
         #
         self.orig_axes = self.figure.add_subplot(1, 1, 1)
         self.orig_axes.set_zorder(1)  # preferentially select on click.
-        self.orig_axes._adjustable = "box-forced"
+        self.orig_axes._adjustable = "box"
```
DavidStirling: box-forced was deprecated and got removed, this doesn't appear to change anything.

PR #3871, opened 2019-11-22
[WIP] Issues/310

bethac07:
- Implement user option to avoid FlagImage-induced-skipping of final cycles
- Raise more informative error in SaveImages


bethac07:
Closing since we actually want this in master, not 319, just needed it working and to be able to confirm tests now pass.

PR #3874, opened 2019-12-02
Fix MeasureObjectNeighbors with two object sets

### LINKED ISSUE #2019: MeasureObjectNeighbors doesn't work for two different objects
Runs fine if both objects are set the same, errors out even if the number of both kinds of objects is forced to be the same; seems to be independent of which neighbor-finding method is used.  Still present as of the 20160504 nightly, Windows 10 64 bit.

> Traceback (most recent call last):
>   File "cellprofiler\gui\pipelinecontroller.pyc", line 2847, in do_step
>   File "cellprofiler\pipeline.pyc", line 2103, in run_module
>   File "cellprofiler\modules\measureobjectneighbors.pyc", line 430, in run
> IndexError: index 82 is out of bounds for axis 1 with size 78

See here http://forum.cellprofiler.org/t/measureobjectneighbors-error/3510 
And also http://forum.cellprofiler.org/t/win64-cp-2-2-0rc2-rev-1b13225-console-errors/3507/2
http://forum.cellprofiler.org/t/measure-the-object-from-related-objects/3176

DavidStirling:
Hopefully fixes #2019

### DIFF
--- a/cellprofiler/modules/convertobjectstoimage.py
+++ b/cellprofiler/modules/convertobjectstoimage.py
@@ -63,22 +63,21 @@ following options:
    colors for your objects.
 -  *Binary (black & white):* All object pixels will be assigned 1 and
    all background pixels will be assigned 0, creating a binary image.
--  *Grayscale:* Assigns all background pixels to 0 and each object a
-   different number from 1 to 255 (the maximum value that you can put in an 8-bit
-   integer) and numbers all pixels in each object with the object’s number.  This creates an image where
-   objects in the top left corner of the image are very dark and where the colors progress to white
-   toward the bottom right corner of the image. Use **SaveImages** to write the resulting image as a
-   .npy file or 8-bit or 16-bit .tiff file to disk if you want to process the label matrix image using
-   another program or in a separate CellProfiler pipeline.
--  *uint16:* Assigns all background pixels to 0 and each object a different number from
-   1 to 65535 (the maximum value that you can put in a 16-bit integer) and numbers all
-   pixels in each object with the object’s number.  This creates an image where
-   objects in the top left corner of the image are very dark and where the colors progress to white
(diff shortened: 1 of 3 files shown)


DavidStirling:
The primary aim here is to get comparisons between different object sets working.

I've added a plot of the neighbour object set which is shown when objects and neighbours are different. This lets the user see where the neighbours actually are when using different sets.

While doing this we also found that FilterObjects (rightly) doesn't apply filters to objects previously removed for touching image borders, but these hidden objects are still carried forward into the filtered group in some cases, which causes them to be considered inappropriately by this module. To resolve this I've added a setting to disable analysis of border objects.

I've also reworked how border objects are identified. Previously border objects were identified and re-added by taking the unedited segmentation and finding objects which touch the edge. This strategy allows the module to capture objects which were struck down by a size filter during identification (theoretically for being only partial visible), but this may also re-introduce debris and other junk as undesirable neighbour objects.

The new logic takes the small_removed (size-filtered but not border-filtered) object set and compares it to the final segmentation, optionally keeping border-filtered objects if desired. This could miss objects which are only partially visible on the edge of the image and so were removed by a size filter, but avoids including other unwanted border-touching debris. Opinions over which approach is better are welcome!

@braymp - Do you know if this would break any corner cases you can remember from your time here?

braymp:
@DavidStirling: Nope, no cases that come to mind.

DavidStirling:
I've now enabled the Percent Touching measurement when object and neighbour sets aren't identical. This will currently consider the percent touching the perimeter of the cell rather than it's entire area, since the concern is neighbouring objects rather than overlap.

I think there may be a need to update how pipelines from older versions of CP are upgraded, but I'm not sure how it's supposed to work and the tests for this appear to fail on Windows.

DavidStirling:
@0x00b1 @bethac07 - This should be ready for review/merging now!

Issue #3341, opened 2017-10-13
Allow for URLs as argument with --data-file switch


braymp:
If image list CSVs are served up through URLs (e.g., via S3), it would be nice for CP to accept them through the `--data-file` switch.

(If already implemented, then please close)

PR #3877, opened 2019-12-09
download pipeline and/or file list files from online

### LINKED ISSUE #3340: Allow for URLs as argument with -p switch
If pipelines are served up through URLs (e.g., via S3), it would be nice for CP to accept them through the `-p` switch.

(If already implemented, then please close)

bethac07:
Resolves #3340 and #3341 

We may or may not want to actually pull this, since we seem to be solving it differently in master, but this is a fix that would work for 3.1.9 and backwards

### DIFF
--- a/cellprofiler/__main__.py
+++ b/cellprofiler/__main__.py
@@ -33,6 +33,7 @@ import pkg_resources
 import re
 import site
 import tempfile
+import urlparse
 import six.moves
 
 OMERO_CK_HOST = "host"
@@ -69,7 +70,8 @@ def main(args=None):
 
     options, args = parse_args(args)
 
-    if any(options.pipeline_filename.startswith(protocol) for protocol in ('http', 'https', 'ftp')):
+    o = urlparse.urlparse(options.pipeline_filename)
+    if o[0] in ("ftp", "http", "https"):
         import urllib2
         temp_pipe_file = tempfile.NamedTemporaryFile(mode='w+b',suffix='.cppipe')
         downloaded_pipeline = urllib2.urlopen(options.pipeline_filename)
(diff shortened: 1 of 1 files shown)


cellprofiler/__main__.py
```diff
@@ -69,7 +70,8 @@ def main(args=None):
 
     options, args = parse_args(args)
 
-    if any(options.pipeline_filename.startswith(protocol) for protocol in ('http', 'https', 'ftp')):
+    o = urlparse.urlparse(options.pipeline_filename)
```
braymp: Need to add `if options.pipeline_filename:` around lines 73-81, or else NoneType error results.

cellprofiler/__main__.py
```diff
@@ -69,7 +70,8 @@ def main(args=None):
...
+    if o[0] in ("ftp", "http", "https"):
         import urllib2
         temp_pipe_file = tempfile.NamedTemporaryFile(mode='w+b',suffix='.cppipe')
```
braymp: Need to add something like the below before the `temp_pipe_file` assignment, or else an error can occur if the default temp location doesn't exist:
```
if options.temp_dir is not None:
    if not os.path.exists(options.temp_dir):
        os.makedirs(options.temp_dir)
        temp_dir = options.temp_dir
    else:
        temp_dir = cpprefs.get_temporary_directory()
```

cellprofiler/__main__.py
```diff
@@ -69,7 +70,8 @@ def main(args=None):
 
     options, args = parse_args(args)
 
-    if any(options.pipeline_filename.startswith(protocol) for protocol in ('http', 'https', 'ftp')):
+    o = urlparse.urlparse(options.pipeline_filename)
+    if o[0] in ("ftp", "http", "https"):
         import urllib2
```
braymp: Need to add something like the below before the `temp_pipe_file` assignment, or else an error can occur if the default temp location doesn't exist:
```
if options.temp_dir is not None:
    if not os.path.exists(options.temp_dir):
        os.makedirs(options.temp_dir)
        temp_dir = options.temp_dir
    else:
        temp_dir = cpprefs.get_temporary_directory()
```

cellprofiler/__main__.py
```diff
@@ -78,23 +80,27 @@ def main(args=None):
                 a.write(line)
```
braymp: Changed 'with' loop to the following to deal with "IOError: [Errno 13] Permission denied" error; see https://stackoverflow.com/a/23212515
```
for line in downloaded_pipeline:
    temp_pipe_file.write(line)
```

bethac07:
Closing this here in favor of doing this in Nucleus

Issue #3891, opened 2020-01-31
Python 3 - Activating 'Groups' module crashes CellProfiler


DavidStirling:
In the current `master` branch, trying to enable the 'Groups' module triggers a crash with the following error:

```
Traceback (most recent call last):
  File "C:\Program Files\Python37\lib\site-packages\wx\core.py", line 2228, in Notify
    self.notify()
  File "C:\Program Files\Python37\lib\site-packages\wx\core.py", line 3384, in Notify
    self.result = self.callable(*self.args, **self.kwargs)
  File "C:\Users\dstirlin\Github\CellProfiler\CellProfiler\cellprofiler\gui\moduleview.py", line 2436, in do_reset
    self.set_selection(self.__module.module_num)
  File "C:\Users\dstirlin\Github\CellProfiler\CellProfiler\cellprofiler\gui\moduleview.py", line 607, in set_selection
    control = self.make_table_control(v, control)
  File "C:\Users\dstirlin\Github\CellProfiler\CellProfiler\cellprofiler\gui\moduleview.py", line 2175, in make_table_control
    grid.Table.bind_to_grid(grid)
AttributeError: 'GridTableBase' object has no attribute 'bind_to_grid'
```

Issue #3882, opened 2020-01-06
Increment copyright to include 2020


AnneCarpenter:
See prior work! And consider our other, related repos that might need this.

https://github.com/CellProfiler/CellProfiler/search?q=copyright&unscoped_q=copyright
https://github.com/CellProfiler/CellProfiler/pull/3444/files

Issue #3873, opened 2019-11-26
In Test Mode using 'Random Image Set' mistakes composite images for greyscale.


DavidStirling:
I have two-channel composite images that I'd like to split using ColorToGray. In NamesAndTypes I made sure to identify these as Color images.

In test mode, switching image set normally with 'Choose' or 'Next Image Set' works just fine. However, pressing 'Random Image Set' appears to make CellProfiler interpret composite colour images as greyscale. I get the following error when running ColorToGrey. 

![ColorToGrey](https://user-images.githubusercontent.com/26802537/69670340-b3a49000-1061-11ea-9e03-49e8680edf2c.png)

It has no trouble with the same image set when switched to by other methods. I'm attaching a pipeline and test images which can be used to replicate this.

[CompositeBug.zip](https://github.com/CellProfiler/CellProfiler/files/3893940/CompositeBug.zip)

Issue #3870, opened 2019-11-14
Exporting a blank tif image with SaveImages doesn't produce a blank image


DavidStirling:
I ran ConvertObjectsToImage to convert a segmentation into uint16 format. When running SaveImages to export a .tif of this, any image which is blank (no objects) is saved with the entire image being set to half of the maximum possible intensity value rather than 0. The image produced by ConvertObjects is correct and has every pixel set to 0, so something appears to be going wrong when saving.

PR #3894, opened 2020-02-04
Assorted CP4 Fixes

DavidStirling:
Should address #3882, #3873, #3870, #3540 and #3891, along with some other problems.


cellprofiler/modules/metadata.py
```diff
@@ -1164,6 +1177,19 @@ def on_setting_changed(self, setting, pipeline):
                 self.refresh_group_joiner(group)
 
     def update_table(self):
+        for group in self.extraction_methods:
+            if group.extraction_method == X_AUTOMATIC_EXTRACTION and not group.metadata_autoextracted:
+                    response = wx.MessageBox(
```
bethac07: Can you move the wx import down here, so that it only gets imported if it needs to be?

DavidStirling: @bethac07 It was originally also imported by do_update_metadata() as well, so I'd need to restore that too. Wasn't sure if it's better practice to just import it with the module or to load wx each time you hit the update button/do an extraction?

PR #3909, opened 2020-02-18
Disable mousewheel on dropdown menus

### LINKED ISSUE #2363: Dropdown menus are often inappropriately active (on Windows)
As mentioned here: http://forum.cellprofiler.org/t/youtube-tutorials-feedback-thoughts/3935/7

Relevant portion pasted here:

> When I click in the window area away from the variable/parameter fields etc. and then roll the mouse wheel to scroll the main window down (my natural response from years of web browsing) it decides to change the drop down of something I was editing before and causes a mess. 

We haven't found anyone using Macs that reports this but it seems to plague Windows users so we think it's platform-specific.

@bethac07 Anything more specific you can add? I found it hard to understand from the written description.

### LINKED ISSUE #3726: Optionally disable using scrolling to change values in dropdown menus
When I teach people how to use CellProfiler, one of the most common errors they encounter is when they use the mouse scroll wheel to navigate up and down the options (e.g. in NamesAndTypes), but accidentally hovers over a dropdown menu and changes the value instead of scrolling the page without noticing it. This is very confusing to troubleshoot, because in their mind they have never touched the setting that is giving the error (e.g. maybe they accidentally change "Does contain" to "Does not contain" in NamesAndTypes), so they often get stuck here.

The current behavior of the scroll wheel means that if I use it to navigate the options pages, I need to double check every single option on those pages when I encounter an error because I am not sure where I might have hovered over a dropdown box when trying to scroll the page. My own workaround is to not use the scroll wheel at all and drag the scroll bar instead, but the scroll wheel behavior is quite ingrained in people so many still use it by habit.

Is there a way to disable the behavior where the scroll wheel can change the values in the dropdown menus (I could not find one in the preferences), or would you consider adding an option to disable this?

DavidStirling:
Closes #2363. Closes #3726.

This is a common problem seen when training users - using the scroll wheel while hovering over a dropdown menu will change the selected item. Users tend to scroll down module settings panels quickly and accidentally change settings while doing so because of this. This change will ignore mousewheel events over such widgets.

### DIFF
--- a/cellprofiler/gui/moduleview.py
+++ b/cellprofiler/gui/moduleview.py
@@ -857,6 +857,10 @@ class ModuleView(object):
             def callback(event, setting=v, control=control):
                 self.__on_combobox_change(event, setting, control)
 
+            def ignore_mousewheel(event):
+                return
+
+            control.Bind(wx.EVT_MOUSEWHEEL, ignore_mousewheel)
             self.__module_panel.Bind(wx.EVT_COMBOBOX, callback, control)
             if style == wx.CB_DROPDOWN:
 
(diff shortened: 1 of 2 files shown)


bethac07:
Does this prevent the user from using a mousewheel when the menu is actually selected?

DavidStirling:
I had reasoned that people accidentally changing settings was more common than people actually having enough options in a menu to need the scrollwheel. I'll look into whether we could enable mousewheel only when the box is opened.

PR #3914, opened 2020-02-21
ExportToSpreadsheet tries to write into wrong folder

### LINKED ISSUE #3756: ExportToSpreadsheet tries to write into wrong folder (Win 10, CP 3.1.8)
I observed this bug in CP 3.1.8 (binary distribution) on Windows 10:

In `ExportToSpreadsheet` I select the output location `Elsewhere` and click the little folder button to bring up a folder select dialog. In the Dialog I select the folder `C:\Users\Volker\Data\`

The elsewhere gets changed automagically to Default Input-folder subfolder
image
but that still looks correct.

However, during the actual run, Cellprofiler seems to try and write into its own Program Folder, which is not permitted under Windows 10, so I run into this:
image

I have a feeling that it somewhere the default input folders get messed up.

DavidStirling:
Fixes #3756.

For some reason the 'Elsewhere' option when selecting a file was set to fetch the current directory CellProfiler is running from as a base path. This causes problems on Windows as CP's workers don't have permissions to write to those folders.

When a path is generated CP attempts to join a starting directory to whatever the user entered in the subfolder box. Given that 'Elsewhere' should be a full folder path rather than just a subfolder, I've added a little more validation to check that whatever the user entered actually functions as a path on it's own. This still won't fully check write permissions because Windows, but it's at least a sanity check.

I've also changed the default 'starting' directory for 'Elsewhere' from the working directory to the default output folder. This shouldn't technically matter since a full path is required, but if somebody does manage to fool it any written files should at least end up within the output directory instead of in random locations. This default directory will also apply to selecting input files, but that's still better than the install folder. 

Note that javabridge seems to be broken in CP4 at the moment, so I've only been able to test these fixes in 3.1.9 and port them over. I don't anticipate problems though.

### DIFF
--- a/cellprofiler/setting.py
+++ b/cellprofiler/setting.py
@@ -503,10 +503,12 @@ class DirectoryPath(Text):
             and self.custom_path.find(r"\g<") != -1
         ):
             raise ValidationError("Metadata not supported for this setting", self)
-        if self.dir_choice == ABSOLUTE_FOLDER_NAME and (
-            (self.custom_path is None) or (len(self.custom_path) == 0)
-        ):
-            raise ValidationError("Please enter a valid path", self)
+        if self.dir_choice == ABSOLUTE_FOLDER_NAME:
+            if self.custom_path is None or len(self.custom_path) == 0:
+                raise ValidationError("Please enter a path", self)
+            elif os.path.join(self.custom_path, self.custom_path) != os.path.join(self.custom_path):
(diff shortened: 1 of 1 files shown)


DavidStirling:
One foreseeable issue would be if someone wanted to use metadata tags to define their full path rather than just a relative portion. That said, such a scenario should be easy enough to work around.

PR #3918, opened 2020-02-25
Fix RescaleIntensity crashing when told to scale a color image

### LINKED ISSUE #3506: RescaleIntensity failing: The truth value of an array with more than one element is ambiguous
From the forum http://forum.cellprofiler.org/t/value-error-the-truth-value-of-any-array-with-more-than-one-element-is-ambiguous/5692/2
 
I've reproduced it with the stable release and when built from source, error below happens for all three channels. Possibly related to fact that all images have a lot of zero intensity values?

Error message:
Error while processing RescaleIntensity:
The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

Pipeline and Dropbox link to images 
https://github.com/CellProfiler/CellProfiler/files/1741389/Value.error.pipe.cppipe.zip
https://www.dropbox.com/sh/qr0u58wdkiiqsi3/AAAF79lotBTR7lflgfchkQ3ja?dl=0

DavidStirling:
Fixes #3506.

I considered just adding a warning when trying to use a color image with this module, but it turns out that all the modes aside from 'stretch intensity' already worked just fine with color images. You might also imagine a scenario where a user wants to rescale their channels before splitting them.

I've made it so that if the input image has multiple colour channels the 'stretch intensity' mode will be applied to each channel individually. An alternative solution might be to have max intensity across all channels considered as the stretching 'target'. Either is better than the module just breaking when you do this.

I also noticed that the output display for RescaleIntensity was the default viewer. Since this includes intensity normalisation prior to display it seems to make it difficult to see the actual results of the module (it gets rescaled again for display). That's especially problematic for multi-colour images. I've added a dedicated display method which disables the normalisation. I'm not sure if we'll want this or not, since while it'll visualise rescaling of low values better the plots would look washed out if the resulting values exceed 1.0. - Thoughts welcome!

### DIFF
--- a/cellprofiler/modules/rescaleintensity.py
+++ b/cellprofiler/modules/rescaleintensity.py
@@ -13,6 +13,11 @@ texture measurements derived from images that have been rescaled because
 certain options for this module do not preserve the relative intensities
 from image to image.
 
+As this module rescales data it will not attempt to normalize displayed previews
+(as this could appear to undo the scaling). As a result images rescaled
+to large ranges may appear washed out. To normalize values for viewing,
+right-click an image and choose an image contrast transform.
+
 |
 
 ============ ============ ===============
--- a/cellprofiler/modules/rescaleintensity.py
+++ b/cellprofiler/modules/rescaleintensity.py
@@ -14,8 +14,8 @@ certain options for this module do not preserve the relative intensities
 from image to image.
 
 As this module rescales data it will not attempt to normalize displayed previews
-(as this could appear to undo the scaling). As a result images rescaled
-to large ranges may appear washed out. To normalize values for viewing,
+(as this could make it appear that the scaling had done nothing). As a result images rescaled
+to large ranges may appear dim after scaling. To normalize values for viewing,
 right-click an image and choose an image contrast transform.
 
 |
--- a/cellprofiler/modules/rescaleintensity.py
+++ b/cellprofiler/modules/rescaleintensity.py
@@ -439,11 +439,25 @@ Select the measurement value to use as the divisor for the final image.
 
     def stretch(self, input_image):
         data = input_image.pixel_data
-
         mask = input_image.mask
 
-        in_range = (min(data[mask]), max(data[mask]))
+        if len(data.shape) == len(mask.shape) + 1:
+            #Image is multi-channel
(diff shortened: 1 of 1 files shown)


cellprofiler/modules/rescaleintensity.py
```diff
@@ -439,11 +467,25 @@ def rescale(self, image, in_range, out_range=(0.0, 1.0)):
 
     def stretch(self, input_image):
         data = input_image.pixel_data
-
         mask = input_image.mask
 
-        in_range = (min(data[mask]), max(data[mask]))
+        if len(data.shape) == len(mask.shape) + 1:
```
bethac07: See if you can replace with ndim

cellprofiler/modules/rescaleintensity.py
```diff
@@ -439,11 +467,25 @@ def rescale(self, image, in_range, out_range=(0.0, 1.0)):
...
+            #Image is multi-channel
+            splitaxis = len(data.shape)-1
```
bethac07: Is there also a "channels" thing in the image class?

cellprofiler/modules/rescaleintensity.py
bethac07: https://github.com/CellProfiler/CellProfiler/blob/master/cellprofiler/image.py#L105-L106

DavidStirling:
I think that's ready now.

cellprofiler/modules/rescaleintensity.py
```diff
@@ -13,6 +13,11 @@
 certain options for this module do not preserve the relative intensities
 from image to image.
 
+As this module rescales data it will not attempt to normalize displayed previews
+(as this could appear to undo the scaling). As a result images rescaled
```
bethac07: "as this could appear that the scaling had done nothing"

cellprofiler/modules/rescaleintensity.py
```diff
@@ -13,6 +13,11 @@
...
+to large ranges may appear washed out. To normalize values for viewing,
```
bethac07: "may appear dim after scaling"

DavidStirling:
@bethac07 Applied those changes, thanks.

Issue #1701, opened 2015-11-17
Dragging modules in pipeline window creates a copy rather than moves


dlogan:
They used to move and not copy, which seems more intuitive.

braymp:
Definitely not the case in Windows.

AnneCarpenter:
I confirm this behavior in the April 18 beta on my Mac 10.10.5

AnneCarpenter:
Confirmed it's still the case in 3.0.0rc3

DavidStirling:
Can confirm this is still a Mac-specific issue as of 3.1.9, Windows is unaffected.

bethac07:
I am reasonably sure we looked it up at one point this is a wx3 bug, not a CellProfiler bug.  Leaving open until we confirm it's fixed by wx4.

PR #3922, opened 2020-02-26
Prevent module duplication when dragging on mac

DavidStirling:
It seems that on mac dragging a module returns a wx DragCopy object rather than DragMove. On Windows this can be triggered by holding control when dragging, which does sometimes happen when selecting and moving multiple modules quickly.

This will disable module duplication by ctrl+drag, but fixes moving modules by dragging on mac. If ctrl+drag duplication was precious perhaps we could try making this OS-specific?


DavidStirling:
Okay, this should be ready now. Fixes #1701

PR #3935, opened 2020-03-03
Fix Random Image Set, add Random Image Group

### LINKED ISSUE #2549: Random image set selection fails when 1 image set per group
To replicate:
* Load the welcome screen pipeline
* Enable metadata extraction for "Extract from file folder names", set regex as <code>^(?P<Plate>.\*)\_POS(?P<Site>[0-9]\*)_</code> and hit the "Update metadata" button
* Enable Groups module and set metadata tag for grouping as "Plate" and "Site" (so there's only 1 image set per group)
* Enter test mode and select "Random image set" to yield the following error:
```
  File "cellprofiler\gui\pipelinecontroller.pyc", line 2969, in on_debug_random_image_set
  File "mtrand.pyx", line 941, in mtrand.RandomState.randint (numpy\random\mtrand\mtrand.c:10717)
ValueError: low >= high
```

DavidStirling:
Fixes #2549. Related to #3873.

The logic in the original version of these commands was a bit weird.

It attempted to select a random image group but never actually set that group up to run, so random numbers were generated based on the length of a random group what wasn't running. I expect this is why the % operation was needed - so that if the proposed index was too high you'd still get a valid result, but this would then bias you towards low index numbers.

It also seems that in modern numpy specifying size=1 in numpy randomint forces it to return an array rather than a single int.

To resolve this I've revised the randomisation approach and split random into two functions - Random Image Set and Random Image Group. These use separate menu buttons. Random image set will now only randomise within the current group (which it did anyway but probably wasn't intended to). I imagine that keeping them separate gives the user more control than having them as a single function, since you can always click both and "I want a random field from only this group" is a common scenario.

### DIFF
--- a/cellprofiler/gui/cpframe.py
+++ b/cellprofiler/gui/cpframe.py
@@ -125,6 +125,7 @@ ID_DEBUG_NEXT_GROUP = wx.NewId()
 ID_DEBUG_CHOOSE_GROUP = wx.NewId()
 ID_DEBUG_CHOOSE_IMAGE_SET = wx.NewId()
 ID_DEBUG_CHOOSE_RANDOM_IMAGE_SET = wx.NewId()
+ID_DEBUG_CHOOSE_RANDOM_IMAGE_GROUP = wx.NewId()
 ID_DEBUG_RELOAD = wx.NewId()
 ID_DEBUG_PDB = wx.NewId()
 ID_DEBUG_RUN_FROM_THIS_MODULE = wx.NewId()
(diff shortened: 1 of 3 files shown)


DavidStirling:
I've added 4959c36 as a sorta-fix for another issue. Currently the groups module's tables will only update once upon changing which metadata is to be used for group definition. After that the table fails to refresh. For some reason after the first run of on_setting_changed the workspace seems to lose it's record of which image sets exist and I can't figure out why. As in theory the image set list cannot be changed without leaving the module and triggering the on_activated function (which resets everything) I don't think this will cause any problems, but it is an imperfect solution.

bethac07:
Can you elaborate a bit more on this last commit? Is this a thing that only happens in master? I'm not quite following the explanation, sorry.

DavidStirling:
It happens in 3.1.9 and Master under the right conditions. To replicate this:

- Click on a Groups tab and activate grouping.
- Change metadata category --> The tables and groups displayed update to reflect the new category.
- Change metadata category again --> The tables and groups fail to update. The previously selected category is still displayed.

Note that in 3.1.9 the tables fail to update when turning on metadata in the first place, at least on my system. You have to view another module then return to see any groupings but then you can replicate this. 

This is happening because somewhere after the first table update CellProfiler is losing track of how many image sets there were, and if the variable storing image set count is empty the Setting Changed function returns without updating tables. Since the actual number of image sets can't really change while inside the metadata module I've just made it refresh the table regardless.

Hope that's clearer.

bethac07:
Sounds fair.

PR #3937, opened 2020-03-05
Fix "Open A New CP Window"

### LINKED ISSUE #3334: Failure to open new CP window from source and distribution
File > Open new CP window , gives error:
```
Couldn't get a file descriptor referring to the console
```
Ubuntu 16.04
CP3.0.0rc7

### LINKED ISSUE #3359: "Open new CP window" when having both 2.2 and 3.0
When you have both CP 2.2 and 3.0 installed, "Open new CP window" while running CP3.0 will launch CP 2.2 instead of 3.0

### LINKED ISSUE #3735: Cellprofiler uses the shell command "open" which can be missing in Linux
If someone wants to start a new CellProfiler instance via `File -> Open a New CP Window` it invokes the follwing code: 

https://github.com/CellProfiler/CellProfiler/blob/f3d77528cd39a4543f6c1dc99dd0fa77d81ad769/cellprofiler/gui/cpframe.py#L950-L957

This can lead to an error in Linux-Distributions, where the command `open` does not exists. 

Currently i am new to Cellprofilier and by exploring it, i found this Issue. I am not sure whether this function is also used by Mac or Windows.

I am using this commit f3d77528cd39a4543f6c1dc99dd0fa77d81ad769 and have 
executed `pip2.7 install --user -e .` with no errors and am starting it via `~/.local/bin/cellprofiler`

Here is the specific Output by trying to open a new CP-Window: 

```bash
sh: open: command not found
```

Also nothing happens afterwards.

---
Are there also other locations, where such calls (using the shell) are executed which may also be OS-dependent and/or executables might missing? As i can see, this is the only code-snippet where `os.system` is referenced. Is there another way to open a new instance in a 'more python way'?

I my case:
The program to 'open' things via bash is called `xdg-open`. 

Some Desktop-Environmets might also use an alias `open` or a symlink `/bin/open` which actually references `xdg-open` or `kde-open`, etc... (Even in Ubuntu).

---

Besides of that, if i replace `open` via `xdg-open` i get another error:
```shell
xdg-open: file 'CellProfiler_python.command' does not exist
```

But this might get resolved if we are able to solve this in a 'more python way'.

Greetings!

DavidStirling:
Closes #3334, Closes #3359 and Closes #3735.

By somewhat popular demand, this menu button should now work. I've made it OSX-specific as opposed to OSX and Linux.

I've also moved the menu button to a more sensible spot rather than having it between the two pipeline running options.

### DIFF
--- a/cellprofiler/gui/cpframe.py
+++ b/cellprofiler/gui/cpframe.py
@@ -658,14 +658,15 @@ class CPFrame(wx.Frame):
         self.__menu_file.Append(
             ID_FILE_RUN_MULTIPLE_PIPELINES, "Run Multiple Pipelines"
         )
-        if os.name == "posix":
-            self.__menu_file.Append(ID_FILE_NEW_CP, "Open a New CP Window")
         self.__menu_file.Append(
             ID_FILE_RESTART,
             "Resume Pipeline",
             "Resume a pipeline from a saved measurements file.",
         )
         self.__menu_file.AppendSeparator()
+        if sys.platform == "darwin":
+            self.__menu_file.Append(ID_FILE_NEW_CP, "Open A New CP Window")
+            self.__menu_file.AppendSeparator()
         self.__menu_file.Append(
             ID_OPTIONS_PREFERENCES,
             "&Preferences...",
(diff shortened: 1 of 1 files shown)


DavidStirling:
@bethac07 Would you be able to test this?

cellprofiler/gui/cpframe.py
```diff
@@ -1113,12 +1114,10 @@ def __on_close_all(self, event):
...
     def __on_new_cp(event):
-        import os
-
-        if not hasattr(sys, "frozen"):
-            os.system("open CellProfiler_python.command")
+        if hasattr(sys, "frozen"):
+            os.system("open -na /Applications/CellProfiler-{}.app".format(cellprofiler.__version__))
         else:
-            os.system("open -na CellProfiler.app")
+            os.system("cellprofiler")
```
bethac07: In case the person has multiple Pythons and/or CellProfiler's on their system, let's make the command here `python3 -m cellprofiler`

bethac07:
Can't test built until it's built, but let's pull it once you make the requested change.

DavidStirling:
Made that change, should be ready now.

PR #3938, opened 2020-03-09
[WIP] Revised GUI for selecting multiple image or object sets

### LINKED ISSUE #2298: Change "Add another image/object" dropdowns to MultiChoice windows
Beth @bethac07 had a lovely usability improvement idea that would not be rocket science to implement. 

We were lamenting the many modules that offer "Add another image" and "Add another object"... the buttons themselves are pretty chunky and there are not obvious visual cues for each set of settings that relate to each image/object so overall this gets pretty ugly especially when you have multiple images and objects specified in the same module. We are guessing the code that enables this behavior is complex, too.

She suggested switching to `cps.MultiChoice`. An example is in MeasureTexture ("Angles to measure"). This way, there is a single chunk of real estate devoted to choosing images, and you can readily choose all/some images by command-clicking.

In our case, it would be nice for cps.MultiChoice to expand to fit its contents up to say ~10 or so (given a pipeline will typically have a manageable number of images/objects; it's weird to interact with the cps.MultiChoice element when you can only see 3 items in the tiny window).

Current situation (with two images chosen so far):
addimage

Proposed look (from MeasureTexture, with 2 items chosen):
multichoice

(Caution: this may not be feasible for all modules/situations; for example if there are a series of subsequent unique choices to be made for each image/object. Thus the first step towards accomplishing this goal is to itemize which cases can be handled this way.)

DavidStirling:
Closes #2298.

As discussed a long time ago, a common frustration is having to use "add another image" buttons to individually specify images in the measurement modules. This becomes tedious when working with large numbers of image and object sets.

I've been working on implementing checklist boxes as an easier way to handle this. It should be appropriate wherever adding an image or object set doesn't require any additional per-group settings. I think the checklists are a bit easier for users to understand and handle than the normal multiple selection lists.

I've currently added this to the MeasureObjectIntensity module (see below). If we want to go ahead with this I can start adding it to other modules.

NewMultiSubscriberHandling

Oh, you can also right click for select all/none options. Because checking a few dozen boxes is still tedious.

There are probably still bugs to be found and better solutions for some aspects, so this is a WIP for now.

### DIFF
--- a/cellprofiler/modules/measureobjectintensity.py
+++ b/cellprofiler/modules/measureobjectintensity.py
@@ -296,32 +296,6 @@ Select the objects whose intensities you want to measure.""",
             setting_values[1] = setting_values[1].replace('\'', '').strip('][').split(', ')
         return setting_values, variable_revision_number, from_matlab
 
-    def prepare_settings(self, setting_values):
-        """Do any sort of adjustment to the settings required for the given values
-
-        setting_values - the values for the settings
-
-        This method allows a module to specialize itself according to
-        the number of settings and their value. For instance, a module that
-        takes a variable number of images or objects can increase or decrease
-        the number of relevant settings so they map correctly to the values.
-
-        See cellprofiler.modules.measureobjectsizeshape for an example.
-        """
-        #
-        # The settings have two parts - images, then objects
-        # The parts are divided by the string, cps.DO_NOT_USE
-        #
-        return
-        image_count = int(setting_values[0])
-        object_count = len(setting_values) - image_count - 1
-        del self.images[image_count:]
-        while len(self.images) < image_count:
-            self.add_image()
-        del self.objects[object_count:]
-        while len(self.objects) < object_count:
-            self.add_object()
-
     def validate_module(self, pipeline):
         """Make sure chosen objects and images are selected only once"""
         images = set()
@@ -343,8 +317,8 @@ Select the objects whose intensities you want to measure.""",
     def get_measurement_columns(self, pipeline):
         """Return the column definitions for measurements made by this module"""
         columns = []
-        for image_name in [im.name for im in self.images]:
-            for object_name in [obj.name for obj in self.objects]:
+        for image_name in self.images_list.value:
+            for object_name in self.objects_list.value:
                 for category, features in (
                     (INTENSITY, ALL_MEASUREMENTS),
                     (cellprofiler.measurement.C_LOCATION, ALL_LOCATION_MEASUREMENTS),
@@ -352,8 +326,8 @@ Select the objects whose intensities you want to measure.""",
                     for feature in features:
                         columns.append(
                             (
-                                object_name.value,
-                                "%s_%s_%s" % (category, feature, image_name.value),
+                                object_name,
+                                "%s_%s_%s" % (category, feature, image_name),
                                 cellprofiler.measurement.COLTYPE_FLOAT,
                             )
                         )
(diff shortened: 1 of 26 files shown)


cellprofiler/modules/measureobjectintensity.py
```diff
@@ -266,65 +201,59 @@ def upgrade_settings(
...
             variable_revision_number = 3
+        if variable_revision_number == 3:
+            assert not from_matlab
+            num_imgs = int(setting_values[0])
+            images_list = setting_values[1:num_imgs+1]
+            objects_list = setting_values[num_imgs+1:]
+            setting_values = [images_list, objects_list]
+            variable_revision_number = 4
+        else:
+            # Convert saved image and object lists back into python lists
```
DavidStirling: @0x00b1 @bethac07 - I understand that this is a hacky way of getting a CellProfiler setting to store/recall a list rather than a string. Is there a better way to do this? I expect I'm missing something obvious.

AnneCarpenter:
I LOVE this UI! The only downside, which we discussed, is that if you pick a channel you are choosing it for ALL objects (and vice versa) whereas in the old one you could choose more granularly. However, this UI is more convenient for the vast majority of cases, the downside of measuring more than you want is just a bit of compute time ,and if you really want to work around this you can always add two separate copies of the module. So it seems like almost no downside at all! 

Once this settles and is adopted, we need to ensure that old pipelines are imported and translated to this properly (handling the weird case where someone only measured green in nuclei and red in cells, for example - we'd change that to measuring both in both).

DavidStirling:
So for MeasureObjectIntensity there isn't an option to specify different object/channel setting groups, anything selected gets analysed together (in 3.1.9 at least). Due to that this change has no impact on functionality - you'd need a second copy of the module to restrict analysis for particular objects anyway.

Running through the other modules and how we might apply this:

The same applies to MeasureObjectSizeShape, MeasureColocalization and MeasureTexture. We can update those with no loss of function.

For other modules we might just want to add checklists wherever it seems appropriate and leave the original functionality where necessary. MeasureObjectIntensityDistribution has specific "where to measure from" parameters for each object, but we could update selection of images. MeasureGranularity is similar but with images, as is MeasureImageIntensity. 

In other words, we could use single image selectors wherever individual parameters are needed, but include checklists wherever this might simplify workflow without limiting functionality. That should make importing old pipelines straightforward too.

AnneCarpenter:
I am very much in favor of making them all identical, where you cannot
select complicated subsets of channels and objects even for those
apparently-fewer-than-I-thought modules where it used to be possible.

DavidStirling:
For some modules it might be possible to have the parameters be specific to all images/objects selected, then rework it towards "add another analysis group" to allow additional sets of settings with different objects. You might then wonder at what point it'd be better to just add a new copy of the module though. I think the bigger obstacle to this is making it compatible with old pipelines. It's relatively simple to just convert sets of objects into a list, but what do we then do with subset groups if we're not supporting that?

For now perhaps I'll experiment on a per-module basis to see what I can come up with.

AnneCarpenter:
If an old pipeline has subset groups, you'd just switch them over to doing
the full channel x object sets they've specified and pop a warning saying
you've done so and they will be measuring a bit more than they used to with
their pipeline, and if they do not like it they can add a separate module
to measure individual combinations. I really think it's not widespread and
not a big downside, so don't worry about it!

DavidStirling:
Ack, I've messed up rebasing this. Will draft a fresh branch/pull request when this is ready. Sorry!

PR #3942, opened 2020-03-13
Add more descriptive errors to modules requiring matched image dimensions

### LINKED ISSUE #1977: A request for more descriptive errors when image dimensions disagree.
Currently, if the dimensions of images disagree Cellprofiler will throw an error that references numpy, but does not explicitly indicate what is wrong. The troubleshooting would be much easier if the error contained a message to the effect of "The source images must have the same width and height."

DavidStirling:
Closes #1977.

I've gone through and added checks for whether image dimensions are compatible to modules where this wasn't already handled elsewhere.

In the case of IDTertiary, the module actually has some methods for handling when initial object sets are larger than the smaller objects, but this isn't something the user should be doing intentionally. I've therefore made the error simply ask that the dimensions be matched if this process was unsuccessful.

### DIFF
--- a/cellprofiler/modules/correctilluminationapply.py
+++ b/cellprofiler/modules/correctilluminationapply.py
@@ -240,7 +240,7 @@ somewhat empirical.
             raise ValueError(
                 "This module requires that the image and illumination function have equal dimensions.\n"
                 "The %s image and %s illumination function do not (%s vs %s).\n"
-                "If they are paired correctly you may want to use the Resize or Crop module to make them the same size.
+                "If they are paired correctly you may want to use the Resize or Crop module to make them the same size."
                 % (
                     image_name,
                     illum_correct_name,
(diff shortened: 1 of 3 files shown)


bethac07:
I love the idea here, but in addition can we add pointers to the correct modules (Resize and/or ResizeObjects) to fix the problem?

DavidStirling:
We can do that, but resizing the images may not necessarily be the correct fix depending on the situation - like if the image was cropped. Perhaps we should mention it in the module help?

bethac07:
Fair point! You could mention Crop as well- I leave it to your best judgement how to best  clean up a version of the below.

"Your two image/object sets are not the same size; you may make them the same size using the Crop module, or stretch them to the same size with Resize/ResizeObjects module(s)".

DavidStirling:
Okay, I've added that content to the warning messages. Hopefully that's clear.

Issue #3857, opened 2019-10-25
Preference changes aren't persisted


MaximLippeveld:
Hi all,

I stumbled on a problem (on linux) where preference changes made in the GUI preference dialog weren't being persisted.

After some digging I found that the changes are only flushed to disk when the program is ended. The simple fix is to flush to disk after each preference change. I will submit a PR later with this fix!

bethac07:
Thanks for this!

PR #3943, opened 2020-03-13
Relabel "OK" button in preferences GUI

DavidStirling:
Related to #3857.

I'm not sure if anyone has verified the original issue on Linux or if it's already fixed, but on Windows there are no problems with saving. However, the current GUI does not make it obvious that preferences changes will only save when exiting the dialog with the "OK" button. I know this has tripped me up in the past. To address this I've just changed the text on that button to "Save", which will hopefully be more intuitive.

Could someone check that this works fine on Mac? I believe there's a chance that the button order may be reversed with "OK" on the right, in which case I'll add a less lazy solution.

### DIFF
--- a/cellprofiler/gui/preferencesdlg.py
+++ b/cellprofiler/gui/preferencesdlg.py
@@ -219,10 +219,13 @@ class PreferencesDlg(wx.Dialog):
         sizer.AddGrowableCol(3, 1)
 
         top_sizer.Add(wx.StaticLine(scrollpanel), 0, wx.EXPAND | wx.ALL, 2)
-        btnsizer = self.CreateButtonSizer(wx.OK | wx.CANCEL)
-        btnsizer.Children[1].Window.Label = "Save"
-        self.Bind(wx.EVT_BUTTON, self.save_preferences, id=wx.ID_OK)
-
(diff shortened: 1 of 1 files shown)


bethac07:
```
  File "/Users/bcimini/Documents/GitHub/CellProfiler/CellProfiler/cellprofiler/gui/cpframe.py", line 1108, in __on_preferences
    dlg = cellprofiler.gui.preferencesdlg.PreferencesDlg()
  File "/Users/bcimini/Documents/GitHub/CellProfiler/CellProfiler/cellprofiler/gui/preferencesdlg.py", line 223, in __init__
    btnsizer.Children[1].Window.Label = "Save"
AttributeError: 'NoneType' object has no attribute 'Label'
```

DavidStirling:
@bethac07 - Could you try this new version?

bethac07:
Save is indeed on the right now, with cancel on the left, but each button works as expected. 
image

LMK if you want to change the button order (I'm fine either way), or if you want me to pull.

DavidStirling:
Ah good, that's the inbuilt platform adjustment - on Windows the buttons are the other way around. If the save button works correctly then that should be fine to pull. Thanks!

PR #3950, opened 2020-03-17
Add NaN substitution to ImageMath

### LINKED ISSUE #3947: Error while processing Threshold
Thresholding after ImageMath (configuration below) resulted this error:
Screen Shot 2020-03-16 at 9 15 27 PM

ImageMath: 
Operation: Subtract
Raise to the power: **0.4**   (**when 1 instead of 0.4, it runs with no error**) 

Threshold:
Strategy: Adaptive 
method: Otsu (3class, background)

DavidStirling:
Fixes #3947. User can now tell ImageMath to replace nan with 0.


bethac07:
I think that's record time, thanks dude!

Issue #3962, opened 2020-03-26
Text wrapping in module view does not work consistently.


DavidStirling:
Present in both 3.1.9 and Master. For some reason wx isn't wrapping text properly when viewing a module after displaying the Images, NamesAndTypes or an activated Metadata module. Any refresh (e.g. changing a setting or viewing the module again) immediately corrects this.

This is more noticeable with the new measurement modules, but applies to most modules which have settings with long text.

To replicate:
- Make a new pipeline.
- Add the UntangleWorms module (or anything else with long setting name text).
- Adjust window size so that text should wrap.
- View the Images module.
- View the UntangleWorms module, text will not be wrapped.
- Click again on UntangleWorms, wrapping corrects itself.

I'm not sure why this would only be happening when coming from certain modules, but I anticipate it might have something to do with the wide table and file list fields.

PR #3967, opened 2020-03-30
New GUI for selecting multiple image and object sets

### LINKED ISSUE #2298: Change "Add another image/object" dropdowns to MultiChoice windows

DavidStirling:
Continuing from #3938 but with a cleaner branch. Closes #2298.

This pull request will replace the interface for selecting multiple image or object sets in the measurement modules. Previously multiple objects were typically added using groups, with an 'Add an image' button allowing the user to choose an additional image from a dropdown box. This was slow and tedious when working with many images. The new implementation uses checklist boxes. I've also added a right-click menu for 'select all/none' functionality. To preserve functionality items are also labelled with their source module, and annotated appropriately if they're missing from the pipeline (screenshot below).

NewMultiSubscriberHandling

Per Anne's comments, I've tried to make as much use of this as possible to keep things consistent between modules. Changed modules are as follows:

- MeasureColocalization
- MeasureGranularity
- MeasureImageAreaOccupied
- MeasureImageIntensity
- MeasureImageQuality
- MeasureObjectIntensity
- MeasureObjectIntensityDistribution
- MeasureObjectSizeShape
- MeasureTexture

MeasureObjectIntensityDistribution still uses groups for object sets since each needs an individually defined center point.

**Compatibility notes**:

I've made it so that selected image and object sets are converted when importing older pipelines. However, with MeasureGranularity we decided to move from having detection parameters configured per-image to having a single set of parameters shared between all images. In practice most users did this anyway, but when importing old pipelines where multiple sets of parameters were defined CellProfiler will now carry forward the first set and log a warning about the upgrade. Multiple copies of a module can be used to specify multiple sets of analysis parameters. 

Likewise, MeasureImageIntensity now no longer operates on pairs of images and objects, instead all possible image/object pairings will be processed together. Upgrading an old pipeline will build a list featuring all images and objects and log a warning about this change. This could create more computational work, but again a user could duplicate the module to do specific pairwise comparisons.

**Other notes**:

This implementation adds the ListImageNameSubscriber and ListObjectNameSubscriber setting classes to setting.py. This is basically a wx listbox but with checkboxes next to each name to aid selection, which seems to be much easier to use than SubscriberMultiChoice (which has existed for a long time but is rarely utilised, and needed ctrl+clicking to select multiple options). SubscriberMultiChoice is still used in ExportToDatabase but has otherwise been replaced, if we want to remove that class entirely then we could update ExportToDatabase too with a bit of effort.

It's worth mentioning that the large boxes introduced by this change will make the text wrapping bug described in #3962 more noticeable. We might want to try to fix that.

### DIFF
--- a/cellprofiler/gui/moduleview.py
+++ b/cellprofiler/gui/moduleview.py
@@ -2267,7 +2267,7 @@ class ModuleView(object):
     def __on_checklistbox_change(self, event, setting, control):
         if not self.__handle_change:
             return
-        self.on_value_change(setting, control, control.GetChecked(), event)
+        self.on_value_change(setting, control, control.GetChecked(), event, timeout=CHECK_TIMEOUT_SEC * 1000)
 
     def __on_multichoice_change(self, event, setting, control):
         if not self.__handle_change:
--- a/cellprofiler/gui/moduleview.py
+++ b/cellprofiler/gui/moduleview.py
@@ -516,7 +516,8 @@ class ModuleView(object):
                         v, v.get_choices(), control_name, wx.CB_READONLY, control
                     )
                     flag = wx.ALIGN_LEFT
-                elif isinstance(v, cellprofiler.setting.ListImageNameSubscriber):
+                elif isinstance(v, (cellprofiler.setting.ListImageNameSubscriber,
+                                   cellprofiler.setting.ListObjectNameSubscriber)):
                     choices = v.get_choices(self.__pipeline)
                     control = self.make_list_name_subscriber_control(
                         v, choices, control_name, control
@@ -835,7 +836,7 @@ class ModuleView(object):
         return control
 
     def make_list_name_subscriber_control(self, v, choices, control_name, control):
-        """Make a read-only combobox with extra feedback about source modules,
+        """Make a read-only combnobox with extra feedback about source modules,
         and a context menu with choices navigable by module name.
 
         v            - the setting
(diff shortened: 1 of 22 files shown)


DavidStirling:
Point of interest - I know the font spacing between items and their source modules is not perfect. This is because wx's default fonts are not fixed pitch. I can make it align more nicely, but the result is a less attractive font (see screenshot below). I'm open to either option but went with the default fonts because I feel it's easier to read and you can fit a much longer name without needing scrollbars.

ListboxAlt

bethac07:
Spacing, etc looks really bad on Mac; I don't think we can/should pull this without fixing that 
image

bethac07:
(Sorry! Otherwise I'm super in favor)

DavidStirling:
Wow, thanks for checking that! That's really weird, it looks like on mac something is adding spacing between the checkbox and the actual text.

PR #3969, opened 2020-03-30
Fix urlparse in Python 3

DavidStirling:
urlparse was renamed to urllib.parse in Python 3, so imports from #3958 need to be fixed


bethac07:
Awesome, thanks!

PR #3971, opened 2020-03-31
Add search filter when adding modules

### LINKED ISSUE #3964: Make the Add Modules dialog "Search" better
It's currently not clear that the 'Search' button on the 'Add Modules' dialog will search for all mentions of the keyword within module help sections. When used this is pretty slow, but more importantly it's not immediately obvious that this is what the feature is intended for. As I found when first using CellProfiler, users expect that this box will search for a particular module by name.

The current implementation pretty much just launches the dialog from Help-->Search Help (from #467). I would propose replacing the Add Modules search with a simpler function which, when typed in, swaps Module Category to 'All' and filters the displayed list down to just modules with names matching the term. The idea being that if I want "CalculateMath" I can just whack "Calc" into the box and immediately see the module I want to add. I'd certainly find this much more useful, but I'm keen to hear others' opinions on this.

DavidStirling:
Closes #3964.

This pull request changes the 'Add Modules' panel so that typing in the search box automatically displays a filtered list of modules which have names matching the entered text (case-insensitive for convenience). By default the first match is highlighted, and I've made it so that hitting 'enter' will add that selected module. This means that you can now add specific modules using just the keyboard e.g. Type "water" and hit enter and you'll add the Watershed module to your pipeline.

To preserve the old functionality, I've left the dedicated 'search' button in place. Pressing this button will still grab whatever was in the text box and search all help for mentions of the term.

Edit: Attaching a screenshot - 

AddModules

### DIFF
--- a/cellprofiler/gui/addmoduleframe.py
+++ b/cellprofiler/gui/addmoduleframe.py
@@ -59,13 +59,13 @@ class AddModuleFrame(wx.Frame):
         search_sizer = wx.BoxSizer(wx.HORIZONTAL)
         self.GetSizer().Add(search_sizer, 0, wx.EXPAND | wx.ALL, 2)
         search_sizer.Add(
-            wx.StaticText(self, label="Search:"),
+            wx.StaticText(self, label="Find Modules:"),
             0,
             wx.ALIGN_LEFT | wx.ALIGN_CENTER_VERTICAL,
         )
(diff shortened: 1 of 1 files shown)


DavidStirling:
@bethac07 - Hopefully this solution will give us the best of both worlds!

PR #3977, opened 2020-04-02
Fixes for export modules

### LINKED ISSUE #3739: Add Image File and Folder Names to Object Table
Hello,

It would be great if the file and folder names of both  input images as well as of any saved images (such as segmentation overlays) could be added to an objects output table. 

The use case is that one could, in a downstream analysis, easily see from which images the object measurements are derived (and also check the corresponding object segmentations).

See also the discussion here:
https://forum.image.sc/t/cellprofiler-combined-image-and-objects-table/25053/6

### LINKED ISSUE #3973: ExportToSpreadsheet writes image thumbnails if ExportToDatabase also present
If a user is using both ExportToSpreadsheet and ExportToDatabase (which is an edge case but not a corner one), don't have selection of particular measurements turned on in ExportToSpreadsheet AND if they are writing image thumbnails to the database in ExportToDatabse, those `Thumbnail_ImageName` columns ALSO get exported to the spreadsheet, which if the image is large makes the spreadsheet choke pretty hard.  This happens even if ExportToDatabase is after ExportToSpreadsheet.

Ideally, we'd like to avoid this behavior, though it's probably not the end of the world if we can't since we can always tell the user to turn on measurement selection in ExportToSpreadsheet but then just select everything, which I've validated avoids the behavior.

Pipeline to validate below (for use with the welcome screen images)

[thumbnail_in_spreadsheet.cppipe.zip](https://github.com/CellProfiler/CellProfiler/files/4415665/thumbnail_in_spreadsheet.cppipe.zip)

### LINKED ISSUE #3976: ExportToDatabase crashes if trying to write thumbnails of binary images
Per [this forum post](https://forum.image.sc/t/error-while-processing-exporttodatabase-message/35686), the logic for writing thumbnails for binary images is broken in ExportToDatabase. It produces the following error in 3.1.9:

`TypeError: numpy boolean subtract, the '-' operator, is deprecated, use the bitwise_xor, the '^' operator, or the logical_xor function instead.`

Which arises here:

https://github.com/CellProfiler/CellProfiler/blob/5bb5966cde8215302a853599e266a123991ac217/cellprofiler/modules/exporttodatabase.py#L2555-L2562

It should be relatively straightforward to adapt this to handle binary images properly. I'm thinking we just make the rescaling conditional on it not being a boolean image.

DavidStirling:
I've made it so that ExportToDatabase will no longer try (and fail) to rescale boolean thumbnails. Fixes #3976

On the subject of thumbnails, I've made ExportToSpreadsheet exclude thumbnails generated by ExportToDatabase, unless the user is in 'choose columns' mode - in which case they can have them if they so wish. Fixes #3973

I've also added the requested 'add file folder and name in objects table' option to ExportToSpreadsheet. Closes #3739

These are a bit tricky to test on the part of Master's analyse mode being broken, but I've got them working on 3.1.9 and hopefully the port should be fine.

### DIFF
--- a/cellprofiler/modules/exporttodatabase.py
+++ b/cellprofiler/modules/exporttodatabase.py
@@ -2557,7 +2557,7 @@ available:
                     or pixels.dtype == numpy.bool
                 ):
                     factor = 255
-                    if self.auto_scale_thumbnail_intensities:
+                    if self.auto_scale_thumbnail_intensities and pixels.dtype != numpy.bool:
                         pixels = (pixels - pixels.min()) / pixels.max()
                 else:
                     raise Exception(
--- a/cellprofiler/modules/exporttospreadsheet.py
+++ b/cellprofiler/modules/exporttospreadsheet.py
@@ -135,7 +135,7 @@ SETTING_OG_OFFSET_V9 = 15
 SETTING_OG_OFFSET_V10 = 17
 SETTING_OG_OFFSET_V11 = 18
 """Offset of the first object group in the settings"""
-SETTING_OG_OFFSET = 17
+SETTING_OG_OFFSET = 18
 
 """Offset of the object name setting within an object group"""
 SETTING_OBJECT_NAME_IDX = 0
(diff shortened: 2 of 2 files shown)


bethac07:
Triple fix, love it!

DavidStirling:
I'll re-test this once Master is more functional, particularly to make sure that "add_filename" doesn't try to average multiple image names or something when working with multiple channels.

PR #4005, opened 2020-05-14
Fix loading of CP3 workspaces

DavidStirling:
A problem we have is that CP3 .cpproj files have their measurements, particularly filenames, encoded as bytes rather than strings when the HDF5 data is imported in Python 3. I'm still trying to figure out how best to address this, but for now these changes decode elements needed by the GUI as necessary.

Posting this as a draft in case anyone has a better idea, I'm still trying to figure out how to deal with this in analysis mode.

### DIFF
--- a/cellprofiler/gui/imagesetctrl.py
+++ b/cellprofiler/gui/imagesetctrl.py
@@ -251,7 +251,7 @@ class ImageSetGridTable(wx.grid.GridTableBase):
         column = self.columns[col]
         if (
             column.channel_type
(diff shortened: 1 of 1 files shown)


DavidStirling:
I think https://github.com/CellProfiler/core/commit/43719f53bff7c6df5b6d1b891d09270127d49679 should resolve the vast majority of issues caused by imported workspaces. Not sure if we should still merge this to be 'extra safe' with the imageset GUI?

0x00b1:
@DavidStirling I think this is still worth merging.

PR #4009, opened 2020-05-18
Fix test mode/analysis buttons not activating when analysis aborts due to error.

### LINKED ISSUE #3897: Test Mode/Analyze buttons fail to re-enable when a run aborts due to an error
The Test and Analyze buttons in the bottom left are greyed out during a run. If the run is interrupted by some kind of error and the user chooses to cancel the run, these buttons remain greyed out (although they still work). The problem is similar to #1855, albeit with a different scenario to trigger.

DavidStirling:
Seems like this was just a visual issue, it just needed a forced update.

Fixes #3897.


0x00b1:
@DavidStirling Lovely

PR #4010, opened 2020-05-19
[WIP] Add truncation control for high and low values in CorrectIlluminationApply

### LINKED ISSUE #3829: Optionally clip intensities in CorrectIlluminationApply
CorrectIlluminationApply can sometimes lead to images with values "> 1" or "<0"; sometimes this is due to a settings mismatch (this will 100% happen if rescaling is allowed in "Background" mode of CorrectIlluminationCalculate), but sometimes it's due to a scaling or division error (see #3831).  As we do in other image processing modules like ImageMath and RescaleIntensity, it would be nice to optionally allow 0-1 clipping here (maybe set default as "No", just to maintain compatibility with older pipelines) so when such errors happen, they don't throw off the rest of the pipeline downstream.

bethac07:
Resolves #3829 

Truncation of low values was previously done automatically- this now is set to default yes, but under use control.

Truncation of high values is now added, with default set to no to maintain compatibility.


bethac07:
Closed in favor of making this change to master

PR #4013, opened 2020-05-20
Add truncation functions to CorrectIlluminationApply

### LINKED ISSUE #3829: Optionally clip intensities in CorrectIlluminationApply

bethac07:
Resolves #3829

### DIFF
--- a/cellprofiler/modules/correctilluminationapply.py
+++ b/cellprofiler/modules/correctilluminationapply.py
@@ -162,6 +162,7 @@ somewhat empirical.
             result += [image.image_name, image.corrected_image_name,
                        image.illum_correct_function_image_name,
                        image.divide_or_subtract]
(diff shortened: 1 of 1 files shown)


bethac07:
Duplicate of #4010

PR #4020, opened 2020-05-22
remove data tools

0x00b1:


bethac07:
Delete the MergeOutputFiles module too! DELETE EVERYTHING :)

0x00b1:
@bethac07 done!

DavidStirling:
Copying over some notes from the testing session:

When we remove Data Tools from the main menu, keep in mind Plate Viewer still needs to reside somewhere. Two choices we thought of: (1) Windows menu: Add the option Show/Hide Plate Viewer; and (2) add a button for Plate Viewer in the metadata module (because the ONLY time Plate Viewer can be used is if the user has extracted some fancy metadata. Also note: if we had any documentation for Data Tools menu, it could move to Help > Legacy modules and features

Issue #2712, opened 2017-04-21
Design UI to designate and change location in pipeline while in test mode


bethac07:
See #2711, #2531, #2032

bethac07:
Previously this was a slider- highlighting has been mentioned as an option to designate location, though if we go with that I think we need a bolder highlight than we currently have (the one now is quite subtle).  Maybe a right-click->"Jump to this module" to change location?

braymp:
This [page](https://www.nature.com/articles/ncomms12474) has some suggestions; the slider is one way to indicate location, but there other possibilities listed.

The slider (plus the little cross-bar) nicely showed and permitted a number of functions in a simple graphic; it's gonna be hard to replace with something simpler, I think.

On the other hand, the green arrow is more superflous; if real-estate is an issue, that could do away.

0x00b1:
I’m going to close this until someone has a concrete idea or, even better, a functional implementation. We’re trying to avoid having ambiguous discussions on GitHub. Maybe someone should create a thread on the CellProfiler forum?

bethac07:
http://forum.cellprofiler.org/t/poll-testmode-interface-options/4676

PR #4024, opened 2020-05-22
Add Test Mode current module/step from markers

DavidStirling:
This pull introduces a marker column to indicate the current active module. The user can jump back to and execute specific modules by clicking the relevant button. Replaces the 'slider' functionality from  #2712. 

StepFrom

I also dimmed the inactive 'pause' icon to be more consistent with the other buttons.

To add the extra column I've reduced the width of the spacer previously used by the slider module. I don't think that'll be a problem.


bethac07:
LOVE THIS

DavidStirling:
Also fixed being able to run/step from a disabled module.

DavidStirling:
Also fixed the list view breaking when deleting all modules

PR #4038, opened 2020-05-27
Fix console spam in EditObjectsDialog

DavidStirling:
Matplotlib are in the process of moving away from creating a wx statusbar (see matplotlib/matplotlib#17092). For the time being the current version attempts to update whatever wx statusbar is present on a plot window. I'd fixed this for the normal figure windows, but on the edit object dialogs there is no statusbar at all which upsets matplotlib.

As a fix for this version, I've added a dud function to chuck out the update commands sent by matplotlib.


0x00b1:
oh man

DavidStirling:
We might want to check that workspace viewer doesn't also suffer from this problem

PR #4044, opened 2020-05-28
Improve "Search Help" speed

DavidStirling:
Requires CellProfiler/core#6.

The search function spends a lot of time converting every module's help text into HTML before searching for the user's keyword. This pull makes CP fetch the raw help text from each module and do a preliminary scan on that to determine which modules actually need a full fat search performed on them. At least on Windows this dramatically speeds up the function.


0x00b1:
LGTM

Issue #4012, opened 2020-05-20
Make Threshold more consistent, or document its inconsistencies


bethac07:
Threshold is kind of a mess- most modules respect threshold mins and maxes, but some don't (manual), most respect the given smoothing settings, but some don't (manual, measurement), and the "local" and "global" nomenclature sometimes refers to local and global, and sometimes refers to "final" and "original".  

I know we're planning on making changes to skimage, we should consider ripping a LOT out in the rewrite, and documenting either a) what changes we make to bring things more in line, or b) where the inconsistencies remain.

0x00b1:
lol

PR #4048, opened 2020-06-02
Revise Threshold module

### LINKED ISSUE #3738: Can't save pipeline with adaptive thresholding if setting was Global/RobustBackground
From [the forum](https://forum.image.sc/t/identifyprimaryobjects-bug-cannot-save-adaptive-threshold-strategy/24844)

>My pipeline includes an IdentifyPrimaryObjects function that works well when the Threshold strategy is Adaptive. However, when I try to save this pipeline and reopen it, the Threshold strategy automatically resets to Global. (I am able to successfully edit and save other aspects of the pipeline)
In addition, when I Analyze Images, the pipeline executes a Global thresholding even when the Adaptive threshold is selected and saved.

I got this far in debugging:

>I think I see what happened- when you just change from Global (with Robust Background) to Adaptive (which only supports Otsu), the GUI is switching to Otsu and *displaying* Otsu to you, but it still "remembers" that the last threshold strategy *you chose* was RobustBackground, which only is available from Global, so it switches you from Adaptive to Global so that you can use your preferred thresholding algorithm.  If in Global mode I switch from RobustBackground to Otsu and *then* switch from Global to Adaptive, it remembers my choice just fine (see attached).  It seems to be something specific about RobustBackground- if I switch from "Global/Manual" to "Adaptive/Otsu", it save just fine for me. 

Non-working version of the pipeline here:
[dead50_v0_20151028.cpproj.zip](https://github.com/CellProfiler/CellProfiler/files/3103972/dead50_v0_20151028.cpproj.zip)

### LINKED ISSUE #3953: Threshold should use all the cool new stuff in skimage17
At least
- N class otsu
- local everything we can
- Is robust background in skimage?

DavidStirling:
Requires CellProfiler/core#7.

Closes #3953.

Probably fixes #3738.

This is a rewrite of the Threshold module to make things more consistent. Key changes are:

- When writing measurements, "Original Threshold" will always be the uncorrected threshold, while "Final Threshold" will refer to the corrected threshold with multipliers and limits applied. "Guide Threshold" has been added to record a global threshold which was used to constrain an adaptive threshold.

- MCE and RobustBackground are now available as adaptive threshold methods.

- Skimage methods are now used wherever possible. RobustBackground is unfortunately not available but the other methods were. We're still using centrosome's block-based adaptive thresholding as the skimage local_threshold function currently performs very poorly if you try to pass it a complex thresholding algorithm.

- Added Sauvola local thresholding. This will hopefully be useful as it's fast and doesn't suffer from performance degradation when using a small window size.

- Otsu 3-class thresholding now uses the standard Skimage implementation. This will change results from older pipelines. I'm not sure why but the original methods were giving much lower thresholds than you might anticipate, but on the plus side that's now fixed.

- For consistency, adaptive Otsu thresholding will now use the block-based method regardless of whether you're using 2 or 3 class mode. Skimage does have a truly local 2-class Otsu, but there was a performance hit in some situations and the results weren't substantially different.

- Smoothing can now be applied when using the Measurement and Manual methods.

- The display output from the Threshold module now includes a panel showing the local threshold values. This should allow users to inspect and understand how their min/max constraints are impacting the final threshold array.

- Moved towards having constants within Threshold, rather than duplicated between Threshold and centrosome.

I still need to fix all the tests, but making a record of all this now before I forget.

### DIFF
--- a/cellprofiler/modules/threshold.py
+++ b/cellprofiler/modules/threshold.py
@@ -50,6 +50,7 @@ TM_MANUAL = "Manual"
 TM_MEASUREMENT = "Measurement"
 TM_LI = "Minimum cross entropy"
 TM_OTSU = "Otsu"
+TM_ROBUST_BACKGROUND = "RobustBackground"
 TM_SAUVOLA = "Sauvola"
 
 
@@ -113,8 +114,8 @@ The choices for the threshold strategy are:
                 TM_MANUAL,
                 TM_MEASUREMENT,
                 TM_LI,
-                centrosome.threshold.TM_OTSU,
-                centrosome.threshold.TM_ROBUST_BACKGROUND,
+                TM_OTSU,
+                TM_ROBUST_BACKGROUND,
             ],
             value=TM_LI,
             doc="""\
@@ -236,8 +237,8 @@ There are a number of methods for finding thresholds automatically:
                     "PROTIP_RECOMMEND_ICON": _help.PROTIP_RECOMMEND_ICON,
                     "TECH_NOTE_ICON": _help.TECH_NOTE_ICON,
                     "TM_LI": TM_LI,
-                    "TM_OTSU": centrosome.threshold.TM_OTSU,
-                    "TM_ROBUST_BACKGROUND": centrosome.threshold.TM_ROBUST_BACKGROUND,
+                    "TM_OTSU": TM_OTSU,
+                    "TM_ROBUST_BACKGROUND": TM_ROBUST_BACKGROUND,
                     "TM_MANUAL": TM_MANUAL,
                     "TM_MEASUREMENT": TM_MEASUREMENT,
                     "TS_GLOBAL": TS_GLOBAL,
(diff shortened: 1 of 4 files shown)


bethac07:
Also probably closes #4012 :)

Nice work dude!

DavidStirling:
I think this is about ready now. The pipeline loading tests still have that problem with the module class assert statements, since their class is now seen as "threshold.Threshold" rather than "cellprofiler.modules.threshold.Threshold" by Python under the new module loading system. Might be able to fix that elsewhere.

cellprofiler/modules/threshold.py
```diff
@@ -47,7 +48,11 @@
 TS_ADAPTIVE = "Adaptive"
 TM_MANUAL = "Manual"
 TM_MEASUREMENT = "Measurement"
-TM_LI = "Minimum cross entropy"
+TM_LI = "Minimum Cross-Entropy"
```
0x00b1, line 51: You're killing me with these globals.

0x00b1:
This will take some time to review. Holding off until the next RC.

PR #4051, opened 2020-06-04
Fix exporting of cppipe pipeline files

DavidStirling:
Seems like exporting a pipeline didn't actually work (project file saving was ok). This hopefully should get it working.


0x00b1:
Oops. This was my bad. Thanks for fixing.

PR #4055, opened 2020-06-05
Fixup Regex editor for wx4

### LINKED ISSUE #4039: Fix the Regex preview window
Currently the display is messed up. Typing anything also causes the results to vanish.

DavidStirling:
Fixes #4039

This should make the regex window behave and look sensible with WX4. Could someone check how this looks on Mac?


0x00b1:
Just checked LGTM

PR #4058, opened 2020-06-08
Add builtin CP modules list

DavidStirling:
Required by CellProfiler/core#9


DavidStirling:
Also moved LoadData while we're at it

PR #4061, opened 2020-06-11
Add CombineObjects module

### LINKED ISSUE #2140: Combine objects identified at two different times into a single object class
Currently if you have two separately identified sets of objects (such as automatically identified nuclei and then manually identified nuclei, or a case where some nuclei are bright and some are dim so they need separate IdentifyPrimaryObjects modules to pull each type out) it's a pain and several steps to try to turn them into a single umbrella class such as "AllNuclei", but sometimes you need the combined set to proceed with your downstream analysis.  It seems like this might be much simpler to do in the backend by combining the matrices for the two or something like that, but IANASE.  If it is possible it would be a big help for at least one project each for Kyle and I, not to mention [the resolution of a forum issue from one of our top posters](http://forum.cellprofiler.org/t/ability-to-combine-objects-identified-at-different-times/1272?u=bcimini) .

DavidStirling:
Resolves #2140

Bringing back the idea from #2151.

I've gone for trying to keep this relatively simple for now. Some of the methods could probably be more efficient, but it seems to work. The general idea for this module is to allow people to combine two object sets into a single set, which is primarily useful if you needed different detection settings for different objects within the same image.

At present I've added 4 methods of handling overlapping objects, as follows:
- Merge - Touching objects are assigned the label from the first object set. When touching multiple objects each pixel is assigned to the nearest object from the initial set.
- Preserve - Add the second object set array to areas where there are no objects in the first set. Basically 'preserves' all the objects in the initial set.
- Discard - Only add objects which don't overlap with anything from the initial set.
- Segment - Add both object sets together, then re-segment overlapping areas by assigning pixels to the nearest non-conflicting region of an object.

Ultimately people shouldn't be trying to combine object sets with 'real' overlap in which two objects occupy the same space, so hopefully those options will cover the intended use cases.

I still need to write some tests, but figured I'd put a PR together in case people have further input.

### DIFF
--- a/cellprofiler/modules/combineobjects.py
+++ b/cellprofiler/modules/combineobjects.py
@@ -2,16 +2,29 @@
 
 """
 CombineObjects
-=============
+==============
 
-**CombineObjects** allows you to merge two object sets into a single object set.
+**CombineObjects** allows you to combine two object sets into a single object set.
+
+This moduled is geared towards situations where a set of objects was identified
+using multiple instances of an Identify module, typically to account for large
+variability in size or intensity. Using this module will combine object sets to
+create a new set of objects which can be used in other modules.
+
+CellProfiler can only handle a single object in each location of an image, so
+it is important to carefully choose how to handle objects which would be
+overlapping.
+
+When performing operations, this module treats the first selected object set, termed
+"initial objects" as the starting point for a joined set. CellProfiler will try to add
+objects from the second selected set to the initial set.
 
 |
 
 ============ ============ ===============
 Supports 2D? Supports 3D? Respects masks?
 ============ ============ ===============
-YES          NO          NO
+YES          NO           NO
 ============ ============ ===============
 
 """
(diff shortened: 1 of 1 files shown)


0x00b1:
@DavidStirling Do you want to ship this with 4.0.0?

cellprofiler/modules/combineobjects.py
```diff
@@ -0,0 +1,202 @@
...
+                                     sharexy=ax,
+                                     colormap=cmap,
+                                     )
+        figure.subplot_imshow_labels(0, 1, workspace.display_data.output_object,
+                                     workspace.display_data.output_object_name,
+                                     sharexy=ax,
+                                     colormap=cmap,
+                                     )
+
+    def combine_arrays(self, labels_x, labels_y):
```
0x00b1, line 156: Neat

0x00b1:
@DavidStirling Very cool. Have you tested this with volumes? You also need some tests.

DavidStirling:
@0x00b1 - I'm easy with regards to 4.0, depends how much everyone else wants this feature.

I'll be trying to put some tests together shortly. 

Regarding volumes, for now I've just specified it not being compatible with 3D, but it's entirely possible that it might work anyway. Will do some testing when it's ready.

Issue #4064, opened 2020-06-12
CP4 is not working in headless mode with Python 3.8


orgonth:
When running CellProfiler in GUI-less mode, a couple of errors are encountered:
- When using `-i` option: `AttributeError: module 'ntpath' has no attribute 'walk'`
- When `--first-image-set` is not set

This behavior is not observed in CP 3. I will propose a PR to fix this issue.

bmccartyjhu:
Adding on - I got it running by making the edits orgonth did, but did run into an issue in line 888-889 when trying to save data in a .mat format:

>    if len(args) > 0 and not use_hdf5:
        pipeline.save_measurements(args[0], measurements)

It seems that either (a) these couple of lines escaped deletion and a note that it only supports .hdf5 now should be made, or (b) this hasn't been updated to "cellprofiler_core.pipeline" (in which case the entire .mat/handles save code hasn't been transitioned yet either, as it all disappeared from the core.pipeline code)

bethac07:
Matlab reading and writing will be deprecated in CP 4.0; sorry, we've had a deprecation warning on that since CP 3.0.0

PR #4065, opened 2020-06-12
Issues/4064 headless mode

orgonth:
Fix for #4064

Make CP work in headless mode:
- with `-i` option
- without `--first-image-set`

### DIFF
--- a/cellprofiler/__main__.py
+++ b/cellprofiler/__main__.py
@@ -838,17 +838,14 @@ def run_pipeline_headless(options, args):
     elif options.image_directory is not None:
         pathnames = []
 
-        os.path.walk(
-            os.path.abspath(options.image_directory),
-            lambda pathnames, dirname, fnames: pathnames.append(
+        for dirname, _, fnames in os.walk(os.path.abspath(options.image_directory)):
+            pathnames.append(
                 [
                     os.path.join(dirname, fname)
                     for fname in fnames
                     if os.path.isfile(os.path.join(dirname, fname))
                 ]
-            ),
-            pathnames,
-        )
+            )
 
         pathnames = sum(pathnames, [])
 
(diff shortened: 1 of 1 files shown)


0x00b1:
Thanks, @orgonth !

cellprofiler/__main__.py
```diff
@@ -766,7 +766,7 @@ def run_pipeline_headless(options, args):
         else:
             image_set_start = int(options.first_image_set)
     else:
-        image_set_start = None
+        image_set_start = 1
```
0x00b1, line 769: @orgonth What's going on here?

cellprofiler/__main__.py
```diff
@@ -838,17 +838,14 @@ def run_pipeline_headless(options, args):
     elif options.image_directory is not None:
         pathnames = []
 
-        os.path.walk(
-            os.path.abspath(options.image_directory),
-            lambda pathnames, dirname, fnames: pathnames.append(
+        for dirname, subdirs, fnames in os.walk(os.path.abspath(options.image_directory)):
```
0x00b1, line 841: `subdirs` should be `_`

cellprofiler/__main__.py
orgonth: The method `Pipeline.run()` has a parameter `image_set_start`, which default value was changed from `None` to 1 in CP4.
However, in headless mode, `image_set_start` is always explicitly set when calling the method, based on the command-line parameter `--first-image-set`. I simply changed the value if that parameter is not given, to agree with the new default of `Pipeline.run()`.

cellprofiler/__main__.py
orgonth: Done.

0x00b1:
Thanks @orgonth !

PR #4074, opened 2020-06-17
Add true randomisation of figure colourmaps

### LINKED ISSUE #3956: Segmentation difficult to visualize in images with very large number of objects (IdentifyPrimaryObjects).
Using IdentifyPrimaryObjects, in images with **many** small objects it is hard to determine if segmentation is correct because of the limited number of output colors and their ordered nature. Objects with the same color are often positioned next to each other so it's hard to see where one object ends and another begins.
  
Example:
Screen Shot 2020-03-23 at 12 59 18 PM
  
Zoomed in:
Screen Shot 2020-03-23 at 12 49 47 PM

DavidStirling:
Fixes #3956 

By default matplotlib generates a 256 colour colourmap for displaying figures. We shuffle this map, but it still gets stretched when handling large numbers of labels, resulting in neighbouring objects being assigned similar colours. We can solve this by supplying the `lut` parameter in the cmap call to specify how many unique values we need.

### DIFF
--- a/cellprofiler/gui/figure.py
+++ b/cellprofiler/gui/figure.py
@@ -1826,7 +1826,7 @@ class Figure(wx.Frame):
         else:
             # Mask the original labels
             label_image = numpy.ma.masked_where(image == 0, image)
-            if colormap is None:
+            if not colormap:
                 colormap = self.return_cmap(numpy.max(image))
             else:
                 colormap = colormap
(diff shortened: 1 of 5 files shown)


cellprofiler/gui/figure.py
```diff
@@ -1826,8 +1826,8 @@ def subplot_imshow_labels(
         else:
             # Mask the original labels
             label_image = numpy.ma.masked_where(image == 0, image)
-            if colormap == None:
-                colormap = self.return_cmap()
+            if colormap is None:
```
0x00b1, line 1829: `if not colormap`

PR #4077, opened 2020-06-18
Add new regex guess for INCell

### LINKED ISSUE #3725: Add new INCell file naming scheme to metadata regex guesser
The latest version of the INCell (v7.2) acquisition software has a new preference to use a more complete output file name that encapsulates all imaging options (well, laser, filter, timepoint, z-pos, field, etc).

Under the longstanding default settings in the program, the image files are generally named like this:
`A - 01(fld 2 wv Blue - FITC).tif` or sometimes this:
`A - 01(fld 1 wv UV - DAPI z 3 time 2 - 2636ms).tif` or even this occasionally:
`A - 13(fld 2).tif`

With the new preference set, the names now always look like this, regardless of how many of the parameters are being used in the run:
`A_01_f2_c1_xUV_mDAPI_z3_t2.tif`

This naming scheme should be much easier to parse in a consistent manner since all metadata tokens are always present. A new guesser should be added to `cellprofiler/gui/regexp_editor.py` for this new scheme.

I think this is correct:
````
^(?P<Row>[A-P])_(?P<Column>[0-9]*)_f(?P<Site>[0-9]*)_c(?P<Channel>[0-9]*)_x(?P<Wavelength>.*)_m(?P<Filter>.*)_z(?P<Slice>[0-9]*)_t(?P<Timepoint>[0-9]*)\.tif$
````

DavidStirling:
Closes #3725.

### DIFF
--- a/cellprofiler/gui/regexp_editor.py
+++ b/cellprofiler/gui/regexp_editor.py
@@ -40,7 +40,7 @@ RE_FILENAME_GUESSES = [
     r"^(?P<Row>[A-H]*) - (?P<Column>[0-9]*)\(fld (?P<Site>[0-9]*) wv (?P<Wavelength>.*) - (?P<Filter>.*)\)",
     # Phenix
     r"^r(?P<WellRow>\d{2})c(?P<WellColumn>\d{2})f(?P<Site>\d{2})p\d{2}-ch(?P<ChannelNumber>\d)",
-    # INCell
+    # GE InCell Analyzer 7.2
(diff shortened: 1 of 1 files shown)


nieder:
Please note that there's already an IN Cell section in line 39.

What does the leading '`r`' do before the regex, and should this be 2 lines in the code?

DavidStirling:
Good point, I'll make a more detailed comment. It's not visible in the interface.

The 'r' specifies that this is a raw string, so any backslashes won't be treated as escaped characters which Python should interpret, such as \n for "new line". The line break is for compliance with the PEP 8 style guidelines - a single line of Python shouldn't be too long.

PR #4081, opened 2020-06-26
Implement fftconvolve in place of binary_dilate for MeasureObjectNeighbors

DavidStirling:
As raised on Slack, MeasureObjectNeighbors had miserable performance when using larger expansion distances. I found that this was a problem with scipy.ndimage's binary_dilation function (mentioned [here](https://github.com/scikit-image/scikit-image/issues/1190)).

However, per [this thread](https://stackoverflow.com/a/51159444), it's possible to achieve the same results using the Fast Fourier Transform. In my testing this gives us exactly the same results with a substantial increase in speed and better memory usage. The caveat to this is that it's marginally slower when working with small images and structuring elements, although at that size the overall time taken is relatively small to begin with. 

Some testing results on 1 image, with 250 objects:
- Old method, 1 pixel dilation - 0.375s
- New method, 1 pixel dilation - 0.422s
- Old method, 25 pixel dilation - 29.171s
- New method, 25 pixel dilation - 0.625s

As for the 3D monolayer example, with 26 objects:
- Old method, 5 pixel 3D dilation - 1.813s
- New method, 5 pixel 3D dilation - 1.750s
- Old method, 25 pixel 3D dilation - Failed, memory error
- New method, 25 pixel 3D dilation - 6.922s

I'm not sure if it's worth it, but we could have it fall back to using standard binary_dilation when using a distance <5?

### DIFF
--- a/cellprofiler/modules/measureobjectneighbors.py
+++ b/cellprofiler/modules/measureobjectneighbors.py
@@ -544,7 +544,7 @@ previously discarded objects.""".format(
                     else:
                         extended = scipy.signal.fftconvolve(extendme, strel_touching, mode='same') > 0.5
                 else:
-                    if distance <= 3:
+                    if distance <= 5:
                         extended = scind.binary_dilation((npatch != 0), strel_touching)
                     else:
                         extended = scipy.signal.fftconvolve((npatch != 0), strel_touching, mode='same') > 0.5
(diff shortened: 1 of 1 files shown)


DavidStirling:
@0x00b1 I've now made it so that the original method is used when the expansion distance is small (<=5). That seems to be a reasonable cutoff for where convolve method becomes faster.

PR #4102, opened 2020-07-20
Rework MeasureObjectSizeShape to use new regionprops measurements

### LINKED ISSUE #3498: Report more regionprops measures when skimage 0.14 hits
It seems a number of 3D regionprops measurements have been added in skimage since the 0.13 release in 10/17; when the next release hits we should add those to the appropriate modules.

### LINKED ISSUE #3571: Add bounding box (and potentially other regionprops) to MeasureObjectSizeShape

### LINKED ISSUE #3657: Upgrade to skimage 0.14.1

### LINKED ISSUE #3841: Solidity returns a value >1
H/T [the forum](https://forum.image.sc/t/solidity-greater-1/29256);

>When using the Solidity measure in CellProfiler, we get values greater than 1. By definition, Solidity cannot be greater than 1. Anyone had this problem before?

Zoom in of offending area below, and pipeline/image to reproduce are attached (I've reproduced in Mac 3.1.8 and 3.1.9 built).  This is calculated in [centrosome](https://github.com/CellProfiler/centrosome/blob/313d0ddcc23574521688da8fa111d89f50321cc5/centrosome/cpmorphology.py#L2466-L2477) based on centrosome's [convex hull area implementation
](https://github.com/CellProfiler/centrosome/blob/313d0ddcc23574521688da8fa111d89f50321cc5/centrosome/cpmorphology.py#L2466-L2477), so guessing the fix may ultimately end up being over there but I haven't rigorously checked yet.

image

[SolidityBug.zip](https://github.com/CellProfiler/CellProfiler/files/3627744/SolidityBug.zip)

### LINKED ISSUE #3989: Surface area calculations in 3D are not performant
**Describe the bug**
Surface area calculations take about 3 seconds per object and are done sequentially, rather than serially, so for large numbers of objects it can lead to module run times of 10hr or more.  H/T [this forum post](https://forum.image.sc/t/cellprofiler-3-1-9-measureobjectsize-shape-in-3-d-error/36302)

We should consider one or both of a) switching to marching_cubes_lewiner rather than marching_cubes_classic and b) making SA optional to run, like we do with Zernikes (or like we've added in MeasureColocalization).

 - OS: OSX 14.6
 - CellProfiler Version 3.1.9
 - Downloaded from the website or installed from source? Built and Source

DavidStirling:
Resolves #3571. Resolves  #3498. Fixes #3841. Requires CellProfiler/core#19. Rebased/cleaned up from #4099.

I've had a go at trying to explain the new metrics simply, any further ideas would be very helpful.

---

The aim of this pull request is to migrate towards using Skimage regionprops functionality within the MeasureObjectSizeShape module. This provides some performance improvements but more importantly gives us access to a series of new measurements. We were already using regionprops in 3D mode.

I've added a new setting called 'Calculate the advanced features?' to allow users to enable/disable some of the new features. In 2D mode this toggles the 'Moments' features, since these produce an excessive number of potentially unwanted results columns. In 3D mode the setting currently toggles the 'solidity' parameter, since it's computationally expensive. For 3D objects I've also switched the surface area marching cubes method to the Lewiner implementation. Without solidity (which wasn't in the original 3D mode anyway) this gives us a 10-20x speed improvement over the original module when in 3D mode. Hopefully that resolves #3989.

In general most measurements from skimage are almost identical to the original centrosome implementation. One difference is the perimeter metric, which measures in pixels rather than pixel borders (a 'corner' will count as +1, rather than +2). Positional measurements like centroids are also whole pixels rather than decimals.

### DIFF
--- a/cellprofiler/modules/measureobjectsizeshape.py
+++ b/cellprofiler/modules/measureobjectsizeshape.py
@@ -512,10 +512,9 @@ module.""".format(
                 ]
 
         # Check for overlapping object sets
-        try:
-            hasattr(objects, 'segmented')
+        if not objects.overlapping():
             features_to_record = self.analyse_objects(objects, desired_properties)
-        except AssertionError:
+        else:
             # Objects are overlapping, process as single arrays
             coords_array = objects.ijv
             features_to_record = {}
--- a/cellprofiler/modules/measureobjectsizeshape.py
+++ b/cellprofiler/modules/measureobjectsizeshape.py
@@ -809,15 +809,4 @@ module.""".format(
             self.record_measurement(workspace, object_name, feature_name, empty_measure)
 
 
-def form_factor(objects):
-    """FormFactor = 4/pi*Area/Perimeter^2, equals 1 for a perfectly circular"""
-    if len(objects.indices) > 0:
-        perimeter = objects.fn_of_label_and_index(
-            centrosome.cpmorphology.calculate_perimeters
-        )
-        return 4.0 * numpy.pi * objects.areas / perimeter ** 2
-    else:
-        return numpy.zeros((0,))
-
-
 MeasureObjectAreaShape = MeasureObjectSizeShape
--- a/cellprofiler/modules/measureobjectsizeshape.py
+++ b/cellprofiler/modules/measureobjectsizeshape.py
@@ -222,7 +222,7 @@ F_STANDARD = [
 
 class MeasureObjectSizeShape(cellprofiler_core.module.Module):
     module_name = "MeasureObjectSizeShape"
-    variable_revision_number = 2
+    variable_revision_number = 3
     category = "Measurement"
 
     def create_settings(self):
@@ -237,6 +237,19 @@ class MeasureObjectSizeShape(cellprofiler_core.module.Module):
             doc="""Select the object sets whose size and shape you want to measure.""",
         )
         self.spacer = cellprofiler_core.setting.Divider(line=True)
+
+        self.calculate_advanced = cellprofiler_core.setting.Binary(
+            text="Calculate the advanced features?",
+            value=True,
+            doc="""\
+Select *{YES}* to calculate additional statistics for object moments
+and intertia tensors. These features should not require much additional time
+to calculate, but do add many additional columns to the resulting output 
+files.""".format(
+                **{"YES": "Yes"}
+            ),
+        )
+
         self.calculate_zernikes = cellprofiler_core.setting.Binary(
             text="Calculate the Zernike features?",
             value=True,
@@ -252,12 +265,12 @@ module.""".format(
 
     def settings(self):
         """The settings as they appear in the save file"""
-        result = [self.objects_list, self.calculate_zernikes]
+        result = [self.objects_list, self.calculate_zernikes, self.calculate_advanced]
         return result
 
     def visible_settings(self):
         """The settings as they appear in the module viewer"""
-        result = [self.objects_list, self.spacer, self.calculate_zernikes]
+        result = [self.objects_list, self.spacer, self.calculate_zernikes, self.calculate_advanced]
         return result
 
     def validate_module(self, pipeline):
@@ -650,7 +663,7 @@ module.""".format(
                     chulls, chull_counts, objects.indices
                 )
 
-            for f, m in [
+            features_to_record = [
                 (F_AREA, props["area"]),
                 (F_PERIMETER, props["perimeter"]),
                 (F_MAJOR_AXIS_LENGTH, props["major_axis_length"]),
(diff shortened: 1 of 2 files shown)


cellprofiler/modules/measureobjectsizeshape.py
```diff
@@ -119,6 +125,23 @@
    While there is no limit to the order which can be calculated (and
    indeed you could add more by adjusting the code), the higher order
    polynomials carry less information.
+-  *Spatial Moment features:* *(2D only)* A series of weighted averages
```
0x00b1, line 128: Hooray!

cellprofiler/modules/measureobjectsizeshape.py
```diff
@@ -179,45 +205,142 @@
...
 F_ORIENTATION = "Orientation"
 F_COMPACTNESS = "Compactness"
+F_INERTIA = "InertiaTensor"
 F_MAXIMUM_RADIUS = "MaximumRadius"
 F_MEDIAN_RADIUS = "MedianRadius"
 F_MEAN_RADIUS = "MeanRadius"
 F_MIN_FERET_DIAMETER = "MinFeretDiameter"
 F_MAX_FERET_DIAMETER = "MaxFeretDiameter"
 
+F_CENTRAL_MOMENT_0_0 = "CentralMoment_0_0"
```
0x00b1, line 229: So many features! 👏

cellprofiler/modules/measureobjectsizeshape.py
```diff
@@ -336,130 +476,138 @@ def run(self, workspace):
...
+                "equivalent_diameter",
+                "euler_number",
+            ]
+            if self.calculate_advanced.value:
+                desired_properties += [
+                    "solidity",
+                ]
+
+        # Check for overlapping object sets
+        if not objects.overlapping():
```
0x00b1, line 541: Much better, thanks for fixing.

cellprofiler/modules/measureobjectsizeshape.py
```diff
@@ -336,130 +476,138 @@ def run(self, workspace):
...
+                buffer = self.analyse_objects(tempobject, desired_properties)
+                for f, m in buffer.items():
+                    if f in features_to_record:
+                        features_to_record[f] = numpy.concatenate((features_to_record[f], m))
+                    else:
+                        features_to_record[f] = m
+        for f, m in features_to_record.items():
+            self.record_measurement(workspace, object_name, f, m)
+
+    def analyse_objects(self, objects, desired_properties):
```
0x00b1, line 562: uh oh 🇬🇧

cellprofiler/modules/measureobjectsizeshape.py
```diff
@@ -336,130 +476,138 @@ def run(self, workspace):
...
+        """Computing the measurements for a single map of objects"""
+        labels = objects.segmented
+        nobjects = len(objects.indices)
+        if len(objects.shape) == 2:
+            props = skimage.measure.regionprops_table(labels, properties=desired_properties)
```
0x00b1, line 567: Why `regionprops_table` rather than `region props`?

DavidStirling: When given the `properties` parameter the table method will only calculate the measurements we're actually going to use. `_table` also saves us having to construct output arrays and iterate over the objects to extract their properties.

As a point of interest, Skimage 0.18 is going to add an `extra_properties` parameter to regionprops which will allow custom properties to be calculated. That might be a good pathway to replace the rest of the Centrosome functionality.

cellprofiler/modules/measureobjectsizeshape.py
DavidStirling: Ack, need me to fix this?

DavidStirling: Done

DavidStirling:
Confirmed that this also fixes #3841

PR #4104, opened 2020-07-20
Fix tests for CP4

DavidStirling:
In this PR I've started fixing the module tests. Currently that's addressed some bugs in TrackObjects and a few other modules.

For the tests themselves I've improved their ability to find pipeline files, which seemed to just not work on Windows before. ExportToDatabase is still a work in progress.


0x00b1:
merging ... we will revisit this/next week

0x00b1:
(I don't want this to get left behind)

PR #4108, opened 2020-07-21
Fix module list view drag behaviour

### LINKED ISSUE #3966: Viewing module settings sometimes fails
I've noticed thattrying to view modules sometimes produces an entirely blank panel rather than the module notes and settings that should be displayed, particularly with complex modules like IDPrimary and ExportToDatabase. Clicking on the module again in the list to refresh it then makes it display properly. This might be more likely when switching modules rapidly, but I'm not sure how to consistently replicate this yet.

If we're looking into #3962 this might be another issue to be aware of and keep an eye out for.

DavidStirling:
Fixes #3966

It turns out that `ScreenToClient` can sometimes return outdated mouse coordinates if the user moves the mouse and clicks a module too fast. This was making CellProfiler inappropriately try to drag and drop the targetted module rather than viewing it. Since the target module was deleted and replaced with a copy, this would make the settings fail to show.

To fix this I've made the drag/drop start position get set using the x, y coords sent over from the other functions. I've also made it properly check and ignore dropping to the same position in the pipeline (no valid change).

### DIFF
--- a/cellprofiler/gui/pipelinelistview.py
+++ b/cellprofiler/gui/pipelinelistview.py
@@ -715,7 +715,7 @@ class PipelineListView(object):
         drop_source = wx.DropSource(self.list_ctrl)
         drop_source.SetData(data_object)
         self.drag_underway = False
-        self.drag_start = self.list_ctrl.ScreenToClient(wx.GetMousePosition())
+        self.drag_start = None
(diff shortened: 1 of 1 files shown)


0x00b1:
lgtm

PR #4109, opened 2020-07-21
Fix exceptions when trying to exit CellProfiler while Test Mode is enabled.

DavidStirling:
It looks like the BioFormats reader cache is normally cleared when leaving test mode, or when a headless CP instance exits, but not if you quit the GUI without leaving test mode. Adding a step to the exit workflow seems to fix the exception spam when exiting, might be worth verifying this on Mac.

I'm also not sure if the original exit workflow will work properly in headless mode.


0x00b1:
lgtm on macos

PR #4110, opened 2020-07-22
Make SaveCroppedObjects compatible with color images

DavidStirling:
Needed this for a project so it's a quick fix for now. Skimage won't allow you to supply an intensity image with more dimensions than the object mask, but if we're working in 2D we can just use the same mask for each channel. Saving takes longer in RGB mode and isn't as versatile as SaveImages, but I think it's an improvement.

I've also disabled the 'low contrast' warning that Skimage IO puts out if an image is dim, since that gets really annoying when exporting 2000 images.

Documentation also appears to have been incorrect, so I've updated it.


0x00b1:
As one of the few users of this module, I think you.

Issue #3104, opened 2017-09-12
[Ubuntu 16.04] Regular expression box disappear and reappear


minh-doan:
After entering some string in regular expression box, the box disappears.

<img width="1188" alt="screen shot 2017-09-12 at 11 15 51 am" src="https://user-images.githubusercontent.com/22305129/30339319-a48b8878-97bc-11e7-9bb6-b3255cabc353.png">

If you then hit "no" and then "yes" (for question "Extract metadata?") , the box reappears again.

<img width="1188" alt="screen shot 2017-09-12 at 11 17 24 am" src="https://user-images.githubusercontent.com/22305129/30339333-b01b919c-97bc-11e7-8aab-9e044406ff8d.png">

AetherUnbound:
I can verify this behavior. Also, switching the `Metadata source` will cause this to occur too.

AetherUnbound:
Debugging leads me to believe the problem is in the `do_reset` function on line 2312 of `moduleview`

AetherUnbound:
It looks like `on_focus_killed` was calling `set_selection` twice. Essentially the settings were being refreshed twice after a small delay which was triggered by changing the regex textbox. Many of the other `on_<x>` calls have checks to ensure they pass if a settings refresh is in progress, I think `on_focus_killed` was just missing it.

PR #4111, opened 2020-07-22
Fix Metadata regex field behaviour

### LINKED ISSUE #2738: regular expression text field in "Metadata" loses focus while typing
The "Regular Expression" text field in the Metadata module loses focus while I am typing in the current master branch. Cellprofiler is run in a conda environment installed as outlined on the wiki under "conda install".

This behaviour is slightly different depending on whether this is running on Mac OS X Sierra or on Linux (Centos 7). On the Mac, the text field simply loses focus while I'm typing in it, on Linux it completely disappears. Not sure whether this is a bug in wxwidgets or in CP.

DavidStirling:
Fixes #2738

At least on Windows, once a module view refresh was triggered (5 seconds after typing) the regex text boxes in the Metadata module became completely unresponsive to user input (related to #3104).

Allowing event processing to continue after `on_kill_focus` appears to resolve this problem and allows further editing after a settings refresh.

On top of this, I've added an option to disable a text box's timed module refresh and used this to disable such refreshing specifically for the regex fields. It often takes more than 5 seconds to write an expression and so the refresh could disrupt input. Since the regex fields don't directly impact other module content I don't think the lack of a refresh will cause any problems.

### DIFF
--- a/cellprofiler/gui/moduleview.py
+++ b/cellprofiler/gui/moduleview.py
@@ -1435,7 +1435,7 @@ class ModuleView(object):
             sizer.Add(bitmap_button, 0, wx.EXPAND)
 
             def on_cell_change(event, setting=v, control=text_ctrl):
-                self.__on_cell_change(event, setting, control)
+                self.__on_cell_change(event, setting, control, timeout=False)
 
             def on_button_pressed(event, setting=v, control=text_ctrl):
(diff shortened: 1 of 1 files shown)


0x00b1:
Much, much better. Great job investigating.

PR #4112, opened 2020-07-22
Add column sorting to the metadata table

### LINKED ISSUE #3978: Sort by Column Header in Metadata Module
It would be really lovely to be able to click on the headers in the bottom of the Metadata module to  sort by that column. (i.e. When NamesAndTypes isn't playing nicely it's often a problem with Metadata extraction and sorting by column would make browsing for errors in metadata extraction much easier).

DavidStirling:
Resolves #3978

Been looking into some other issues with table display and figured out how to do this. Pretty basic but hopefully functional. Someone should verify this works properly on Mac, as always.

Clicking a column header should make the table sort itself based on that column


0x00b1:
Tested on macOS. It works! Hooray!

PR #4114, opened 2020-07-24
Create ErodeObjects, rename Erosion to ErodeImage

DavidStirling:
Requires CellProfiler/core#22

We had DilateImage and DilateObjects modules, but only Erosion (which only supported images). ExpandOrShrinkObjects is generally better but only works in 2D, so we lacked a pathway for shrinking 3D objects. I've therefore renamed Erosion to ErodeImage and created ErodeObjects.

Another difference is that ErodeObjects will completely remove a tiny object, but ExpandOrShrink will always preserve 1 pixel. That gives it a slightly different use case even in 2D.


braymp:
I could foresee single-pixel object removal producing a problem for downstream modules (or perhaps our intuitive assumptions). Is there a way to preserve ExpandOrShrink's behavior on this score?

0x00b1:
> I could foresee single-pixel object removal producing a problem for downstream modules (or perhaps our intuitive assumptions). Is there a way to preserve ExpandOrShrink's behavior on this score?

Because objects would be reindexed? (Hi, @braymp ! I hope you're doing okay during this weird time!)

DavidStirling:
I thought that might be a potential issue, but wasn't able to trigger an error when playing with it. It's largely using the default ObjectProcessing functions, so I don't think everything gets relabelled afterwards.

Replicating the ExpandOrShrink functionality would be a bit more complicated. Right now we just use skimage's dilate and erode functions in these modules.

braymp:
I think I've had cases where an inadvertently missing object may break any 1:1 object relationship assumptions, e.g., ExportToSpreadsheet where you can horizontally concatenate primary/secondary/tertiary object tables and the like.

DavidStirling:
I can imagine a couple of options here:

- Relabel the objects before adding the eroded object set to the sets list.

- Make a record of the centroid of each label and add those back in after dilation, so as to preserve all objects in the same manner as ExpandOrShrink

Could either of these work?

braymp:
I'm leaning towards the latter, perhaps an option to "Do not remove objects post-erosion" (yes/no)?

DavidStirling:
I started adding that feature and then realised that the Skimage erosion filter is actually not going to work on a segmentation image. It treats the labels as intensities and therefore ends up inappropriately changing touching objects. I'll have to rework this.

DavidStirling:
Okay, I've reworked this to process objects individually and added options for preserving midpoints and relabelling the resulting array.

PR #4115, opened 2020-07-24
expand measuregranularity documentation

### LINKED ISSUE #3992: Make MeasureGranularity help text much more descriptive
I fell down a rabbit hole on this [on the forum](https://forum.image.sc/t/cellprofiler-granularity-measurements/36966/2?u=bcimini), but the good news is that now for the first time after 10 years using that software I finally understand how it works!

We should probably incorporate some of this stuff into the module help; I will keep the notebook around in case we think generating images might be helpful.

ErinWeisbart:
Resolves #3992.

@bethac07 please take a close look that I've described things correctly as you're the one who understands this best!


0x00b1:
Thanks, @ErinWeisbart !

PR #4124, opened 2020-07-31
Fix for pipelinelistview drag/drop functionality with wx 4.1

DavidStirling:
Hopefully backwards compatible.

Things to test on OSX:
- Dragging a module moves it. Module persists when activated.
- Dragging while holding `control` duplicates that module at the targetted position.
- Dragging to somewhere else on the window (not the pipeline list) does not delete modules.
- Dragging to a second CP window moves the module from one to the other, deleting it from the source window (with ctrl it should copy). Pipeline then doesn't revert when a module is activated.

This may also finally enable moving AND duplication on OSX, since we'd disabled the latter.

### DIFF
--- a/cellprofiler/gui/module_view/_module_view.py
+++ b/cellprofiler/gui/module_view/_module_view.py
@@ -128,7 +128,6 @@ class ModuleView:
         self.notes_panel = notes_panel
         self.__frame = frame
         self.top_panel = top_panel
-        self.showing_tables = False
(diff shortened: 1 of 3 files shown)


DavidStirling:
Turns out that in 4.1 the "text wrapping when the previous module showed a table" bug is fixed, so I've removed the workaround I added for that.

This will need a rebase for proper testing, since Master was broken at the time of writing.

PR #4126, opened 2020-07-31
Prompt users to drag & drop pipelines to get started

### LINKED ISSUE #2601: Prompt users to drag & drop pipelines to get started
The Images module file list space has dark gray text stating "Drop files and folders here" to prompt users who are not oriented to the software have a clue how to get started.

Similarly, it would be nice to state "Drop a pipeline file here (.cppipe or .cpproj)" in the pipeline window upon opening the software. Apart from the welcome window which provides guidance, CP is quite blank upon launching without much guidance for what to do.

DavidStirling:
Closes #2601.

I've added this to make using the pipeline list a bit clearer to new users. Are we happy with this text?

I've also made it so that double clicking on the pipeline list will open the 'add modules' window if you aren't clicking on an existing module.

Also fixed a bug where the step function would try to run when not in test mode. The icon was invisible but still tried to do things.

Dropper

### DIFF
--- a/cellprofiler/gui/pipelinecontroller.py
+++ b/cellprofiler/gui/pipelinecontroller.py
@@ -2313,6 +2313,9 @@ class PipelineController(object):
         self.__add_module_frame.Show()
         self.__add_module_frame.Raise()
 
+    def open_add_modules(self):
+        self.__on_add_module(None)
+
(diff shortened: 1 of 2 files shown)


0x00b1:
Oh, nice. I like it.

PR #4129, opened 2020-08-03
New figure contrast adjustment dialog

### LINKED ISSUE #2446: Improve brightness/contrast options for image display
Copied from a discussion of #2285 by @tischi  regarding image display in CellProfiler:

I think CP would greatly benefit from a more sophisticated Brightness and Contrast adjustment. It is probably enough to be able to enter(slider) min and max values which are mapped to dark and bright, as sometimes neither the log-scale nor the auto-contrast work.

DavidStirling:
Resolves #2446.

This PR replaces the right click submenus for adjusting figure contrast with an all-in-one popup. Right clicking a figure plot now has an option labelled "Change Contrast" which brings up the following tool window:

Contrast

Within this you can change the normalisation mode, intensity display limits and normalisation factor (for log and gamma modes). The sliders represent a 'sensible' range, but these bounds can be exceeded by entering values into the boxes next to them. Changing these settings should update the plot immediately to give an instant preview.

I've also added an 'Apply to all' button. This will grab the display/normalisation settings and try to apply them to other subplots in the same window which are displaying an image... so that you don't have to repeat the process on multiple plots.

For colour images, normalisation modes work fine but the brightness sliders won't take effect due to the way the normalisation was originally set up. Nonetheless brightness adjustment is no longer exclusive to raw/linear normalised mode so I think that's an improvement.

Could probably do with some more testing on 'weird' images.

### DIFF
--- a/cellprofiler/gui/figure.py
+++ b/cellprofiler/gui/figure.py
@@ -51,7 +51,6 @@ from cellprofiler_core.preferences import (
     get_default_colormap,
     INTENSITY_MODE_GAMMA,
     INTENSITY_MODE_LOG,
-    INTENSITY_MODE_NORMAL,
     INTENSITY_MODE_RAW,
     get_intensity_mode,
 )
@@ -59,7 +58,6 @@ from cellprofiler_core.utilities.core.object import overlay_labels
 from numpy.ma import MaskedArray
 from wx.lib.intctrl import IntCtrl
 from wx.lib.masked import NumCtrl, EVT_NUM
-from wx.lib.plot import PolyHistogram, PlotGraphics, PlotCanvas
 
 import cellprofiler.gui
 import cellprofiler.gui.artist
--- a/cellprofiler/gui/figure.py
+++ b/cellprofiler/gui/figure.py
@@ -51,10 +51,13 @@ from cellprofiler_core.preferences import (
     get_default_colormap,
     INTENSITY_MODE_GAMMA,
     INTENSITY_MODE_LOG,
+    INTENSITY_MODE_NORMAL,
     INTENSITY_MODE_RAW,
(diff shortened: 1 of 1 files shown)


0x00b1:
I immediately ran into a need for this today:

image

cellprofiler/gui/figure.py
```diff
@@ -306,7 +309,7 @@ def get_matplotlib_interpolation_preference():
     elif interpolation == IM_BILINEAR:
         return "bilinear"
     elif interpolation == IM_BICUBIC:
-        return "bilinear"
+        return "bicubic"
```
0x00b1, line 312: This was my bad!

cellprofiler/gui/figure.py
```diff
@@ -1173,59 +1142,265 @@ def show_hist(evt):
...
+            size = self.images[(x, y)].shape[:2]
+            axesdata = axes.plot([0, 0], list(size), "k")[0]
 
-        def adjust_gamma(evt):
-            dlg = wx.TextEntryDialog(self, "Normalization factor", "Adjust gamma")
-            dlg.SetValue(get_normalization_factor())
-            if dlg.ShowModal() == wx.ID_OK:
-                params["normalize_args"] = {"gamma": float(dlg.GetValue())}
-            dlg.Destroy()
+            with wx.Dialog(self, title="Adjust contrast", size=wx.Size(250, 300)) as dlg:
```
0x00b1, line 1169: Would you mind extracting this into a seperate Python module and class?

DavidStirling: Note for the future: We decided it'd be best to do this as part of a larger refactor in which we'd split figure.py into a more sensible structure.

PR #4138, opened 2020-08-10
warn user for large image sets

alicelucas:
Pop-up warning window now appears when user hits "Analyze images" in the cases where number of image sets is greater than 500. Th e warning window explains that number of image sets is high (>500) and it is recommended that user switches module to ExportToDataset prior to continuing with the pipeline.


0x00b1:
It looks like you committed some logs.

cellprofiler/modules/exporttospreadsheet.py
```diff
@@ -721,6 +721,23 @@ def prepare_run(self, workspace):
...
         returns False if analysis can't be done
         """
+        maximum_image_sets = 500
+        if workspace.measurements.image_set_count > maximum_image_sets:
+            msg = "You are using ExportToSpreadsheet to export " + str(maximum_image_sets) + " image sets. " \
+                                                                     "Instead we suggest using ExportToDatabase" \
+                                                                     "because ExportToSpreadsheet " \
+                                                                     "may fail on large image sets." \
+                                                                     " Do you want to continue?"
+            import wx
```
0x00b1, line 731: This will break on headless mode if wx isn't installed. You should whether wx is available before importing like this.

cellprofiler/modules/exporttospreadsheet.py
```diff
@@ -721,6 +721,23 @@ def prepare_run(self, workspace):
 
         returns False if analysis can't be done
         """
+        maximum_image_sets = 500
```
0x00b1, line 724: Why 500? 🤔

cellprofiler/modules/exporttospreadsheet.py
```diff
@@ -721,6 +721,23 @@ def prepare_run(self, workspace):
...
+        if workspace.measurements.image_set_count > maximum_image_sets:
+            msg = "You are using ExportToSpreadsheet to export " + str(maximum_image_sets) + " image sets. " \
```
0x00b1, line 726: Use an f-string.

cellprofiler/modules/exporttospreadsheet.py
```diff
@@ -721,6 +721,23 @@ def prepare_run(self, workspace):
...
+                                                                     "because ExportToSpreadsheet " \
+                                                                     "may fail on large image sets." \
+                                                                     " Do you want to continue?"
+            import wx
+            result = wx.MessageBox(
+                msg,
+                caption="ExportToSpreadsheet: Large number of image sets",
+                style=wx.YES_NO | wx.NO_DEFAULT | wx.ICON_QUESTION,
+            )
+            if result != wx.YES:
```
0x00b1, line 737: I would consider flipping this for readability sake.

Issue #4141, opened 2020-08-11
Example pipeline settings should be updated before release.  


bethac07:
**Is your feature request related to a problem? Please describe.**

The current example settings lead to bad segmentation, we should fix before releasing.

![image](https://user-images.githubusercontent.com/6721515/89902071-fa1c2980-dbb3-11ea-9745-453b8b5af20f.png)

bethac07:
Closed by #4142

PR #4142, opened 2020-08-11
better example pipeline settings

bethac07:
image


bethac07:
Resolves #4141

DavidStirling:
Thanks, had been meaning to do this!

PR #4145, opened 2020-08-11
Set sensible colormap defaults

DavidStirling:
This PR makes modules which display objects coloured by data values use a more sensible (linear) colormap by default.

In the case of MeasureObjectNeighbours, it'll no longer use the default (random) map when colouring by number of neighbors. The map was/is only customisable if you enable retention of the plot as an image, so I think a better default makes sense.

In MeasureObjectIntensityDistribution the same applies to heatmaps, and I've also disabled colorbars on this module since they really don't play well with the WX table plot.

### DIFF
--- a/cellprofiler/modules/measureobjectneighbors.py
+++ b/cellprofiler/modules/measureobjectneighbors.py
@@ -206,6 +206,7 @@ of neighbors to be selected later in the pipeline.""",
 
         self.count_colormap = Colormap(
             "Select colormap",
+            value="Blues",
             doc="""\
 *(Used only if the image of objects colored by numbers of neighbors is
 to be retained for later use in the pipeline)*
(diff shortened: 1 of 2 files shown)


cellprofiler/modules/measureobjectneighbors.py
```diff
@@ -727,7 +729,7 @@ def run(self, workspace):
             count_image = Image(img, masking_objects=objects)
             image_set.add(self.count_image_name.value, count_image)
         else:
-            neighbor_cm_name = get_default_colormap()
+            neighbor_cm_name = "Blues"
```
0x00b1, line 732: Why did we choose these defaults?

0x00b1: Asking to make sure they are colorblind friendly.

DavidStirling: For future record - a single colour scale will always be fine, since it's the relative brightness that matters instead of colour hue.

PR #4149, opened 2020-08-13
Bin grey levels before running MeasureTexture

### LINKED ISSUE #3386: MeasureTexture on objects is exceptionally slow in 3.0.0
Measuring 3 channels in 3 objects at 3 scales is taking >3 hours; measuring just the image texture features for those 3 channels/3 scales inthat image takes 20-30 seconds, so it's likely something in the objects.  Pipeline + images are confidential but will be sent via email.

bethac07:
Resolves #3386 .  

This PR allows you to adjust your grey levels before running texture feature extraction; it doesn't look like this produces an exponential speedup though, just about 3x in my (very limited) tests going from 256 grey levels to 2, and almost no appreciable speedup at all going from 256 to 8.

I think the slowdown actually didn't come from the change in gray levels, but from the fact that in 2.X we run [all objects at once](https://github.com/CellProfiler/CellProfiler/blob/stable/cellprofiler/modules/measuretexture.py#L564-L568) and in 3.X we run [one object at a time](https://github.com/CellProfiler/CellProfiler/blob/v3.1.9/cellprofiler/modules/measuretexture.py#L608-L619); image texture features are still pretty fast in 3/4, no matter what the number of gray levels.  I'll look into if there's an easier way to run all the objects at a time.

### DIFF
--- a/cellprofiler/modules/measuretexture.py
+++ b/cellprofiler/modules/measuretexture.py
@@ -17,10 +17,12 @@ behavior, use multiple **MeasureTexture** modules to specify the
 particular image-object measures that you want.
 
 Note also that CellProfiler in all 2.X versions increased speed by binning 
-the image into only 8 greyscale levels before calculating Haralick features; 
-this is not done in CellProfiler versions 3.0.0 and after. Values calculated in 
-MeasureTexture in CellProfiler 2 versions will therefore not directly correspond 
-to those in CellProfiler 3 and after. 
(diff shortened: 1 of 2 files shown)


0x00b1:
So should we merge this if this doesn't fix the issue?

bethac07:
1) I think maybe, because being able to set grey levels might be nice for people who want to, say, have a more CellProfiler2-like-functionality
2) Even a nominal speedup might be nice for something that currently can take 3+ hours (as in the linked issue).

PR #4173, opened 2020-08-25
Make metadata matching auto-fill empty columns

### LINKED ISSUE #4027: Populate row for image set matching in NamesAndTypes
**Is your feature request related to a problem? Please describe.**
In NamesAndTypes when matching many image sets (e.g my current edge-case pipeline matches >40 image sets), it is slow and frustrating to pick from the drop down the metadata for each. Even in more typical pipelines, usually most of the image sets match a single metadata across a row.

**Describe the solution you'd like**
It would be nice if there could be the option to add a single metadata to all the sets across a single row (of course maintaining the option to then change individual image sets by using its dropdown). Perhaps when you + to create a new row it could ask if there was a metadata you'd like to fill that row with?

DavidStirling:
Resolves #4027

Per @ErinWeisbart's request, because this was indeed annoying and it's a simple fix. This PR makes selecting a metadata matching key automatically populate any other empty fields in that row with the same key (if it's available for that item).

I figure there's no need for a dedicated button, if the cells aren't empty they'll be left alone, and if there's an exactly matching key then users will in most instances want to use it.

### DIFF
--- a/cellprofiler/gui/module_view/_joiner_controller.py
+++ b/cellprofiler/gui/module_view/_joiner_controller.py
@@ -215,10 +215,11 @@ class JoinerController:
             joins.append(dict([(cn, "") for cn in self.column_names]))
         join = joins[row].copy()
         join[self.column_names[column]] = new_value
-        # Fill other empty fields if present.
-        for col in self.column_names:
-            if join[col] in ("", None) and new_value in self.v.entities[col]:
-                join[col] = new_value
+        if wx.GetKeyState(wx.WXK_SHIFT):
+            # Fill other empty fields if present.
+            for col in self.column_names:
+                if join[col] in ("", None) and new_value in self.v.entities[col]:
+                    join[col] = new_value
         joins[row] = join
         self.module_view.on_value_change(
             self.v, self.panel, self.v.build_string(joins), event
--- a/cellprofiler/gui/module_view/_joiner_controller.py
+++ b/cellprofiler/gui/module_view/_joiner_controller.py
@@ -215,6 +215,10 @@ class JoinerController:
             joins.append(dict([(cn, "") for cn in self.column_names]))
         join = joins[row].copy()
         join[self.column_names[column]] = new_value
+        # Fill other empty fields if present.
+        for col in self.column_names:
+            if join[col] in ("", None) and new_value in self.v.entities[col]:
+                join[col] = new_value
         joins[row] = join
         self.module_view.on_value_change(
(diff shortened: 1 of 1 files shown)


bethac07:
I'm not actually sure if this is a good idea? I can see use cases where this is nice (in general, where the number of images which have every "row" of your metadata filled out > the ones where only some are filled out (ie, if you have 4 channels and 1 mask)), but in some cases it's more-or-less-a-wash (if you have equal numbers of orig + illum images) and in some it's more of a pain (if you have > number of images that only partially fill the metadata rows than that fill all of them, like if you have one RGB image and multiple masks for each of several objects).  I think use cases 1 and 2 predominate, but we should stop and consider a second if we think messing up use case 3 is a problem.

DavidStirling:
I figure that in most cases with mixed sets the unwanted sets will probably lack the key and so shouldn't be auto-filled (e.g. illum). In other instances you're still able to go through and set those extra images to "None" or some other key. I guess we need to consider how common that case is?

bethac07:
Ah, I missed the part where if that key is missing for that image, it doesn't fill.  I withdraw my objection.

bethac07:
PS Nice work!

DavidStirling:
Actually you might be right after all. I think the Metadata keys list is populated based on absolutely every possible key, rather than being specific to the image sets those keys are applied to. This would be fine for widgets like CSV metadata pairing, but might potentially cause extra work when setting up illum channels.

If we think having to deselect keys is a major problem, one alternative might be to just add a right click menu option for "fill row with this key" or something. That said, for the vast majority of cases I expect auto-filling as this PR does should reduce the work needed.

DavidStirling:
Better solution: Autofill if the user is holding the shift key?

ErinWeisbart:
What about a right click on a cell gives the option to “populate row with same selection” or something like that?

DavidStirling:
I'm leaning more towards "hold ctrl when selecting to disable auto-fill" right now, because a right-click menu would be more complicated to implement.

bethac07:
Checkbox to autofill maybe?

DavidStirling:
A checkbox is doable, but I'm not sure how to position such a thing when it's already a large widget of arbitrary size? I think the same widget is also used in a few other modules, such as Metadata for CSV pairing.

bethac07:
Decided it's cool, we'll just make it be actively triggered by holding down Shift rather than passively doing something the user doesn't expect.  Nice job!

PR #4195, opened 2020-09-02
CheckListSubscriber improvements

### LINKED ISSUE #4190: Mac Multiselect boxes scroll up every few seconds, which is VERY annoying
**Describe the bug**
In Mac, multiselect boxes scroll up to the top every couple seconds, so if your image/object list is longer than 4-5, it is very annoying to keep having to scroll down to find the thing you care about.

OSX 14.6, Python 3.8.2, wx 4.1.0

DavidStirling:
Fixes #4190, there's now a 10 second refresh timeout.

I've also added a "Refresh List" entry to the right-click menu on these boxes, so that if another module has been enabled/disabled the user has a way to force an update.

While I was at it, I've added some logic to prevent the user from selecting entries that were detected as duplicates. These are already highlighted, but previously you could check them and they'd deselect on the next refresh. With no refresh it's clearer to disable checking altogether.


bethac07:
Tried on mac, looks great!

PR #4213, opened 2020-09-09
Move 3D plane controls to the Figure toolbar

DavidStirling:
Previously I made the 3D viewer use a Matplotlib slider to allow the user to control which plane is displayed, but this came with some positioning issues on multi-subplot figures and was generally a bit laggy since any change in the widget forced a redraw.

Since we figured out how to put arbitrary content into the toolbar, I've played around with having this control use a native wx slider in the toolbar instead of as part of the plot. 

New3DControl

This implementation is more consistent with our other UI elements, prevents slider duplication when working with multiple subplots, avoids layout issues and doesn't impact performance so much (since we aren't adding extra matplotlib axes). As a bonus the user can type the plane number they want, too.


bethac07:
Looks great! Can you rebase, then we can merge?

DavidStirling:
@bethac07 That should have fixed the conflicts, hopefully it still works!

bethac07:
Still looks good! Thanks.  Will merge.

PR #4231, opened 2020-09-10
Adjust Figure parameters to handle large numbers of subplots

### LINKED ISSUE #4228: CalculateIlluminationApply: Large # of displayed images causes individual images to be shrunk to almost nothing
**Describe the bug**
In CorrectIlluminationCalculate, if there is a large number of input images, the display window shrinks the individual images to a tiny size.

**To Reproduce**
1. Load a cell painting pipeline and point to collection of raw images plus illumination correction functions (ICFs). Alternately, mock up 5 channels worth of input images and corresponding ICFs
2. Run the pipeline to the CorrectIlluminationApply module, making sure to yield the display window.
3. The display window appears as the following:
image

**Expected behavior**
* Expected the display figure to look similar to that for v.2.2.0:
image
* Maximizing the window is a workaround but this was not needed previously.
* Not known if v3.1.9 was similarly affected.

 - OS: Windows 10
 - CellProfiler Version v4.0.2
 - Downloaded from the website

**Additional context**
* Not sure if other modules which can accept/display a large number of images are similarly affected.

DavidStirling:
Fix #4228

Figures with excessive numbers of subplots are still cramped, but I've made CellProfiler use a smaller font size when a row or column will contain more than 3 plots. This should leave the plot visible.

I've also tidied up the formatting for CorrectIlluminationApply's plot titles. It now rounds properly and uses a shorter label for the middle plot when we have a lot to display.

Manyplots

### DIFF
--- a/cellprofiler/gui/figure/_figure.py
+++ b/cellprofiler/gui/figure/_figure.py
@@ -150,6 +150,7 @@ class Figure(wx.Frame):
         self.subplot_menus = {}
         self.widgets = []
         self.mouse_down = None
+        self.many_plots = False
         self.remove_menu = []
         self.figure = matplotlib.pyplot.Figure(constrained_layout=True)
         self.figure.set_constrained_layout_pads(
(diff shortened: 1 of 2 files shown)


bethac07:
LGTM- they're still on the small size, but users can also still adjust the hspace and wspace on their own.  Nice job!

PR #4238, opened 2020-09-13
Add automatic update notifications

DavidStirling:
This PR adds a system to automatically check for and notify the user of updates when running frozen builds of CellProfiler.

A setting has been added to the preferences dialog (default:Enabled), and a "Help-->Check For Updates" menu option has been added to allow users with this function disabled to force an update check.

Once per week on app startup, this system will check the version of the latest GitHub release and show the notification below if they're running an older version. Date of the last update check is stored within the config key.

UpdatePrompt

I've added a bit of code to catch errors which could happen with the web request. If running from startup errors are printed to the console. If the user activated the check using the menu option instead they'll get a dialog with the relevant message, including if CellProfiler was up-to-date.

This system does assume the version will take the format "v0.0.0", I expect we won't be changing that anyway. Pre-releases and drafts are excluded from the update search.

Requires cellprofiler/core#51

### DIFF
--- a/cellprofiler/gui/app.py
+++ b/cellprofiler/gui/app.py
@@ -55,13 +55,14 @@ class App(wx.App):
         super(App, self).__init__(*args, **kwargs)
 
     def OnInit(self):
-        from .cpframe import CPFrame
-        from cellprofiler import __version__
         if platform.system() == "Windows":
             import locale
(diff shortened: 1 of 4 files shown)


DavidStirling:
As a bonus, this also helped me find what should be a fix for the image asset locale issues, per 8525f65

bethac07:
Great!

PR #4247, opened 2020-09-16
Pack pipelines into Sentry error reports from test mode

DavidStirling:
Requires cellprofiler/core#53

It's a bit messy, but this should provide pipeline data in Sentry errors logged from the main app thread. This will be super helpful for figuring out problem settings.

Sentry has some weird limits on container sizes which aren't immediately obvious. We override those which we can, but that shouldn't cause problems elsewhere since we don't send large amounts of context data to begin with.

### DIFF
--- a/cellprofiler/gui/pipelinecontroller.py
+++ b/cellprofiler/gui/pipelinecontroller.py
@@ -124,7 +124,8 @@ from cellprofiler_core.preferences import (
     set_show_analysis_complete_dlg,
     set_default_output_directory,
     get_max_workers,
-    set_default_image_directory, get_telemetry,
+    set_default_image_directory,
+    get_telemetry,
 )
 from cellprofiler_core.setting import ValidationError
 from cellprofiler_core.utilities.core.modules import (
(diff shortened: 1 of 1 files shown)


cellprofiler/gui/pipelinecontroller.py
```diff
@@ -3136,6 +3137,40 @@ def remote_debug(evtlist=None):
...
+                }
+                scope.set_context("Pipeline_Info", pipeline_info)
+                index = 1
+                buffer = 0
+                buffer_dict = {}
+                # Sentry hard limits context content to >~10000 characters per container.
+                # To avoid this we send in chunks if the pipeline is large.
+                for idx, part in enumerate(pipeline_parts):
+                    idx += 1
+                    keyname = f"Pipeline_{index:02d}"
```
0x00b1, line 3162: Is it possible to have multiple pipelines? 🤔

DavidStirling: For future reference to anyone reading - Sentry has a cap on how much data can be sent in a single context object, so we split long pipelines up into chunks.

Issue #4025, opened 2020-05-22
CellProfiler should check if an image is used anymore, and if not it should forget it


bethac07:
**Is your feature request related to a problem? Please describe.**
Sometimes pipelines create many intermediate images, which are only used to make the next intermediate image and never again downstream (for overlays, measurements, etc).  These use up a bunch of memory, sometimes causing CellProfiler to crash.

**Describe the solution you'd like**
It would be nice in analysis mode (not necessarily in test mode) if CellProfiler would check at the end of every module if any of the images it touched are never used again, and if so to delete them from memory.

PR #4253, opened 2020-09-18
Add option to trace assets needed or produced by a module.

DavidStirling:
Random evening project for y'all. Someone suggested this a while ago yet I sadly can't find the link, but this has been mentioned by a few users.

When working with very long pipelines it can become challenging to quickly find modules which create or use inputs/outputs from a module of interest. The mentality being "I've changed this smoothing module setting, so I need to find the Identify module which used the output image and tweak it too, but I have a dozen identify modules!".

To address this I've played around with adding another right-click menu option to the pipeline list when the user is not in Test Mode:

TracerMenu

On clicking this option, CellProfiler will display an icon next to any modules which created or used an image, object or measurement related to the targetted module. Changing any settings, modules or entering test mode will reset the display.

TracerDisplay

In this example, the IdentifyTertiaryObjects module was examined, and it needs images from IDPrimary and IDSecondary. It's output objects are then used by some of the Measure modules and it's own measurements are potentially used by the Export module.

Currently the scan for measurements doesn't specify the object which a measurement relates to. So if a filter module uses "AreaShape" it'll current tag all MeasureObjectSizeShape instances. Not sure how much of a problem that is, or quite how to fix it robustly.

We also don't inspect the NamesAndTypes module, because it's in a different list controller and that's a pain.

The logic used for this may eventually be helpful for #4025


bethac07:
Ho. ly.  Crap.  I CANNOT EXPRESS HOW MUCH I LOVE THIS.

AnneCarpenter:
Woweee! It looks very intuitive and useful!! My only question is the icon at the module in question was a little confusing. Maybe a simple dot or even nothing at all would be better? That way there are only two kinds of arrows happening which seems easier to immediately understand as inputs and outputs.

DavidStirling:
@AnneCarpenter That could certainly work, I toyed with a few icon ideas but wasn't too sure what would be best for the "source" module. I'm not the best pixel artist, but if others would prefer a dot or have other ideas I can have a go at making it!

DavidStirling:
@AnneCarpenter @bethac07 What do we think about this icon as an alternative marker for the "source" module?

DotMarker

AnneCarpenter:
perfection!

PR #4273, opened 2020-10-04
delete unused sentences

szcf-weiya:
these unused sentences might confuse people at the first glance, especially for the unexpected constant number 10.


DavidStirling:
Thanks for doing that @szcf-weiya, much appreciated!

PR #4281, opened 2020-10-08
Set multichannel plot to subplot_imshow_color in Resize module

votti:
I found that the resize module was still using the old `subplot_imshow` command that fails if a multichannel image has >3 channels for display. This should fix this.

Or was there a specific reason this used `subplot_imshow` only?

### DIFF
--- a/cellprofiler/modules/resize.py
+++ b/cellprofiler/modules/resize.py
@@ -429,11 +429,11 @@ resized with the same settings as the first image.""",
             )
         ):
             if multichannel:
-                figure.subplot_imshow(
+                figure.subplot_imshow_color(
                     0, i, input_image_pixels, title=input_image_name,
                 )
 
-                figure.subplot_imshow(
+                figure.subplot_imshow_color(
                     1, i, output_image_pixels, title=output_image_name,
                 )
             else:


DavidStirling:
Unfortunately using `imshow_color` then breaks the figure display when working with a color 3D image (since shape is ZXYC rather than XYC). We might want to add some code to have it use standard `imshow` if the pipeline is 3D? Otherwise we could probably revise handling of 3D color images but that risks breaking more modules than a local fix would.

votti:
What about checking the "Image.volumetric" attribute and dispatch the
appropriate method?
Maybe even within "subplot_imshow_color"?

Is there any specific reason why this wouldn't work?

Cheers!

DavidStirling:
Part of the issue is that `imshow_color`'s handling of images with more than 3 channels is to just discard the extra channels. You might also want to test what happens when you input an image with only 2 channels since looking at the code it assumes 3. I reckon that'd be why it wasn't used in Resize.

`image.volumetric` should work for detecting a 3D image, but you may also want to ensure 3 channels are present.

Handling of colour images is certainly something that could be improved, but I reckon that'd need a more extensive refactor than we have time for internally at the moment.

votti:
Okay I think I will try to first write some unittests for `subplot_imshow` for the different scenarios and then see if I find a solution that handles all of them.
Given that the `Image` class nowadays has both a `volumetric` as well as a `multichannel` attribute it does sound very feasible to find a solution that would handle everything.
Let's see.

bethac07:
Closed as inactive

PR #4285, opened 2020-10-13
Fix CombineObjects not recording object counts/positions.

DavidStirling:
Apparently the new module was broken as it didn't record object counts and locations properly. I'd thought that these would be deduced from the parent sets when adding a new object set, but apparently not. To fix this I've made CombineObjects be based off the Identify template and added the relevant measurement fetching functions.


0x00b1:
Ah, OK

PR #4289, opened 2020-10-14
Added ability to add port number to MySQL host.

dahlo:
Added a function that picks out the hostname and port number, if any, from the specified MySQL host.
Has to be in one of the following formats:
```
# IPv4 no port specified
192.168.1.10

# IPv4 with port specified
192.168.1.10:3306

# IPv6 no port specified
9001:0db8:85a3:0000:0000:8a2e:0370:7334

# IPv6 with port specified
[9001:0db8:85a3:0000:0000:8a2e:0370:7334]:3306
```

### DIFF
--- a/cellprofiler/modules/exporttodatabase.py
+++ b/cellprofiler/modules/exporttodatabase.py
@@ -349,9 +349,44 @@ def get_next_result(cursor):
         return None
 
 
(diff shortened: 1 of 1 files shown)


DavidStirling:
Seems to work well, thanks so much!

PR #4292, opened 2020-10-16
Docs update

ErinWeisbart:
added explicit mention of .cppipe


bethac07:
You're not really exporting the pipeline AS a project, I think that's confusing due to how we otherwise confuse project; can we clean that sentence up a little?  I think it's maybe more clear to explain that you're exporting just the list of steps and corresponding settings.

Thanks for kicking this off!

ErinWeisbart:
@bethac07 clarified further based on your comments.

bethac07:
Great, thanks!

PR #4295, opened 2020-10-21
Add "Faster" Costes mode to MeasureColocalization

DavidStirling:
We noticed that performance of the Costes measurements in MeasureColocalization was extremely bad when working with 16-bit images. While potentially more accurate than the CP3 implementation, enabling them took module execution time from 1s to ~40s, which isn't very practical.

In this PR I've added a third mode to supplement the "Fast" mode we introduced in CP4, termed "Faster" mode. This method implements a sort-of bisection search to find the correct threshold quickly. To be precise, we progressively refine the window of possible thresholds using `min + (5/6 * width)` of the window. This prioritises the higher candidate thresholds and in practice seems to prevent us missing the target in unusual images where R value can vary wildly towards the lower threshold values.

In my testing we get the same results as linear mode. Performance on 8-bit images is about the same as the other modes, whereas 16-bit images are finished in a fraction of the time.

I've also moved the original code into a function to unify the image and per-object modes.

There's probably still a better way of finding the Costes threshold, but this seemed like an improvement.

### DIFF
--- a/cellprofiler/modules/measurecolocalization.py
+++ b/cellprofiler/modules/measurecolocalization.py
@@ -130,7 +130,7 @@ F_COSTES_FORMAT = "Correlation_Costes_%s_%s"
 class MeasureColocalization(Module):
     module_name = "MeasureColocalization"
     category = "Measurement"
-    variable_revision_number = 6
+    variable_revision_number = 5
 
     def create_settings(self):
         """Create the initial settings for the module"""
@@ -1607,9 +1607,6 @@ Alternatively, you may want to disable these specific measurements entirely
             # Add costes mode switch
             setting_values += [M_ACCURATE]
             variable_revision_number = 5
-        if variable_revision_number == 5:
-            # Added "Faster" mode, no settings change
-            variable_revision_number = 6
         return setting_values, variable_revision_number
 
     def volumetric(self):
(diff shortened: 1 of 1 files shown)


bethac07:
In general, fine, and I trust your testing, but why are we bumping module version number if we don't have to?

DavidStirling:
Ha, good point... While we don't add new settings, we do add new options for existing settings which wouldn't be backwards-compatible. I figured that might merit a new version number, but I can revert that if we only care about entirely new settings?

bethac07:
We usually only bump version number for entirely new settings (and even then, only for settings that mess up the order of existing settings, aka I didn't catch it (don't remember if I reviewed or not) but I wouldn't have bumped from 4-to-5 when you added the last new setting because it didn't mess up the order, was just added at the end).  Bumping version number isn't going to change that it won't be backwards compatible- it will either be sad because wrong version number, or sad because of an invalid setting, but either way it will still fail.

If you don't mind reverting just that, that'd be great, if it's too big a hassle I can pull as is.

bethac07:
(To be clear, don't revert BACK to 4, at this point we've already bumped, just... don't bump past 5 right now).

DavidStirling:
I could agree that when we don't change the settings list we shouldn't bump version number, but I do reckon that when adding a new setting, even if it's the last in the list, we should always take the time to bump the number and add an upgrade function to explicitly add the correct new value. This also makes life much easier when the next big change comes around, since the upgrade function already handles the full list of settings correctly.

I'll pull the version number back to 5

bethac07:
Reasonable argument, just saying that's how we've decided when to bump in the past- extra settings at the end should still be handled correctly no matter what, but your way is undeniably safer.

Issue #4294, opened 2020-10-21
`print_groups` command line option fails with JSON serialization error


votti:
**Describe the bug**
I am using the '--print-groups' command line option to print the image groups from a batch file in CP4.

The following error presents itself:
```
Traceback (most recent call last):
  File "/usr/local/bin/cellprofiler", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.8/dist-packages/cellprofiler/__main__.py", line 179, in main
    print_groups(options.print_groups_file)
  File "/usr/local/lib/python3.8/dist-packages/cellprofiler/__main__.py", line 670, in print_groups
    json.dump(groupings, sys.stdout)
  File "/usr/lib/python3.8/json/__init__.py", line 179, in dump
    for chunk in iterable:
  File "/usr/lib/python3.8/json/encoder.py", line 429, in _iterencode
    yield from _iterencode_list(o, _current_indent_level)
  File "/usr/lib/python3.8/json/encoder.py", line 325, in _iterencode_list
    yield from chunks
  File "/usr/lib/python3.8/json/encoder.py", line 325, in _iterencode_list
    yield from chunks
  File "/usr/lib/python3.8/json/encoder.py", line 325, in _iterencode_list
    yield from chunks
  File "/usr/lib/python3.8/json/encoder.py", line 438, in _iterencode
    o = _default(o)
  File "/usr/lib/python3.8/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type int64 is not JSON serializable
``` 

**To Reproduce**
run `cellprofiler -c --print-groups=Batch.h5` on a previously generated batch file.

**Expected behavior**
The command line script just works :)

**Desktop (please complete the following information):**
 - Docker image: docker://cellprofiler/cellprofiler:4.0.6

**Additional context**
I guess the issue is that this that `Measurements.get_groupings` returns Numpy int64:
https://github.com/CellProfiler/CellProfiler/blob/f7d413adb22560cf9163dec3f1801bb5968d276a/cellprofiler/__main__.py#L668

According to stackoverflow one these would need to be 'manually' coerced to `int`

DavidStirling:
Looks like this was fixed by #4297

PR #4297, opened 2020-10-22
Fix `--print-groups` by coercing np.int64 to int

votti:
Json dumping used by `--print-groups` doesnt support the np.int64 returned by Measurement.get_groupings (Issue #4294).
This pull request coerces the image numbers to Python int before generating the json file..

I wasn't sure if this could also be better fixed within the Measurement.get_groupings function. But then it might be better internally  keep image numbers consistently as np.int64.

I now also covered `print_groups` with a test.

### DIFF
--- a/cellprofiler/__main__.py
+++ b/cellprofiler/__main__.py
@@ -667,7 +667,12 @@ def print_groups(filename):
 
     groupings = m.get_groupings(metadata_tags)
 
-    json.dump(groupings, sys.stdout)
+    # Groupings are np.int64 which cannot be dumped to json
+    groupings_export = []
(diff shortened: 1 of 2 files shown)


votti:
Hi!
I just want to check with you if you could consider merging this request or if there is anything I can do to improve the pull request.
In the end this issue is breaking a workflow I am maintaining and thus prevents me from releasing this workflow to CP4.

Thanks for any feedback!

DavidStirling:
Sorry for the delay, looks like this works well! Thanks!

votti:
Tiptop, super Sach 🎉  
Merci and cheers!

PR #4302, opened 2020-11-09
Add log normalisation option to Threshold

DavidStirling:
This pull adds a new setting to Threshold (and therefore IDPrimary/Secondary) which allows the input image to be log-transformed before calculating Otsu or MCE thresholds.

In CP2 (Otsu 2-class) and CP3 (Otsu 3-class) the old Centrosome implementation of these methods was hard-coded to do this transform, but this was inconsistent with the Skimage methods. In CP4 we switched to all-Skimage implementations, but this effectively lost the old transformation functionality. It turns out that in practice the transform is very useful when you need to detect particularly weak staining in the presence of a stronger stain, usually when identifying cell boundaries.

To resolve this I've added the transform as a setting within the module. When enabled the transforms are applied before using the Skimage functions like normal (rather than reverting to using Centrosome Otsu).

### DIFF
--- a/cellprofiler/modules/threshold.py
+++ b/cellprofiler/modules/threshold.py
@@ -1309,15 +1309,20 @@ staining.
             else:
                 setting_values += [centrosome.threshold.TM_OTSU]
             variable_revision_number = 10
+        used_log_otsu = False
         if variable_revision_number == 10:
             # Relabel method names
             if setting_values[3] == "RobustBackground":
                 setting_values[3] = TM_ROBUST_BACKGROUND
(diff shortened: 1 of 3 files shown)


bethac07:
Is there a reason this is only enabled for MCE and Otsu?

DavidStirling:
I limited it to MCE and Otsu because it seemed inappropriate for the other methods. Manual and Measurement certainly have no use for it, and RobustBackground is a bit more complex so I couldn't be confident what it'd do when transformed.

bethac07:
Sounds reasonable to me!  Thanks for adding the extra tests :)

PR #4303, opened 2020-11-10
Adding Conda environment file

marcosterland:
Hi,
I couldn't find a recent Conda Environment for CellProfiler 4.
This seems to be the minimal setup to get CellProfiler 4 working with Python 3.8 on recent Linux distributions.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,19 @@
+name: cp4
+
+channels:
+  - conda-forge
+  - anaconda
+  - bioconda
+  - defaults
+
+dependencies:
+  - python=3.8
+  - numpy
+  - matplotlib
+  - pandas
+  - mysqlclient
+  - openjdk
+  - gtk2
+  - wxpython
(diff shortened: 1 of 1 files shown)


bethac07:
Hey Marc,

Thanks for putting this together!  

We don't have any active conda users on the eng team at the moment, so we're a bit leery of providing this as an "official" instruction set since we likely won't know if/when things break to keep them up to date; I'm going to close this without merging, therefore, but I HAVE added these as instructions on our [Conda Installation Page](https://github.com/CellProfiler/CellProfiler/wiki/Conda-Installation#cellprofiler-40), and if you identify revisions that are needed you're welcome to edit there.  

Again, super grateful for your contribution, just want to make sure we're not "warrantying" anything we can't guarantee to keep up to date at the moment.

~Beth

marcosterland:
Ok great, that seems like a good place! :-)

PR #4307, opened 2020-11-16
Fix warning about using too many image sets with ExportToSpreadsheet.

DavidStirling:
Fixes some bugs from #4138

- Instead of checking whether we're in headless mode, the previous version was checking for whether wx is _installed_. This caused a crash when using headless mode since wx is present, but not loaded as an app.

- Made the function aware of whether images are grouped. If so, the limit of 500 sets is considered per-group, rather than per-experiment.

- Clicking "Yes" in response to the GUI prompt will no longer skip the `check_overwrite` prompt.

### DIFF
--- a/cellprofiler/modules/exporttospreadsheet.py
+++ b/cellprofiler/modules/exporttospreadsheet.py
@@ -726,6 +726,10 @@ desired.
         if workspace.measurements.has_groups():
             group_numbers = workspace.measurements["Image", "Group_Number", workspace.measurements.get_image_numbers()]
             max_image_set_len = max(numpy.bincount(group_numbers))
+        elif workspace.measurements.has_measurements("Image", "Group_Length", 1):
+            num_images = workspace.measurements.image_set_count
+            max_image_set_len = max(workspace.measurements.get_measurement(
+                "Image", "Group_Length", range(1, num_images + 1)))
         else:
             max_image_set_len = workspace.measurements.image_set_count
         if max_image_set_len > maximum_image_sets:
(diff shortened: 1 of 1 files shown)


bethac07:
I'm not sure breaking down by grouping working as expected- I tried to run a pipeline that is grouping by plate, well, and site (240 groups of 3), and I still got the error message, whether I was running all groups or only 1.  However, it was no longer a breaking error, so that's progress!

CSV and pipeline for testing attached- they're for the AdvancedSegmentation/QC image set.

[Archive.zip](https://github.com/CellProfiler/CellProfiler/files/5556720/Archive.zip)

DavidStirling:
That's strange, on my end it's registering that the groups exist and returning the correct value. Perhaps at some point we could take a look at what's happening when it gets to ExportToSpreadsheet L726?

DavidStirling:
@bethac07 I've added some additional conditions to try and resolve this. Looks like LoadData might not fully set up measurements dictating which groups are which, but it does provide the group lengths so we can use that.

bethac07:
I still think this isn't behaving in the way I want/expect - if the pipeline is not using groups, it should use however many sets are present, and if it IS, it should use however many total sets are present in _however many groups are being run at the time_ - aka if I'm running 720 groups of 1 image set each, it should still behave like I'm running 720 image sets (because I am), but if I have 720 groups of 1 image set in my batch file or my CSV and I'm ONLY EXECUTING ONE because of a -g flag, it should behave like I'm only running 1.  

That's a "nice-to-have" though that shouldn't stop us merging a fix for a bug that breaks stuff though; I'll merge and make an issue for the "new" functionality.

PR #4308, opened 2020-11-17
Adding codesign as part of the action workflow

alicelucas:
In the workflow, we now call two scripts (add-osx-certificate.sh and osx-codesign.sh) located in `distribution` that are going to (1) import the apple certificate from the secret and (2) sign the app with `codesign`. 
This is done after building the bundled app, but before zipping it up for upload.


bethac07:
Amazing, thanks so much @alicelucas, this has been on our to-do list for YEARS!

PR #4313, opened 2020-11-24
Untransform before correcting thresholds

### LINKED ISSUE #4312: Log transforming threshold tests against upper and lower bounds while logged
It should respect them AFTER transforming back to linear, not before.  To test, try thresholding the below image with 3 Class Otsu Foreground, Logged; it will auto calculate a threshold of 0.00227.  If you give it a lower bound of 0.003, nothing will happen; if you give it one of 0.1, the "final" threshold will be 0.00246; if you give it one of 0.5, it will be 0.00643, etc.  

[IXMtest_B18_s1_w155496334-1026-4449-B84E-F982A66DA60E.tif.zip](https://github.com/CellProfiler/CellProfiler/files/5591405/IXMtest_B18_s1_w155496334-1026-4449-B84E-F982A66DA60E.tif.zip)

DavidStirling:
Hopefully fixes #4312. Correction factors and bounds are now always applied after transforming back into the linear scale.

This also simplifies the individual thresholding functions somewhat.

@bethac07 Is this what you had in mind?

### DIFF
--- a/cellprofiler/modules/threshold.py
+++ b/cellprofiler/modules/threshold.py
@@ -870,18 +870,25 @@ staining.
             return self._correct_global_threshold(orig_threshold), orig_threshold, None
 
         elif self.threshold_scope.value == TS_GLOBAL or automatic:
-            th_corrected, th_original, th_guide = self.get_global_threshold(image_data, image.mask, automatic=automatic)
+            th_guide = None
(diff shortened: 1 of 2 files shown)


bethac07:
It is PRECISELY what I had in mind, thanks man.

PR #4315, opened 2020-11-30
RunImageJMacro module

alicelucas:
This is the beginning of a RunImageJMacro module in its simplest form. The way it currently works is the following:

- First, it is assumed that the macro expects a "directory" variable. This is the mechanism we chose to transfer images from CellProfiler to Fiji, and vice versa. The macro should use this "directory" variable when specifying the full path of an image to be read. It is also in this "directory" that the macro will write out images, which are then to be read by CellProfiler. The value of this variable is automatically assigned to a temporary directory created by CellProfiler.

- The user can select one or more images in the current workspace. For each of these images, the user inputs a filename expected by the macro. These image files are saved in the temporary directory created by CellProfiler. The macro expects these images to live in that directory and reads them under the specified filenames. 

- The user specifies the filename of the image(s) outputted by the macro. These are the filenames used by the macro when writing output images to the "directory" variable. Once ready, CellProfiler looks for these images in the created directory, reads them, and adds them to the workspace. 

- User can assign variable names and corresponding values expected from the macro script.

This module was tested on MacOS but still needs to be tested on other OS. 

This module's current functionality is a bit basic, but we expect to add more exciting features in the future!

### DIFF
--- a/cellprofiler/modules/__init__.py
+++ b/cellprofiler/modules/__init__.py
@@ -72,6 +72,7 @@ builtin_modules = {
     "rescaleintensity": "RescaleIntensity",
     "resizeobjects": "ResizeObjects",
     "resize": "Resize",
+    "runimagejmacro": "RunImageJMacro",
     "savecroppedobjects": "SaveCroppedObjects",
     "saveimages": "SaveImages",
     "shrinktoobjectcenters": "ShrinkToObjectCenters",
--- a/cellprofiler/modules/runimagejmacro.py
+++ b/cellprofiler/modules/runimagejmacro.py
@@ -234,33 +234,38 @@ should select the directory containing ImageJ-win64.exe (usually corresponding t
 
     def run(self, workspace):
         default_output_directory = get_default_output_directory()
-        tag = str(random.randint(100000, 999999))
+        tag = "runimagejmacro_" + str(random.randint(100000, 999999))
         tempdir = os.path.join(default_output_directory, tag)
         os.makedirs(tempdir, exist_ok=True)
+        try:
+            for image_group in self.image_groups_in:
+                image = workspace.image_set.get_image(image_group.image_name.value)
+                image_pixels = image.pixel_data
+                skimage.io.imsave(os.path.join(tempdir, image_group.output_filename.value), image_pixels)
 
-        for image_group in self.image_groups_in:
-            image = workspace.image_set.get_image(image_group.image_name.value)
-            image_pixels = image.pixel_data
-            skimage.io.imsave(os.path.join(tempdir, image_group.output_filename.value), image_pixels)
+            if self.executable_file.value[-4:] == ".app":
+                executable = os.path.join(default_output_directory, self.executable_directory.value.split("|")[1], self.executable_file.value, "Contents/MacOS/ImageJ-macosx")
+            else:
+                executable = os.path.join(default_output_directory, self.executable_directory.value.split("|")[1], self.executable_file.value)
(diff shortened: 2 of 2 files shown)


alicelucas:
I am attaching here an example project to use when testing. It makes use of a macro that computes the inverse of an image. 
[examplemacroproj.zip](https://github.com/CellProfiler/CellProfiler/files/5618698/examplemacroproj.zip)

bethac07:
Currently not working on Windows

bethac07:
I suspect it's the formatting of the cmd, when I print it, it looks like this:
`['Desktop\\\\Fiji.app\\ImageJ-win64.exe', '--headless', 'console', '--run', 'C:\\Users\\Administrator\\Downloads\\\\examplemacroproj\\InverseMacro.py', 'directory="C:\\Users\\Administrator\\773223", size=\'1\'']`

```
Failed to run module RunImageJMacro
  File "c:\users\administrator\documents\github\cellprofiler\cellprofiler\cellprofiler\gui\pipelinecontroller.py", line 3354, in do_step
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\site-packages\cellprofiler_core\pipeline\_pipeline.py", line 1292, in run_module
    module.run(workspace)
  File "c:\users\administrator\documents\github\cellprofiler\cellprofiler\cellprofiler\modules\runimagejmacro.py", line 253, in run
    subp = subprocess.call(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\subprocess.py", line 340, in call
    with Popen(*popenargs, **kwargs) as p:
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\site-packages\sentry_sdk\integrations\stdlib.py", line 190, in sentry_patched_popen_init
    rv = old_popen_init(self, *a, **kw)  # type: ignore
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\subprocess.py", line 854, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "c:\users\administrator\appdata\local\programs\python\python38\lib\subprocess.py", line 1307, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

alicelucas:
Updated PR to fix bugs for windows. Two issues were found:
- We weren't reading the absolute full path so Windows was not finding the appropriate macro directory. 
- Windows was unhappy with the double quotes in the directory argument. Replacing with single quotes did the trick. 

The module currently works both on Windows and MacOS.

DavidStirling:
Finally got around to testing this, nice job @alicelucas! Some thoughts:

- The macro/pipeline you provided as an example didn't work. It appears to expect a 'size' variable which wasn't present in the pipeline, despite it not being used. Adding that variable  as a pipeline setting got it working, so that's nice.

- When something goes wrong there seems to be a risk that the ImageJ process will hang. We might want to consider having an optional timeout.

- We're going to need some error handling for when the macro fails. Currently the user will see a "file not found error" since the file to import is absent. There should be a prompt informing the user that the macro did not produce the expected output. If we can get error messages in from ImageJ (could we log them to the CP console?) that'd be great.

- Alongside that, the macro call and image loading should be within a `try` block so that if something does go wrong the temp folders are still cleaned up.

- If we're going to use the default output directory as the temp directory, the generated temp folder id should be rejected if a folder with that name already exists. We don't want to accidentally wipe someone's data from plate '308101'.

When the macro and pipeline are in sync, it seems to work pretty well. Nicely done!

PR #4317, opened 2020-12-02
Improve colour image detection

DavidStirling:
Turns out the 3D viewer was still mistakenly thinking greyscale images are color. I need to do further testing with this, but I believe this PR will still be an improvement that gets us most of the way there.


DavidStirling:
Seems to work, or at least I've not found a 3D image which fails to display!

PR #4318, opened 2020-12-02
Exporting and importing pipelines as json

alicelucas:
In the process of adding the possibility of exporting/importing a pipeline as/from a json object, from CellProfiler's application window.


alicelucas:
When exporting the pipeline, the user can now select the *.json format as the export pipeline. When importing a pipeline, a json pipelines can be selected and loaded. See [this PR on core](https://github.com/CellProfiler/core/pull/56) for more information on how we handle the export and import of our json pipelines.

PR #4326, opened 2021-01-04
CallBarcodes

0x00b1:


bethac07:
I think I'd rather leave this as a plugin than move to the main repo, unless you feel strongly otherwise?  Happy to discuss it more :)

0x00b1:
@bethac07 I don't have a preference 😆

bethac07:
It's too weird and niche a thing to go in the main repository, I think.  THANK YOU for working on it, very very much appreciated, I just think we should put the updated version back into plugins if that's cool.

PR #4361, opened 2021-02-27
Fix OMERO login dialog for wxPython>=4.1.0

emilroz:
wxPython 4.1.0 introduces stricter checks for alignment flags, this causes the OMERO login dialog to fail:

image

Full error:

```
Traceback (most recent call last):
  File ".../cellprofiler/gui/pipelinecontroller.py", line 3365, in do_step
    self.__pipeline.run_module(module, workspace_model)
  File ".../CellProfiler4.venv/lib/python3.8/site-packages/cellprofiler_core/pipeline/_pipeline.py", line 1298, in run_module
    module.run(workspace)
  File ".../CellProfiler4.venv/lib/python3.8/site-packages/cellprofiler_core/modules/loaddata.py", line 1092, in run
    image = image_set.get_image(image_name)
  File ".../CellProfiler4.venv/lib/python3.8/site-packages/cellprofiler_core/measurement/_measurements.py", line 1557, in get_image
    image = matching_providers[0].provide_image(self)
  File ".../CellProfiler4.venv/lib/python3.8/site-packages/cellprofiler_core/image/abstract_image/file/_file_image.py", line 327, in provide_image
    self.__set_image()
  File ".../CellProfiler4.venv/lib/python3.8/site-packages/cellprofiler_core/image/abstract_image/file/_file_image.py", line 264, in __set_image
    self.cache_file()
  File ".../CellProfiler4.venv/lib/python3.8/site-packages/cellprofiler_core/image/abstract_image/file/_file_image.py", line 182, in cache_file
    rdr = get_image_reader(id(self), url=url)
  File ".../CellProfiler4.venv/lib/python3.8/site-packages/bioformats/formatreader.py", line 961, in get_image_reader
    rdr = ImageReader(path=path, url=url)
  File ".../CellProfiler4.venv/lib/python3.8/site-packages/bioformats/formatreader.py", line 580, in __init__
    self.rdr = get_omero_reader()
  File ".../CellProfiler4.venv/lib/python3.8/site-packages/bioformats/formatreader.py", line 522, in get_omero_reader
    omero_login()
  File ".../CellProfiler4.venv/lib/python3.8/site-packages/bioformats/formatreader.py", line 472, in omero_login
    __omero_login_fn()
  File ".../cellprofiler/gui/pipelinecontroller.py", line 790, in omero_login
    with cellprofiler.gui.omerologin.OmeroLoginDlg(
  File ".../cellprofiler/gui/omerologin.py", line 50, in __init__
    sub_sizer.Add(
[...]

PR #4362, opened 2021-02-27
Update wxPython to 4.1.1?

emilroz:
On MacOS wxPython 4.1.0 runs into some issues:

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x000000012ed78d0f, pid=73123, tid=0x0000000000000307
#
# JRE version: OpenJDK Runtime Environment (8.0_242-b08) (build 1.8.0_242-b08)
# Java VM: OpenJDK 64-Bit Server VM (25.242-b08 mixed mode bsd-amd64 compressed oops)
# Problematic frame:
# C  [libwx_osx_cocoau_core-3.1.4.0.0.dylib+0x6ad0f]  _ZN33wxMacCoreGraphicsPenBrushDataBase22CalculateShadingValuesEPvPKdPd+0x11f
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# An error report file with more information is saved as:
# /Users/emil/Git/CellProfiler/CellProfiler-glencoe/hs_err_pid73123.log
#
# If you would like to submit a bug report, please visit:
#   http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#
Abort trap: 6
```
This particular error is triggered after successful login to OMERO.

The error is fixed after upgrading wxPython to 4.1.1:
image

Together with https://github.com/CellProfiler/CellProfiler/pull/4361 and https://github.com/CellProfiler/prokaryote/pull/50 it should be possible to use CellProfiler 4.x with the latest OMERO.server versions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,7 +86,7 @@ setuptools.setup(
         "scikit-learn>=0.20",
         "scipy>=1.4.1",
         "six",
-        "wxPython==4.1.0",
+        "wxPython==4.1.1",
     ],
     license="BSD",
     name="CellProfiler",


DavidStirling:
wxpython 4.1.1 currently suffers from additional issues on OSX, which particularly break the module drag and drop functionality. Ultil that is fixed we can't unpin, so resolving the errors with the OMERO reader might be a better approach.

emilroz:
@DavidStirling OmeroReader is deprecated since version 5.3.0 (https://github.com/ome/omero-blitz/blob/master/src/main/java/loci/ome/io/OmeroReader.java#L86) so we should probably think about a different way of reading images directly from OMERO. I'm happy to propose a fully Python based workflow for python-bioformats if that's of interest.

We have already added a pure Python reader before https://github.com/CellProfiler/python-bioformats/pull/74. If that's ok with you I can propose changes to omero_login and other methods to "pythonize" them. It will add https://github.com/ome/omero-py and https://github.com/glencoesoftware/omero-reader as dependencies but will the benefit the users by giving them the flexibility to work with the OMERO.server version of their choosing simply by installing different omero-py versions.

DavidStirling:
@emilroz A pure python solution would actually be fantastic, please do feel free to propose changes! As far as dependencies go, anything which can be installed via `pip` or delivered within the builds shouldn't be a problem.

DavidStirling:
Unfortunately there's still no sign of wxPython 4.1.2, so we're going to be stuck on 4.1.0 for the time being. To my knowledge we don't have access to a local OMERO server with which to test this functionality, so there currently isn't much we can do on this end. Did you get a chance to look at the potential Python implementation?

PR #4367, opened 2021-03-11
combine objects in 3d and fix combine objects type bug

pearlryder:
Fixes bug identified in forum w/ combining objects: https://forum.image.sc/t/improper-combineobject-merging-in-cellprofiler/49631

Added 3D functionality to combine objects


cellprofiler/modules/combineobjects.py
```diff
@@ -116,15 +116,17 @@ def run(self, workspace):
...
         ), "Objects sets must have the same dimensions"
 
-        labels_x = objects_x.segmented.copy()
-        labels_y = objects_y.segmented.copy()
+        labels_x = objects_x.segmented.copy().astype("uint16")
+        labels_y = objects_y.segmented.copy().astype("uint16")
 
         output = self.combine_arrays(labels_x, labels_y)
-        output_labels = skimage.morphology.label(output)
+        output_labels = skimage.morphology.label(output).astype("uint16")
```
DavidStirling, line 129: Forcing type here shouldn't be necessary, did you find you needed this?

cellprofiler/modules/combineobjects.py
```diff
@@ -172,7 +175,7 @@ def display(self, workspace, figure):
         )
 
     def combine_arrays(self, labels_x, labels_y):
-        output = numpy.zeros_like(labels_x)
+        output = numpy.zeros_like(labels_x).astype("uint16")
```
DavidStirling, line 178: Type casting is definitely not necessary here. `zeros_like` will copy the dtype of the input, which you already set.

cellprofiler/modules/combineobjects.py
```diff
@@ -232,3 +251,6 @@ def get_measurements(self, pipeline, object_name, category):
 
     def get_measurement_columns(self, pipeline):
         return get_object_measurement_columns(self.output_object.value)
+
+    def volumetric(self):
+        return True
```
DavidStirling, line 256: Very nice!

DavidStirling (commented):
Looking very good! We might want to see if we can cut down the casting if we can get away with it, since it'll save on unnecessary work.

Fix for 3D looks perfect and works on my end, would it be worth adding some tests though?

bethac07:
Sure, we can plan to add some tests and remove some of the unnecessary casts.  I also realized we didn't update the documentation to say the it DOES support 3D, so add that to the to-do list!

cellprofiler/modules/combineobjects.py
pearlryder: You're right, it's not necessary -- the data type output from skimage.morphology.label is int64, so we were actually scaling back (IIUC)

cellprofiler/modules/combineobjects.py
pearlryder: Yep, agreed

pearlryder:
Tests are now ready for review

tests/modules/test_combineobjects.py
```diff
@@ -16,6 +16,11 @@ def image():
 
     return cellprofiler_core.image.Image(data)
 
+@pytest.fixture
+def image_volume():
```
0x00b1, line 20: Try and use structuring elements (or the random data function we wrote for skimage).

tests/modules/test_combineobjects.py
```diff
@@ -112,16 +142,29 @@ def test_one_object_first_image(
...
+            segment = numpy.zeros((10, 10, 10))
+            segment[2:4,2:4,2:4] = 1
+
+            objects_x_volume.segmented = segment
+
+            for method in merge_methods:
+                module.merge_method.value = method
+                module.output_object.value = method
+                module.run(workspace_volume)
+                assert (workspace_volume.get_objects(method).segmented == segment).all()
```
0x00b1, line 159: Use a `numpy.testing` function to compare arrays.

tests/modules/test_combineobjects.py
0x00b1: You could also put an image in the `tests/data` directory.

tests/modules/test_combineobjects.py
```diff
@@ -290,3 +444,28 @@ def test_overlap_segment(self, objects_x, objects_y, module, workspace):
...
+            assert len(merged.indices) == 2
+            expected_segment = numpy.zeros_like(segment_x)
+            # in 3D, the objects get segmented on a diagonal
+            expected_segment[1:5, 1:6, 1:5] = 1
+            expected_segment[1:5, 1:5, 5:6] = 1
+            expected_segment[1:5, 6:9, 4:5] = 2
+            expected_segment[1:5, 5:9, 5:6] = 2
+            expected_segment[1:5, 4:9, 6:9] = 2
+            expected_segment[5:9, 4:9, 4:9] = 2
+            assert (merged.segmented == expected_segment).all()
```
0x00b1, line 471: Missing newline.

tests/modules/test_combineobjects.py
```diff
@@ -290,3 +444,28 @@ def test_overlap_segment(self, objects_x, objects_y, module, workspace):
...
+            objects_y_volume.segmented = segment_y
+
+            module.merge_method.value = "Segment"
+            module.run(workspace_volume)
+            merged = workspace_volume.get_objects("merged")
+
+            assert len(merged.indices) == 2
+            expected_segment = numpy.zeros_like(segment_x)
+            # in 3D, the objects get segmented on a diagonal
+            expected_segment[1:5, 1:6, 1:5] = 1
```
0x00b1, line 465: If you're looking to improve your NumPy skills, investigate ways to rewrite this as:

`expression = [1, 1, 2, 2, 2, 2]`

tests/modules/test_combineobjects.py
pearlryder: If I understand the way that these tests are structured correctly, rather than random data, we define specific numpy arrays that we can predict how they should be combined in order to then compare the output from combineobjects (@DavidStirling can confirm, since I think he set up the structure in 2D that I'm following in 3D). I don't think we could use random shapes w/ this approach, but I may be missing something.

tests/modules/test_combineobjects.py
```diff
@@ -39,6 +51,13 @@ def objects_y():
 
     return segmented
 
+@pytest.fixture
+def objects_y_volume():
```
DavidStirling, line 55: These appear to be identical to `objects_x` and `objects_x_volume`, so we might just want to have a single objects_volume fixture. These scripts were originally based off another module's so this area could probably use some cleanup.

tests/modules/test_combineobjects.py
DavidStirling: In 2D we generate some very specific object sets to test that each of the different merge methods works as expected (covering complete overlap, partial overlap, multiple overlap and so on). To recreate this in 3D using the ball strel could be a good approach, just need to be careful about placement. I agree that random shapes probably isn't what we want.

pearlryder: We've decided to use the specific object sets. In the future, the ball strel approach would be an option for improvement

tests/modules/test_combineobjects.py
pearlryder: Done - now we test numpy arrays using `numpy.testing.assert_array_equal`

pearlryder:
PR is ready for review

DavidStirling:
Tests needed a small fix, but otherwise this seems to be working well. Nicely done!

PR #4372, opened 2021-03-29
Add optional tiff compression

bethac07:
Currently set to be on for new addition to the pipeline, off to upgraded pipelines.

### DIFF
--- a/cellprofiler/modules/saveimages.py
+++ b/cellprofiler/modules/saveimages.py
@@ -108,7 +108,7 @@ WS_LAST_CYCLE = "Last cycle"
 class SaveImages(Module):
     module_name = "SaveImages"
 
-    variable_revision_number = 15
+    variable_revision_number = 16
 
     category = "File Processing"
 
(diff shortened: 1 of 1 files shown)


cellprofiler/modules/saveimages.py
```diff
@@ -745,6 +759,9 @@ def save_image(self, workspace):
             if image.volumetric and self.file_format.value == FF_TIFF:
                 save_kwargs.update({"compress": 6})
 
+            if self.tiff_compress.value and self.file_format.value == FF_TIFF:
```
DavidStirling, line 762: Looks like the compression is still forced on volumetric images by the block above? If this is the case we should either hide the setting when in 3D mode or make compression optional on volumetric as well.

bethac07: I'm not sure how to hide a setting only in volumetric mode (or change its default only in volumetric mode)- to know if a pipeline is working on 3D images or not, you need the pipeline or the workspace, neither of which is passed into the settings by default.  Do you know a way to work around this? Otherwise, it's a bit messy, but we could say "this setting will be ignored on 3D images, for which compression is always used".

DavidStirling: Looks like you're right, the modules themselves only have access to workspace info in certain functions. With that in mind I'd say that leaving a note in the help text is fine. Alternatively we could default to 'on' when upgrading and make the setting dictate what to do even in volumetric.

We should probably merge the two `if` blocks anyway to avoid problems down the line, just in case we ever want to change compression method.

0x00b1:
merging, we can revist the conditional in the future

PR #4388, opened 2021-04-29
Add "Keep Removed Objects" option to FilterObjects, support CPA3 classifiers

DavidStirling:
This PR adds a new setting to FilterObjects, "Keep removed objects", which makes the module generate an object set from any objects which were excluded by the user's defined filter.

The rationale behind this is that users often want to group objects into positive and negative groups, but in the current release a second copy of the module is needed if the user wants to carry forward a set of objects that didn't pass the filter.

At least in this version, "removed" sets aren't generated from additional sets which the user has asked the module to relabel - it's only operating on the input objects. We could add this relatively easily, but I fear it might make the settings a bit too complicated if the user has to provide two names for each group. Thoughts appreciated.

I've also added support for CPA3-style classifiers and scalers. In doing so I've put in some proper validation when loading classifiers, previously we had no check for whether the necessary measurements were available in this mode.

### DIFF
--- a/cellprofiler/modules/filterobjects.py
+++ b/cellprofiler/modules/filterobjects.py
@@ -138,7 +138,7 @@ PO_ALL = [PO_BOTH, PO_PARENT_WITH_MOST_OVERLAP]
 class FilterObjects(ObjectProcessing):
     module_name = "FilterObjects"
 
-    variable_revision_number = 8
+    variable_revision_number = 9
 
     def create_settings(self):
         super(FilterObjects, self).create_settings()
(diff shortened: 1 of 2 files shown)


DavidStirling:
Found a problem with running classifiers in analysis mode. Please hold off on merging this for the moment.

DavidStirling:
Fixed, all good now

bethac07:
A couple things here- 
0) Needs a rebase now
1) You need to add your new objects to get_measurement_columns, or many other things will not work with them
2) Worth supporting FGB?
3) Worth updating the graph to show both sets of object counts (or at least specifying which objects the count is for) if keeping both classes?
image

DavidStirling:
I'll sort this out once ClassifyObjects is finished and merged, since we'll re-use the FGB support tools I added there.

Good point about the measurements, sorry I missed that!

Counts should be simple enough to add, it's just the difference between the two.

PR #4391, opened 2021-05-14
Add CPA classifier model support to ClassifyObjects

DavidStirling:
This PR introduces a third 'mode' to the ClassifyObjects module which allows the user to specify a machine learning `.model` file (arising from CellProfiler Analyst). The primary goal here is to mirror the functionality of FilterObjects' classifier mode, but with support for working with multiple classes at once.

When in this mode, CellProfiler will import the model and attempt to use it to classify objects in an image. This will record a per-object 'Classifier_Class' measurement containing the name of an object's assigned class, as well as 'Classifier_Probability_(ClassName)' columns reflecting the determined probability of an object belonging to each class. Probabilities may be useful for those who want to set a certain confidence threshold downstream, or for identifying areas of confusion.

The user also has the ability to extract individual classes as new child object sets. The current settings interface is below.

image

In terms of figure display, I've made it so that the module colours each object by class and then draws a table mapping class colours to names and counts.

image

This should be in a good enough state for testing/general usage. An expansion we might want to make would be to support .rules text files from FastGentleBoosting. I probably will also need to figure out some way of generating probabilities for FastGentleBoosting model files, since they're not sklearn-based.

Might also be good to add some tests, but beyond validating that the correct measurements are generated I'm not sure how to robustly test the classifiers when any bump in sklearn version risks invalidating a .model file.


bethac07:
I could not save when I did this with the provided simplemodel_scale.model (didn't try the others) and the attached pipeline, traceback below.
The issue seems to be with calculating the per-image mean, as the issue is avoided if you don't. Classifying with a model gives a class name that is text, whereas with other classifiers (left) you can see it just gives each class a 0 or 1.  ExportToSpreadsheet barfs when trying to take the mean of a text column.

You either need to add type-checking to ExportToSpreadsheet, or not use a text column.

image

```
Failed to complete post_run processing for module ExportToSpreadsheet.
  File "/Users/bcimini/Documents/GitHub/CellProfiler/core/cellprofiler_core/pipeline/_pipeline.py", line 1471, in post_run
    module.post_run(workspace)
  File "/Users/bcimini/Documents/GitHub/CellProfiler/CellProfiler/cellprofiler/modules/exporttospreadsheet.py", line 845, in post_run
    self.run_objects(object_names, workspace, first_group)
  File "/Users/bcimini/Documents/GitHub/CellProfiler/CellProfiler/cellprofiler/modules/exporttospreadsheet.py", line 901, in run_objects
    self.make_image_file(
  File "/Users/bcimini/Documents/GitHub/CellProfiler/CellProfiler/cellprofiler/modules/exporttospreadsheet.py", line 1122, in make_image_file
    agg_measurements = m.compute_aggregate_measurements(img_number, aggs)
  File "/Users/bcimini/Documents/GitHub/CellProfiler/core/cellprofiler_core/measurement/_measurements.py", line 1199, in compute_aggregate_measurements
    values = values[numpy.isfinite(values)]
TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

[classifyerror.cppipe.zip](https://github.com/CellProfiler/CellProfiler/files/6625730/classifyerror.cppipe.zip)

DavidStirling:
Eep, it took a while to get around the first instance of this error. I think we'd be best off adapting ExportToSpreadsheet to not crash on non-numeric columns, I don't think it's unreasonable to want text columns in your output sheets.

I think the text column is much more convenient to read and interpret, so I'd like to keep it if possible.

PR #4396, opened 2021-05-25
Make ExportToDatabase support multichannel images when generating CPA properties files

### LINKED ISSUE #3991: Generated CPA .properties files should properly handle colour images.
When generating a CellProfiler-Analyst .properties file with ExportToDatabase, CellProfiler currently assumes that all input images are single channel. As a result colour/multichannel images don't display properly in CPA out of the box.

Ideally the `image_channel_blend_modes` line should be included in this file, alongside appropriate channel colours, when the user is working with colour images.

DavidStirling:
Fixes #3991

This PR updates the ExportToDatabase CPA properties file template for CPA 3, so generated files will now contain all the new optional parameters (set to defaults).

I've also put together a system to automatically figure out appropriate channel number and colour settings when generating a properties file for multichannel images.

This proved to be a little complicated due to the way the files are generated. The CPA properties file contents are actually generated twice for some reason, in both the pre-run and post-run phases. In the former it looks like this is sometimes logged into the database itself as the Experiment_Properties table. We have no access to image information before the run starts so that'll default to assuming 1 channel per image as before. However, during the pre-run phase we do now have access to the list of images we want the properties file to reference, so we store that as an experiment measurement for later. During the run we then call back that list and store the number of channels we see in each image as a hidden measurement in the HDF5 file. Because of how measurements work I couldn't find an effective way to only determine that list once per run - image data isn't available in the post run phase, so it'll currently generate the list for every image set. It's quick so I doubt that's too much of a problem. We can call the stored channels list back during the post run and use it to populate the properties file properly.

In effect, the generated .properties file will now include `channels_per_image` and a correct list of `image_channel_colors`, but the properties table in a database will retain the old system of assuming 1 channel per image. Hopefully this is workable enough to make use of anyway, since we primarily use .properties files.

I've also switched the properties file generator to use f-strings, since it's easier to read and a bit faster.

### DIFF
--- a/cellprofiler/modules/exporttodatabase.py
+++ b/cellprofiler/modules/exporttodatabase.py
@@ -732,7 +732,8 @@ images. This option will do the following:
    **SaveImages** will be included.
 -  The CellProfiler image name will be used for the *image\_name* field.
 -  A channel color listed in the *image\_channel\_colors* field will be
-   assigned to the image by default order.
+   assigned to the image by default order. Multichannel images will be 
+   added as separate R, G and B channels.
 
 Select "*{NO}*" to specify which images should be included or to
 override the automatic values.""".format(
@@ -1405,7 +1406,9 @@ Enter a name for the specified image.""",
                 doc="""\
 *(Used only if creating a properties file and specifying the image information)*
 
-Enter a color to display this channel.""",
+Enter a color to display this channel.
+
+Multichannel images will use this colour for all 3 image components""",
             ),
         )
 
--- a/cellprofiler/modules/exporttodatabase.py
+++ b/cellprofiler/modules/exporttodatabase.py
@@ -2339,6 +2339,12 @@ available:
         if pipeline.test_mode:
             return True
 
+        # Clear out any redundant records from previous runs.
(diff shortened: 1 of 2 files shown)


bethac07:
Can you put a couple tiny documentation bits here?  Then good to pull.

At 726, explain the default behavior (multichannel images will be added as RGB, otherwise they go in channel order, etc)

At line 1408, specify that if multichannel image is selected, this color is used for all of the channels of that image (I tested and that is what happens)

DavidStirling:
Added those, thanks

cellprofiler/modules/exporttodatabase.py
```diff
@@ -1405,7 +1406,9 @@ def add_image_group(self, can_remove=True):
                 doc="""\
 *(Used only if creating a properties file and specifying the image information)*
 
-Enter a color to display this channel.""",
+Enter a color to display this channel.
+
+Multichannel images will use this colour for all 3 image components""",
```
bethac07, line 1411: Consistent spelling please :)

bethac07: (sorry, was written mostly by damned dirty americans ;) )

DavidStirling: Is fixed

Issue #4373, opened 2021-04-02
Incorporate the declump objects plugin into the CellProfiler Watershed module


pearlryder:
**Is your feature request related to a problem? Please describe.**
The DeclumpObjects plugin is especially useful for 3D segmentation. Since the functionality is so useful, incorporating it into the CellProfiler Watershed module would make it more accessible for users. 

**Describe the solution you'd like**
Add an option to perform declumping on objects generated by the Watershed module (akin to the optional declumping in IdentifyPrimaryObjects)

**Describe alternatives you've considered**
Users can import it as a plugin now. This request is an enhancement to streamline workflows.

**Additional context**
Feature discussed w/ @bethac07

PR #4397, opened 2021-05-27
Issues/4373

pearlryder:
This PR closes issue #4373 - it incorporates optional advanced settings for the Watershed module to allow declumping using either a shape-based or an intensity-based method. Tests are included.

### DIFF
--- a/cellprofiler/modules/watershed.py
+++ b/cellprofiler/modules/watershed.py
@@ -405,6 +405,7 @@ the image is not downsampled.
 
         x_data = x.pixel_data
 
+        # watershed algorithm for distance-based method:
         if self.operation.value == O_DISTANCE:
             original_shape = x_data.shape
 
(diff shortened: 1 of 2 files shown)


cellprofiler/modules/watershed.py
```diff
@@ -83,6 +85,25 @@
        Higher values result in more regularly-shaped watershed basins. 
        
        .. _compact watershed: http://scikit-image.org/docs/0.13.x/api/skimage.morphology.html#r395
+
+Selecting *Advanced Settings* will split the detected objects into smaller objects based on a seeded watershed method
```
bethac07, line 89: Could you add a line at the end here acknowledging that this code comes from the DeclumpObjects plugin by Madison Bowden?

pearlryder: Oh yes, of course! I've since added attribution to Madison w/ my most recent commit.

PR #4399, opened 2021-06-01
RunImageJMacro enhancements

DavidStirling:
Two changes in this PR:

- I've added a 'debug mode' setting which prevents CellProfiler from deleting the temporary directory used to pass data to ImageJ. This only applies in test mode and isn't saved into the pipeline at all, since otherwise you'd risk making a mess. The user needs to explicitly enable it within the module settings.

- I've made it so that CellProfiler attempts to capture and display error messages from ImageJ by attaching them to the error trace. These required some cleanup because the logs are far from tidy, but I think I have something functional:

image

### DIFF
--- a/cellprofiler/modules/runimagejmacro.py
+++ b/cellprofiler/modules/runimagejmacro.py
@@ -44,7 +44,7 @@ from cellprofiler_core.setting.do_something import DoSomething, RemoveSettingBut
 from cellprofiler_core.setting._settings_group import SettingsGroup
 from cellprofiler_core.setting import Divider, HiddenCount, Binary
 from cellprofiler_core.setting.subscriber import ImageSubscriber
-from cellprofiler_core.preferences import get_default_output_directory
+from cellprofiler_core.preferences import get_default_output_directory, get_headless
 
 import random
 import skimage.io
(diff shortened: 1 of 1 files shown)


bethac07:
Have you confirmed that having a setting in `visible_settings` but not `settings` allows you to open previously-saved-pipelines ok? I just don't recall ever doing that otherwise.  If you have, ok by me for you to merge.

DavidStirling:
As far as I could tell, it works just fine. If I understand things correctly the `settings` method just needs to provide a list of settings which will be assigned values during pipeline loading/saving. `Visible_settings` is only used when generating the module display. There are objects such as dividers and separators which exist within `prepare_settings` but are never saved, although they don't get values associated with them. My reasoning was that, since we don't want to persist the setting, having a fresh copy generated with the default value each time you load the pipeline won't be an issue.

bethac07:
Yup, that aspect seems to work ok.

Can you add a line to the error message stating - "you should delete the files in `directory`, as CellProfiler will not?"

Also, if the run succeeds, and no error message is generated, how will folks know where the files are to delete? If you're running in built mac you don't have a console.

DavidStirling:
> Also, if the run succeeds, and no error message is generated, how will folks know where the files are to delete? If you're running in built mac you don't have a console.

Good question, temp folders should be in the default output directory. I considered making a wx popup but thought that'd be a bit annoying. Would we rather have that?

Another option would be to only abandon cleanup if something went wrong? That'd prevent successful runs from leaving a folder, but I figured people might want to look at their intermediates during debugging when things do work.

bethac07:
I think in this case we want (a bit) annoying.

bethac07:
```
  File "/Users/bcimini/Documents/GitHub/CellProfiler/CellProfiler/cellprofiler/gui/pipelinecontroller.py", line 3363, in do_step
  File "/Users/bcimini/Documents/GitHub/CellProfiler/core/cellprofiler_core/pipeline/_pipeline.py", line 1298, in run_module
    module.run(workspace)
  File "/Users/bcimini/Documents/GitHub/CellProfiler/CellProfiler/cellprofiler/modules/runimagejmacro.py", line 322, in run
    if not get_headless():
NameError: name 'get_headless' is not defined
```

image

bethac07:
@DavidStirling can you confirm my slightly longer message looks ok on Windows?

DavidStirling:
Looks fine, not sure how my get_headless import went missing - thanks for fixing that!

PR #4407, opened 2021-06-14
FilterObjects CPA classifier support and 'keep removed objects' option.

DavidStirling:
Continued and rebased from #4388.

---

This PR adds a new setting to FilterObjects, "Keep removed objects", which makes the module generate an object set from any objects which were excluded by the user's defined filter.

The rationale behind this is that users often want to group objects into positive and negative groups, but in the current release a second copy of the module is needed if the user wants to carry forward a set of objects that didn't pass the filter.

At least in this version, "removed" sets aren't generated from additional sets which the user has asked the module to relabel - it's only operating on the input objects. We could add this relatively easily, but I fear it might make the settings a bit too complicated if the user has to provide two names for each group. Thoughts appreciated.

I've also added support for CPA3-style classifiers and scalers. In doing so I've put in some proper validation when loading classifiers, previously we had no check for whether the necessary measurements were available in this mode.

### DIFF
--- a/cellprofiler/modules/filterobjects.py
+++ b/cellprofiler/modules/filterobjects.py
@@ -649,6 +649,27 @@ value will be retained.""".format(
                     % self.rules_file_name.value,
                     self.rules_file_name,
                 )
+            features = []
+            for feature_name in self.get_classifier_features():
+                feature_name = feature_name.split("_", 1)[1]
+                if feature_name == "x_loc":
+                    feature_name = M_LOCATION_CENTER_X
+                elif feature_name == "y_loc":
+                    feature_name = M_LOCATION_CENTER_Y
+                features.append(feature_name)
+            available_features = set([col[1] for col in pipeline.get_measurement_columns(self) if col[0] == self.x_name.value])
+            for feature in features:
+                if feature not in available_features:
+                    raise ValidationError(
+                        (
+                            "The classifier %s, requires the measurement, %s "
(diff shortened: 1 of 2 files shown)


bethac07:
```
Failed to run module FilterObjects
  File "/Users/bcimini/Documents/GitHub/CellProfiler/CellProfiler/cellprofiler/gui/pipelinecontroller.py", line 3363, in do_step
  File "/Users/bcimini/Documents/GitHub/CellProfiler/core/cellprofiler_core/pipeline/_pipeline.py", line 1298, in run_module
    module.run(workspace)
  File "/Users/bcimini/Documents/GitHub/CellProfiler/CellProfiler/cellprofiler/modules/filterobjects.py", line 716, in run
    indexes = self.keep_by_class(workspace, src_objects)
  File "/Users/bcimini/Documents/GitHub/CellProfiler/CellProfiler/cellprofiler/modules/filterobjects.py", line 1155, in keep_by_class
    target_class = classifier.classes_[target_idx]
AttributeError: 'Rules' object has no attribute 'classes_'
```
I also got this, even though that feature definitely should be there 
image

DavidStirling:
>  'Rules' object has no attribute 'classes_'

I see what's happening, I forgot that FilterObjects handles Rules and Classifiers with completely seperate pathways (unlike ClassifyObjects). Will try to fix that tomorrow.

DavidStirling:
@bethac07 Hoping this will work now!

PR #4469, opened 2021-10-01
Update FindMaxima

bethac07:
This pull request adds 3D functionality to the display for FindMaxima.

Thanks @alicelucas, @ErinWeisbart, @NasimJ, @bdiazroh, and @rsenft1 for hanging out and working on this together!

### DIFF
--- a/cellprofiler/modules/findmaxima.py
+++ b/cellprofiler/modules/findmaxima.py
@@ -161,12 +161,13 @@ maxima are within objects of interest."""
         else:
             raise NotImplementedError("Invalid background method choice")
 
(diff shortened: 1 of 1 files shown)


callum-jpg:
The display for this module now works in 3D.

In this PR I've also added an additional change. By default, the proposed maxima will be labelled with `scipy.ndimage.label`. This is because if the maxima are to be used as markers in watershed they need to have unique labels. That said, if a user does not want to label the maxima they can then select the color to display them in the CellProfiler GUI window.

gnodar01 (approved):

PR #4478, opened 2021-11-03
Fix issue with downsampling when advanced watershed features are on

rsenft1:
In watershed.py, resize x_data back up on line 459 to fix mismatch later (line 565) when watershed is generated from distance and downsampling is on (>1).

### DIFF
--- a/cellprofiler/modules/watershed.py
+++ b/cellprofiler/modules/watershed.py
@@ -456,6 +456,10 @@ the image is not downsampled.
                 )
 
                 y_data = numpy.rint(y_data).astype(numpy.uint16)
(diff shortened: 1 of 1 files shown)


rsenft1:
In case anyone in the future s curious about how resizing vs. recreating the mask affects the behavior of the advanced watershed (and how this compares to the basic watershed option), here are some examples: 

**Watershed BASIC downsample = 4**
CropBlue

**Watershed ADVANCED (recreate mask), downsample = 4 (76 objects)**
CropBlue

**Watershed ADVANCED (resize mask), downsample = 4 (76 objects)**
CropBlue

0x00b1:
LGTM

PR #4483, opened 2021-11-22
Reduce the memory usage of MeasureObjectNeighbors

LukeOZN:
Hi!

In using CellProfiler for large images (~8GB) we've found we run out of memory even on our cluster's 1.5TB nodes. The cause of the problem is in MeasureObjectNeighbors. Specifically at the point of the code modified in this PR, the original allocates 4 arrays: i, j, distance_matrix and order. Each of these arrays is of size nobjects*nneighbors. For us we have ~200000 objects. For 8 byte elements, this results in memory usage on the order of 4 * 200000 * 200000 * 8 * 10^-9 = 1280GB. But of these four arrays, the remaining code only makes use of the first three columns of order (the nearest neighbors). We have modified the code to only allocate the memory that is needed, which for the data mentioned above should not need to allocate over 1GB.

This code also runs faster than the original as we don't do a full sort on the distances, and instead find just the closest three with argpartition (N^2 rather than N^2*logN complexity). One might worry the explicit Python loop would slow things down, but in testing with random arrays of various sizes we've found about a 10x speed improvement.

The replacement code should be functionally equivalent to the original and in testing on our pipelines we are now able to segment much larger images.

### DIFF
--- a/cellprofiler/modules/measureobjectneighbors.py
+++ b/cellprofiler/modules/measureobjectneighbors.py
@@ -393,7 +393,7 @@ previously discarded objects.""".format(
             # order[:,1] should be the nearest neighbor
             # order[:,2] should be the next nearest neighbor
             #
-            order = numpy.zeros((nobjects, min(nobjects, 3)), dtype=numpy.uint32)
+            order = numpy.zeros((nobjects, min(nneighbors, 3)), dtype=numpy.uint32)
             j = numpy.arange(nneighbors)
             # (0, 1, 2) unless there are less than 3 neighbors
             partition_keys = tuple(range(min(nneighbors, 3)))
--- a/cellprofiler/modules/measureobjectneighbors.py
+++ b/cellprofiler/modules/measureobjectneighbors.py
@@ -393,13 +393,13 @@ previously discarded objects.""".format(
             # order[:,1] should be the nearest neighbor
             # order[:,2] should be the next nearest neighbor
             #
-            order = np.zeros((nobjects, min(nobjects, 3)), dtype=np.uint32)
-            j = np.arange(nneighbors)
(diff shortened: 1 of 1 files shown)


0x00b1:
LGTM

bethac07:
Hi,

First of all, sorry for the delay in review!

A couple of issues before we can pull this- 
1) Can you un-alias `np` back to `numpy`?
2) 4 tests are failing for measureobjectneighbors with these changes- see below. We won't be able to merge this code until all tests are passing.  If you run into issues, let us know, and hopefully we can help out!

```
tests/modules/test_measureobjectneighbors.py::test_neighbors_zeros
tests/modules/test_measureobjectneighbors.py::test_two_neighbors
tests/modules/test_measureobjectneighbors.py::test_different_neighbors_touching
tests/modules/test_measureobjectneighbors.py::test_small_removed
```

bethac07:
Ah, I found the issue - if you switch the below, all tests now pass.
`order = np.zeros((nobjects, min(nobjects, 3)), dtype=np.uint32)`
to
`order = numpy.zeros((nobjects, min(nneighbors, 3)), dtype=numpy.uint32)`

LukeOZN:
Sorry, as way of explanation I made these changes on v3.1.9 which we use and then checked if the current master branch was running the same code with the memory issues. I didn't spot that you'd switched from using the np alias to just using numpy in the newer version.

Have made the changes you requested, let me know if there is anything else.

bethac07:
Hi @LukeOZN ; it doesn't look like this change is included, can you please do so? Otherwise tests are failing. Thanks! https://github.com/CellProfiler/CellProfiler/pull/4483#issuecomment-1016657915

LukeOZN:
Included.

bethac07:
Awesome, thanks so much!

PR #4499, opened 2022-02-01
Updated marching_cubes function name

### LINKED ISSUE #4497: Update marching_cubes
We currently have `marching_cubes_classic` and `marching_cubes_lewiner`, which are deprecated in skimage and gone in 0.19+; we need to upgrade to marching_cubes(method='whichmethod') instead. (It looks like what used to be called "classic" is now referred to as "lorenson")

https://scikit-image.org/docs/0.19.x/api/skimage.measure.html?highlight=marching#skimage.measure.marching_cubes

callum-jpg:
Resolves #4497


bethac07:
Tests passing on my end. Nice job!

PR #4504, opened 2022-02-09
Resolves #4490 by adding in a description of correspondence vectors for MeasureTexture

### LINKED ISSUE #4490: Add the angle correspondences to the MeasureTexture documentation
Every year or so I have to look this up, which means it probably is needed there somewhere - in fact, I was pretty sure we HAD done it but evidently we (*cough* **I** *cough cough*) forgot.

_00 = horizontal (0 degrees)
_01 = diagonal (45 degrees) - down
_02 = vertical (90 degrees)
_03 = anti-diagonal (135 degrees) - down

rsenft1:


cellprofiler/modules/measuretexture.py
```diff
@@ -84,6 +87,39 @@
...
       value to InfoMeas1 and is on a different scale.
 
+**Note**: each of the above measurements are computed for different 
+'directions' in the image, specified by a series of correspondence vectors. 
+These are indicated in the results table in the *scale* column as n_00, n_01,
+n_02... for each scale *n*. In 2D, the directions and correspondence vectors *(y, x)* 
+for each measurement are given below:
+
+- _00 = horizontal, 0 degrees   (0, 1)
+- _01 = diagonal, 135 degrees   (1, 1)
```
bethac07, line 97: Can we add the diagonal directions here from mahotas? I think that adds an extra layer of intuitability

cellprofiler/modules/measuretexture.py
```diff
@@ -84,6 +87,39 @@
...
+- (1, 0,-1)
+- (0, 1,-1)
+- (1, 1,-1)
+- (1,-1,-1)
+
+In this case, an image makes understanding their directions easier. 
+Imagine the origin (0, 0, 0) is at the upper left corner of the first image
+in your z-stack.
+
+|MT_image0|
```
bethac07, line 121: How tricky is it to edit the image? I feel like the angle it's at right now makes it a little challenging for me personally to interpret, it might be nice to try a couple rotations but I don't know how doable that is. (Also, should we credit how we made the image?)

rsenft1: I've added a credit in for the website I used to make it (GeoGebra) We can rotate! If helpful, I could include a few views of the same image: 
image
image
We can also show the XY plane and/or change up the colorcoding, though I personally get a 'too much' vibe from this: 
image

PR #4510, opened 2022-02-17
fixed 3d cmap

### LINKED ISSUE #4506: ConvertImageToObjects 3D cmap limitation
For images that contain over 255 masks in 3D view, the colourmap is reset at every z-plane. The underlying objects remain the same. A colourmap should be generated ahead of time and passed to the subplot.display.

callum-jpg:
Fixes #4506

Originally, when an image is detected to have >255 objects a new matplotlib normalization is calculated up to the image.max(). However, for 3d images this leads to a new normalization to be calculated for each plane that has >255 objects and is plane specific, meaning that each plane has a different colormap applied. These changing colormaps are purely cosmetic, but can be seen in any module that uses the 3d viewer.

This fix first determines the maximum object value for an entire 3d stack. Then, it checks whether the image is 3d and if so will normalize based on the entire 3d image maximum rather than each individual plane maximum. 2d colormap generation remains unchanged.

### DIFF
--- a/cellprofiler/gui/figure/_figure.py
+++ b/cellprofiler/gui/figure/_figure.py
@@ -2029,6 +2029,7 @@ class Figure(wx.Frame):
         in_range = (image.min(), image.max())
         image = image.astype(numpy.float32)
         if self.dimensions == 3:
+            orig_image_max = image.max()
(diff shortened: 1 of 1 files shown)


bethac07:
🎉

PR #4511, opened 2022-02-17
WIP: Add Flatpak support

ximion:
Hi!
[Flatpak](https://flatpak.org/) is a widely supported bundling system on Linux, which makes it possible to "freeze" and app with its dependencies, deploy it easily to pretty much any Linux distribution, and allow it to run even on more modern distributions (or older ones) with ease.

I created this recipe to bundle CellProfiler for use in our lab and easy deployment on multiple different Linux desktops.
To test this yourself, you can do the following (on Ubuntu/Debian, on other distributions the steps to install Flatpak are different):

#### 1. Install Flatpak
```bash
sudo apt install flatpak flatpak-builder
sudo systemctl reboot # restarts the computer
```
#### 2. Configure Flatpak to use Flathub for SDKs/Runtimes
```bash
sudo flatpak remote-add --if-not-exists flathub https://flathub.org/repo/flathub.flatpakrepo
```
#### 3. Build the bundle
```bash
cd CellProfiler/flatpak
./autobuild-bundle.sh
```
One drawback of Flatpak is that we now have to build almost *everything* from scratch - this will take a while! In future, this step can hopefully be improved a bit to not be quite so intense.
#### 4. Install the application!
You can open the resulting `CellProfiler.flatpak` in KDE Discover or GNOME Software for a GUI to install it, or install it via the command-line:
```bash
flatpak install ./CellProfiler.flatpak
```
You can then just launch CellProfiler by clicking its icon on the menu!

The `CellProfiler.flatpak` file could then be shipped alongside the Windows and macOS builds of CellProfiler, and should work reliably on all modern Linux distributions. Currently, the script builds the Git master versions, but creating release builds is of course possible as well.

This PR is a work-in-progress, and it may actually be the case that it is better to submit this app and the Flatpak manifests and auxiliary data in this PR directly to [Flathub](https://flathub.org/), an open repository of bundled apps, and have it built and maintained there.
In any case though, I wanted to make this work available to more people who might like to use it as early as possible. I also wanted to show it to the CellProfiler maintainers in case they are interested in providing the Flatpak bundle or repo themselves (if at all), and get some feedback on whether they would have any interest in helping with maintaining this for Flathub (as this is of course only useful if it is adjusted to future updates of CellProfiler, so merging this would require some maintenance, even though the bundles will work for a long time once they are built). If Linux support like this isn't a priority or maintaining this upstream is not seen as feasible, I would just maintain this externally as long as we use it in our lab.
Cheers,
    Matthias


ximion:
Well, it seems to work very well for our lab, the only things it needs is:
* Some more people to test it and
* someone to maintain it once merged.

Especially the latter may be difficult...

gnodar01:
Ah, sorry hadn't realized it was in working condition. Thanks for doing this! In that case, we'd just need to move it to the [CellProfiler/distribution](https://github.com/CellProfiler/distribution) repo where we build our other packages. We can probably spin up a VM to test, but if it works now, we can maintain it until/unless we deem it too cumbersome in the future.

PR #4538, opened 2022-04-21
Set up initial library example

bethac07:
Initial example of what the library divide might look like. Tests are passing.

Thoughts 
* Work will need to be done to actually make this ie pypi publishable; that has not been done here.
  * Will also need to think about how to version/release this - tandem with CellProfiler makes the most sense to me but open to other ideas.
* Is this the structure we want? It seems to me to fall in with what we discussed but open to suggestions/revisions.


bethac07:
Closed in favor of #4554

Issue #4529, opened 2022-04-01
More customization options for DisplayDataOnImage


ErinWeisbart:
It'd be nice to have some extra customization options to the way that numbers print, especially for viewing of small numbers.
Including:
- remove leading 0 (or 0 to the left of the decimal point, or however we want to phrase it)
- use scientific notation (potentially with option to set exponent)
- multiply all values by some number (less important to implement, and I know it can be done by adding in a CalculateMath module first, but could be handy to have built into this module)

ErinWeisbart:
And while we're at it, could we add in the option of a couple fonts to pick from and/or the ability to make the print bold?

bethac07:
Resoved by #4539

PR #4539, opened 2022-04-21
Display data on image enhancement

callum-jpg:
Added additional options to DisplayDataOnImage, as requested by #4529. Though the requested features for multiplication of a value or removing 0's to the left of the `.` were not added.


bethac07:
@callum-jpg Pardon my bad memory, but you checked already that this works ok on Windows, yes? I think I recall this conversation but want to confirm before pulling. We should definitely in any case make sure to check this functionality out in any 4.2.2 release candidate.

bethac07:
Great, thanks!

PR #4545, opened 2022-05-03
Added additional save features to SaveCroppedObjects

callum-jpg:
Added the option to add input filenames to output crops and also the possibility to save output crops into nested folders.

### DIFF
--- a/cellprofiler/modules/savecroppedobjects.py
+++ b/cellprofiler/modules/savecroppedobjects.py
@@ -7,9 +7,12 @@ assigned the value from the input image. All other pixels (i.e., background pixe
 objects) are assigned the value 0. The dimensions of each image are the same as the original image. Multi-channel color
 images will be represented as 3-channel RGB images when saved with this module (not available in 3D mode).
 
-The filename for an exported image is formatted as "{object name}_{label index}.{image_format}", where *object name*
-is the name of the exported objects, *label index* is the integer label of the object exported in the image (starting
-from 1).
+The filename for an exported image is formatted in one of two ways. 
+By default, when the *Prefix saved crop image name with input image name* option is enabled, the format is
+"{input image name}_{object name}_{label index}.{image_format}",
+and when disabled the format is, "{object name}_{label index}.{image_format}", 
+where *object name* is the name of the exported objects, 
+and *label index* is the integer label of the object exported in the image (starting from 1).
 
 |
 
--- a/cellprofiler/modules/savecroppedobjects.py
+++ b/cellprofiler/modules/savecroppedobjects.py
@@ -80,8 +80,27 @@ The choices are:
             value=DEFAULT_OUTPUT_FOLDER_NAME,
         )
 
+        self.use_filename = Binary(
+            "Prefix saved crop image name with input image name?",
+            value=False,
+            doc="""\
+If *Yes*, the filename of the saved cropped object will be prefixed with
+the filename of the input image.
+
+For example:
+
+**Input file name**: positive_treatment.tiff
+
+
+**Output crop file name**: positive_treatment_Nuclei_1.tiff
+
+
+where "Nuclei" is the object name and "1" is the object number. 
+            """,
+        )
+
         self.file_image_name = FileImageSubscriber(
-            "Select image name for file prefix",
+            "Select image name to use as a prefix",
             "None",
             doc="""\
 Select an image loaded using **NamesAndTypes**. The original filename
@@ -110,6 +129,7 @@ after the selected image name prefix.
         settings = [
             self.objects_name,
             self.directory,
+            self.use_filename,
             self.file_image_name,
             self.file_format,
             self.export_option,
@@ -128,7 +148,11 @@ after the selected image name prefix.
         result += [
             self.objects_name,
             self.directory,
-            self.file_image_name,
+            self.use_filename
+        ]
+        if self.use_filename.value:
+            result += [self.file_image_name]
+        result += [
             self.nested_save,
             self.file_format,
         ]
--- a/cellprofiler/modules/savecroppedobjects.py
+++ b/cellprofiler/modules/savecroppedobjects.py
@@ -66,7 +66,7 @@ The choices are:
 
         self.objects_name = LabelSubscriber(
             "Objects", 
-            doc="Select the objects you want to export as per-object crops.",
+            doc="Select the objects to export as per-object crops.",
         )
 
         self.image_name = ImageSubscriber(
@@ -94,7 +94,7 @@ The choices are:
             value=False,
             doc="""\
             If *Yes*, the output crops will be saved into a folder named
-            after the input image. 
+            after the image from which the crops were derived. 
             """,
         )
         self.include_filename = Binary(
--- a/cellprofiler/modules/savecroppedobjects.py
+++ b/cellprofiler/modules/savecroppedobjects.py
@@ -122,6 +122,9 @@ will be used as the prefix for the output filename."""
             doc="""\
 If *Yes*, the output crops will be saved into a folder named
 after the selected image name prefix. 
+
+If no image name prefix is selected, crops will be saved into 
+a folder named after the input objects.
             """,
         )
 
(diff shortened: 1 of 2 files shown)


cellprofiler/modules/savecroppedobjects.py
```diff
@@ -97,6 +147,14 @@ def run(self, workspace):
         if not os.path.exists(directory):
             os.makedirs(directory)
 
+        input_filename = workspace.measurements.get_current_measurement("Image", self.source_file_name_feature)
+        input_filename = os.path.splitext(input_filename)[0]
+
+        if self.nested_save:
+            nested_folder = os.path.join(directory, input_filename)
+            if not os.path.exists(nested_folder):
+                os.makedirs(nested_folder)
```
bethac07, line 156: You may want to add an exist_ok here.

callum-jpg:
So for this I've hit a bit of a roadblock, namely trying to get the filename tests to work. I've included what I've tried so far in [here](https://github.com/CellProfiler/CellProfiler/pull/4545/commits/9f9d348206cdf5462e1cdd017be0957b9f678789). Basically, the error arises in response to the `source_file_name_feature` method used in SaveCroppedObjects, which is: `AssertionError: Feature FileName_example for Image does not exist`. I'm posting this here so that there may be light shed on what I'm missing. 

The same method I'm using is also used in SaveImages, but test_saveimages does not cover this method.

bethac07:
So, a couple of things here -

First of all, I missed that in the main module itself, we definitely need to bump the module version and add an upgrade_settings path - apologies for missing that earlier!

Second of all, as currently written, we're not "[Adding] the _option_ to add input filenames to output crops" (emphasis mine), we're actually enforcing it. 95% of the time, I think this is what users want, but it's ultimately a breaking change. As such, we either need to a) make it an option and/or b) we should hold it for CP5. In general, I'm nearly always in favor of making things optional, but I understand the need to sometimes stop the option-creep. If we do make it optional, I'm fine with having the default be True for newly added modules, but we want it to be "False" for upgraded modules in your upgrade_settings that you'll be adding. 

Finally, RE: tests, I see what you were trying to do here @callum-jpg, but you were going at it in a harder way than you needed to - rather than forcing CellProfiler to load the data the way that you wanted, just making sure CellProfiler has the information it's going to need to do the thing you want to check is WAY easier, and ultimately, what we care about here - we don't actually want to test CellProfiler image loading here, we just want to test that it will make file names the way we want them to. Make sense? I removed them, because I think there are going to be easier ways to do what you want to do, and because there are still module decisions to make, but you can always get them back in the commit history.
The existing tests are failing, due to the fact that we've changed the way file names are constructed. I fixed just the first one, and it passes now. So depending on the decision on optional-ness, a couple of different things to possibly do here -
1) Decide this change is going to be mandatory, in which case, fix the rest of the failing tests according to the model of how I fixed the first one, and then just add yourself one more test for nested foldernames (which should be more clear to do after fixing the others)
2) Decide this change is going to be optional, add the code to the module to make it so, and then in the tests a) copy my "fix" of the first test as a new test that now checks this "optional" behavior, b) restore the first test to the way it previously was, and make sure it and the others have the appropriate options set (ie "use image file name" or whatever you call it = False) so that they follow the previous behavior and c) add any other tests you think are necessary - I think probably still just something for nested folders but up to you.

I threw some time on your calendar to walk through all of this later today, in case anything was unclear here!

bethac07:
Per an offline conversation, we're going to change this behavior to be optional, and try to get it into 4.2.2. Should be simple from here!

bethac07:
Me: "this module is already on revision 2, why didn't it have an upgrade_settings already?"
Also me: 
image

callum-jpg:
When running pytest on `test_savecroppedobjects.py`, `test_run_images` fails for the `grayscale_volume` and `multichannel_volume` images. However, running `pytest tests/modules/test_savecroppedobjects.py::test_run_images` does not lead to the same test failing. I'm not sure what is causing this. Can anyone else confirm if this happens on their side?

bethac07:
I could validate the behavior - if you look at the order the tests were running, it will give you an initial clue why this was happening. 

If I run just that one test -
image
If I run all the tests - 
image
It gets even more clear when you add some print statements - here to the failing test and the nested test I've added prints for the name of the directory and whether it exists; also to the failing test, I've had it print the results of its glob, then an os.listdir of the directory
image

So based on all that, here is what was happening -
When its running all the tests using the 3 "standard" fixtures (the last 4 tests do NOT use the standard fixtures), rather than test-by-test going through all 3 fixtures, it fixture-by-fixture goes through all the tests that use those fixtures and then goes onto the next fixture.
Also, there are two ways tests can be set up - re-initializing the workspace and the module every time (this is typical in the older tests, where you see those "make_workspace" functions), OR you can re-initialize the module and the workspace every time in each test (you can see an example of this in the ExportToSpreadsheet tests) OR they can be using the conftest-provided-workspace and typically, sharing a copy of the module that's set at the top of the test file with "instance". The savecroppedobjects tests are using the last one. 

The first fixture, since nested_save default is false, and because the test that expects nested_save to be off runs first, both tests pass fine. BUT, on the second and third fixtures, since the tests are sharing a copy of the module, you've now set nested_save is true when you ran the test for that, but you haven't explicitly re-set it BACK to false. So it's still set to true, and so the first test (whose `asserts` think nested_save will be off) fails. Adding an explicit set of that property back to false in whichever is the first module using those fixtures AFTER the nested_save test fixes the test. 

I've done that, but also re-added-back the initial image test. I think this is probably now ready for review!

cellprofiler/modules/savecroppedobjects.py
```diff
@@ -126,67 +210,64 @@ def run(self, workspace):
...
-            self.directory,
-            self.file_format,
-        ]
-
-        if self.export_option.value == SAVE_PER_OBJECT:
-            result += [self.image_name]
-
-        return result
+    def upgrade_settings(self, setting_values, variable_revision_number, module_name):
+        if variable_revision_number < 3:
```
bethac07, line 258: This will almost certainly barf if it gets a v1 module, since it won't have enough values below. Can you add what I should have done in #3671, which is add a 1->2 upgrader that appends [SAVE_MASK,None] to the settings? Thank you!

cellprofiler/modules/savecroppedobjects.py
```diff
@@ -126,67 +210,64 @@ def run(self, workspace):
...
+            # Older module version, revert to not using file names in output crops
+            # Also, reorder setting_values
```
bethac07, line 260: Absolutely fine to reorder, but hard to follow the changes, can you add one more line of comment saying what you're reordering from and what you're reordering to? I can see it in the diff right here right now, of course, but for future folks reading the code but not looking at the specific PR that introduced the change

cellprofiler/modules/savecroppedobjects.py
```diff
@@ -255,11 +255,21 @@ def run(self, workspace):
...
+            # [objects_name, directory, file_format, export_option, image_name]
+            # New order:
+            # [export_option, objects_name, directory, use_filename, file_image_name, nested_save, file_format, image_name]
             setting_values = (
-                [setting_values[3]] + setting_values[:2] + [False, "No", None] + [setting_values[2]] + [setting_values[4]]
+                [setting_values[3]] + setting_values[:2] + [False, "None", False] + [setting_values[2]] + [setting_values[4]]
+            )
+            variable_revision_number = 3
+
+        if variable_revision_number == 1:
```
bethac07, line 269: This needs to be first, so that if someone has a 1, it bumps to a 2, and then at the end of that if statement it will be a 2, so now can bump to a 3, etc

bethac07: Ah, nvm, I see what you did - for parallelness, if you don't mind following that structure (1 bumps to 2 bumps to 3 etc as opposed to a "2 OR 1 bumps to 3, then 3 bumps to 4, etc"), I think future-us-es will be less confused.

cellprofiler/modules/savecroppedobjects.py
bethac07: Thanks!

cellprofiler/modules/savecroppedobjects.py
```diff
@@ -126,67 +210,78 @@ def run(self, workspace):
...
-
-        return result
+    def upgrade_settings(self, setting_values, variable_revision_number, module_name):
+        if variable_revision_number == 1:
+            # Old order: 
+            # [objects_name, directory, file_format]
+            # New order:
+            # [objects_name, directory, file_format, export_option, image_name]
+            setting_values = (
+                [setting_values[:3] + [SAVE_PER_OBJECT, "Image"]]
```
bethac07, line 264: Can you check the data structure here and below? I think you're starting from a single un-nested list and ending in a nested list.

bethac07: Actually, below looks fine, I think it's just here

callum-jpg: Ah, I didn't spot that. Thank you

gnodar01 (approved):

gnodar01:
Added the new file format to the help docs, but other than that looks good to me.

PR #4552, opened 2022-05-13
Remove java dependencies from input modules

DavidStirling:
Companion PR for CellProfiler/Core#97 - full details there

Key changes impacting the CellProfiler/CellProfiler repo:

- PathListCtrl (used by Images module) redesigned to support display of image plane details alongside entries. The main file list can now display metadata about an image file (number of series and their dimensions). UI has been put together with a proper TreeCtrl to give more flexibility in doing this.
- Java VM is no longer started upon launching CellProfiler, so any modules still using JavaBridge (not many) should call the new `start_java` utility function before doing so.


deflaux:
Is there an estimate for when this PR might be merged?

gnodar01:
@deflaux We're releasing 4.2.3 soon (release candidate today, and soon after 4.2.3 proper). After that this PR will kick off version 5. There's still a few features to add before we publish 5 to PyPI, but it will be rebased on master.

PR #4564, opened 2022-07-11
Pin numpy below 1.23

bethac07:


gnodar01:
#4576 resolves this without putting upper bounds on numpy and mohotas.

PR #4568, opened 2022-07-15
Use correct scaler, populate choices

### LINKED ISSUE #4382: Can't Read Image in FlagImage Module
**Describe the bug**
I'm measuring colocalization between a list of images (of similar names, e.g. DAPI_round1, DAPI_round2, DAPI_round3) and want to create a flag if their correlation is low. However, in the FlagImage module, the drop down bar for Image isn't wide enough to show my full image names so I can't actually tell which comparison I'm selecting.

Can we make it expand to actually fit the text?

Screen Shot 2021-04-21 at 1 23 20 PM

 -  OSX 15.7
 - CellProfiler Version 4.1.3
 - Source

bethac07:
Resolves #4382 

It looks like there were two issues here contributing to these boxes not scaling correctly. Firstly, the scaler was grabbing the names of the choices (literally "Categories" instead of the actual categories, etc). Secondly, it didn't know what the choices were, because they hadn't yet been populated. By grabbing them earlier, we can populate all the possible options (hence combining images and objects) when we do the sizing, allowing the boxes to be properly scaled. 

We should check that this doesn't break anything on Windows, of course.

### DIFF
--- a/cellprofiler/gui/module_view/_module_view.py
+++ b/cellprofiler/gui/module_view/_module_view.py
@@ -1822,6 +1822,17 @@ class ModuleView:
         # * panel = None - create the controls
         # * panel != None - find the controls
         #
+        category = v.get_category(self.__pipeline)
+        categories = v.get_category_choices(self.__pipeline)
+        feature_name = v.get_feature_name(self.__pipeline)
+        feature_names = v.get_feature_name_choices(self.__pipeline)
+        image_name = v.get_image_name(self.__pipeline)
+        image_names = v.get_image_name_choices(self.__pipeline)
+        object_name = v.get_object_name(self.__pipeline)
+        object_names = v.get_object_name_choices(self.__pipeline)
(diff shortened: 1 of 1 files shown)


cellprofiler/gui/module_view/_module_view.py
```diff
@@ -1822,6 +1822,17 @@ def make_measurement_control(self, v, panel):
         # * panel = None - create the controls
         # * panel != None - find the controls
         #
+        category = v.get_category(self.__pipeline)
```
gnodar01, line 1825: Out of the scope for the related issue, but `v` is an item from the list returned by `__module.visible_settings()`, so I assume `v` means `visible_setting`? If yes, can we change it to that explicitly?

bethac07: Fine with me, though I can neither confirm nor deny that that's what v is

gnodar01 (approved):

PR #4575, opened 2022-07-17
Allow fuzzy column matching

### LINKED ISSUE #4370: FilterObjects fails when column names have been truncated
Originally reported on [the forum](https://forum.image.sc/t/measuretexture-parameters-not-recognised-as-rules/48610/5?u=bcimini) in CellProfiler 4.0.7
When exporting to database, CellProfiler's default is to truncate the column names if they are greater than 64 characters; it's not a just "truncate from the end" but drops vowels first, etc.  

If you then want to use your created rules in FilterObjects, the rules which relate to truncated column names can't be matched to the "real" names; I feel like this used to work, but can't say 100% for certain. 

**To Reproduce**
@pearlryder has a set we can use to reproduce this.

Originally reported on in CellProfiler 4.0.7.

bethac07:
Resolves #4370 .

ExportToDatabase has a functionality to truncate column names so they fit in MySQL (where columns at least in the past, possibly still now, can only be 64 characters long), but it means that if you train something in CPA based on some measurements then want to use your rules/classifier, you can't, because the column names no longer match. This PR adds the ability to optionally allow fuzzy matching to get around this, while not enforcing it if that isn't your situation. 

We will presumably want to also do this for the two other modules that use rules, ClassifyObjects and FlagImage, but this does technically resolve the issue as stated.


bethac07:
I think this is now done for all 3, but a) I don't think the FlagImage classifier mode likely works at all (the tests don't look like our other CPA-classifier-based tests, and the module hasn't received any pushes since CPA3 came out), so I can't get a test to fail properly there but I think it's just that it doesn't work at all and needs to be added and b) apparently ClassifyObjects has no tests at all for Rules or Classifier mode :( :( :( :( so I haven't actually checked that those work yet, because it's writing a whole bunch of tests, not just 1 or 2

bethac07:
Ok, I've now confirmed that it's working as expected to the degree I can, and I've made new issues to remind us to fix the stuff in FlagImage and add more tests to ClassifyObjects but I think that's out of scope here. And please, for all that is kind, squash my commits when pulling, because for some reason the nested class structure entirely just made me eat it on this for WAYYY too long.

One thing to note for future us's though, especially as we move towards the library; one place I got stuck for a while was that once I'd refactored the settings into the Rules module, in order for them to actually pass properly in the GUI I had to make sure I was passing the self.setting into the Rules, NOT the self.setting.value. Passing stuff across the classes still trips me up sometimes.

bethac07:
Now also resolves #4581

gnodar01 (approved):

PR #4576, opened 2022-07-18
Numpy123support

gnodar01:
non-tuple multidimensional index was deprecated in numpy v1.15, and finally dropped [here](https://github.com/numpy/numpy/pull/21029). This patch addresses that, and all tests pass using numpy v1.23.1 now.

We are however still receiving a 9 or 10 warnings on numpy features that were deprecated in numpy v1.20. We should probably resolve them sooner rather than later, and we may as well on this PR.

### DIFF
--- a/cellprofiler/modules/filterobjects.py
+++ b/cellprofiler/modules/filterobjects.py
@@ -976,7 +976,9 @@ measurement is not available at this stage of the pipeline. Consider adding modu
             #
             # Get the label of the pixel at each location
             #
-            indexes = src_labels[best_pos.transpose().tolist()]
+            # Multidimensional indexing with non-tuple values is not allowed as of numpy 2.23
+            best_pos = tuple(map(tuple, best_pos.transpose()))
+            indexes = src_labels[best_pos]
             indexes = set(indexes)
             indexes = list(indexes)
             indexes.sort()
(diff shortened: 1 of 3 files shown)


cellprofiler/modules/filterobjects.py
```diff
@@ -976,7 +976,9 @@ def keep_per_object(self, workspace, src_objects):
             #
             # Get the label of the pixel at each location
             #
-            indexes = src_labels[best_pos.transpose().tolist()]
+            # Multidimensional indexing with non-tuple values is not allowed as of numpy 2.23
```
bethac07, line 979: Prob want to fix this to 1.23 to not confuse future-us :)

cellprofiler/modules/relateobjects.py
```diff
@@ -614,7 +614,9 @@ def calculate_minimum_distances(self, workspace, parent_name):
             perim_loc = numpy.argwhere(pperim != 0)
 
             # Get the label # for each point
-            perim_idx = pperim[perim_loc.transpose().tolist()]
+            # multidimensional indexing with non-tuple values not allowed as of numpy 2.23
```
bethac07, line 617: Prob want to fix this to 1.23 to not confuse future-us :)

bethac07:
🎉 ! Thanks, you're a rock star :)

PR #4584, opened 2022-07-20
Add test workflow

gnodar01:
Setup a github action to run tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,14 +45,14 @@ jobs:
       - name: Linux - install mysql and build tools
         env:
           PATH: $PATH:/home/ubuntu/.local/bin
-        if: startsWith(runner.os, 'Linux')
+        if: startsWith(matrix.os, 'ubuntu')
         run: |
           sudo apt install -y make gcc build-essential libgtk-3-dev
           sudo apt-get install -y default-libmysqlclient-dev libnotify-dev libsdl2-dev
       - name: Mac - setup mysql
         env:
           LC_ALL: "en_US.UTF-8"
-        if: ${{ matrix.os == 'macos-11' }}
+        if: startsWith(matrix.os, 'macos')
         run: |
           brew install mysql
           mysql.server start
@@ -67,10 +67,10 @@ jobs:
           pip install -e .[test]
       - name: Test
         run: |
-          mysql.server status
-          mysql --version
-          mysql -h $CP_MYSQL_TEST_HOST -u $CP_MYSQL_TEST_USER -e "show databases;"
-          mysql -h localhost -u root -e "use mysql; select user from user;"
+          # mysql.server status
+          # mysql --version
+          # mysql -h $CP_MYSQL_TEST_HOST -u $CP_MYSQL_TEST_USER -e "show databases;"
+          # mysql -h localhost -u root -e "use mysql; select user from user;"
           pytest
 name: run-cp-tests
 on: [workflow_dispatch]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,3 +1,4 @@
+name: run-cp-tests
 jobs:
   test:
     name: Install and test Cellprofiler
@@ -53,5 +54,10 @@ jobs:
       - name: Run tests
         run: |
           pytest
-name: run-cp-tests
-on: [workflow_dispatch]
+on:
+  pull_request:
+    types: [opened, reopened, ready_for_review, review_requested]
+  push:
+    branches:
+      - 'master'
+  workflow_dispatch:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,8 @@ jobs:
           java-version: '14.0.1' # The JDK version to make available on the path.
           java-package: jdk
           architecture: x64
-      - run: |
+      - name: Initial installs
+        run: |
           pip install pyinstaller
           pip install --upgrade pip setuptools wheel
           pip install numpy>=1.20.1
@@ -36,13 +37,15 @@ jobs:
           git clone https://github.com/CellProfiler/centrosome.git ~/centrosome
           git clone https://github.com/CellProfiler/core.git ~/core
           # git clone https://github.com/CellProfiler/CellProfiler.git ~/cellprofiler
-      - env:
+      - name: Linux - install mysql and build tools
+        env:
           PATH: $PATH:/home/ubuntu/.local/bin
         if: startsWith(runner.os, 'Linux')
         run: |
           sudo apt install -y make gcc build-essential libgtk-3-dev
           sudo apt-get install -y default-libmysqlclient-dev libnotify-dev libsdl2-dev
-      - env:
+      - name: Mac - setup mysql
+        env:
           LC_ALL: "en_US.UTF-8"
           CP_MYSQL_TEST_HOST: "127.0.0.1"
           CP_MYSQL_TEST_USER: "root"
@@ -51,8 +54,9 @@ jobs:
         run: |
           brew install mysql
           mysql.server start
-          mysql -h 127.0.0.1 -u root -p "" -e "CREATE DATABASE cellprofiler_test;"
-      - run: |
+          mysql -h 127.0.0.1 -u root -e "CREATE DATABASE cellprofiler_test;" --password=""
+      - name: Install CP
+        run: |
           pip install -e ~/python-javabridge
           pip install -e ~/python-bioformats
           pip install -e ~/centrosome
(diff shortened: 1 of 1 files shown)


gnodar01:
Should be squashed on merge.

bethac07:
>Right now I have tests running, with caching, just on python 3.8, and macos-11 (i.e.[macos-latest](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)). macOS 10.15, which we currently have release.yml set to use, has been https://github.com/actions/virtual-environments/issues/5583 and will be completely phased out 8/30/22.

Makes sense to do that for tests, but builds we often want to keep the oldest one we can - these are our Google Analytics of visitors to the CellProfiler site for the last 6 weeks, filtered to just the Mac users, but this seems highly suspicious and likely to be wrong to me. But since if we build on 10.15, 11.X users CAN use, but the reverse isn't true, I'd err on the side of building 4.2.2 on 10.15. What do you think? 
image

bethac07:
>I could add windows-latest (windows-2022) and ubuntu-latest (ubuntu-20.04) to the os matrix, but 1) I'll need to get mysql installed and running on each one (if there are already instructions somewhere on this, please point me to them), and 2) not sure if it's even worth it, given the amount of time it takes on just macos.

I would think that the Windows build instructions for the app, and the ubuntu installation instructions on the GH wiki, would probably take care of most of the pain of this. I think matrixed stuff also typically all builds simultaneously, rather than sequentially (at least up to 4-5 runners), so it wouldn't necessarily slow the testing unless one was dramatically slower than Mac.

With all that being said, I would say the priority for doing this would be low-medium for Windows (there ARE some Windows-only tests) and low-low for Linux. If you forced me to pick between matrixing these and Python versions, I'd definitely pick Python versions.

.github/workflows/test.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Install CP
+        run: |
+          pip install -e ~/python-javabridge
+          pip install -e ~/python-bioformats
+          pip install -e ~/centrosome
+          pip install -e ~/core
+          pip install -e .[test]
+      - name: Test
+        run: |
+          # mysql.server status
```
bethac07, line 70: Any reason to keep these?

.github/workflows/test.yml
```diff
@@ -0,0 +1,76 @@
...
+          sudo apt-get install -y default-libmysqlclient-dev libnotify-dev libsdl2-dev
+      - name: Mac - setup mysql
+        env:
+          LC_ALL: "en_US.UTF-8"
+        if: startsWith(matrix.os, 'macos')
+        run: |
+          brew install mysql
+          mysql.server start
+          mysql --host=$CP_MYSQL_TEST_HOST --user=$CP_MYSQL_TEST_USER --execute="CREATE DATABASE $CP_MYSQL_TEST_DB;" --skip-password
+          # mysql --host=$CP_MYSQL_TEST_HOST --user=$CP_MYSQL_TEST_USER --execute="CREATE DATABASE $CP_MYSQL_TEST_DB;" --password=$CP_MYSQL_TEST_PASSWORD
```
bethac07, line 60: Any reason to keep this?

.github/workflows/test.yml
```diff
@@ -0,0 +1,76 @@
...
+        run: |
+          pip install pyinstaller
+          pip install --upgrade pip setuptools wheel
+          pip install numpy>=1.20.1
+          git clone https://github.com/CellProfiler/python-javabridge.git ~/python-javabridge
+          git clone https://github.com/CellProfiler/python-bioformats.git ~/python-bioformats
+          git clone https://github.com/CellProfiler/centrosome.git ~/centrosome
+          git clone https://github.com/CellProfiler/core.git ~/core
+          # git clone https://github.com/CellProfiler/CellProfiler.git ~/cellprofiler
+      - name: Linux - install mysql and build tools
```
bethac07, line 45: If we decide against adding linux for now, do we wan to clean this up?

.github/workflows/test.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Initial installs
+        run: |
+          pip install pyinstaller
+          pip install --upgrade pip setuptools wheel
+          pip install numpy>=1.20.1
+          git clone https://github.com/CellProfiler/python-javabridge.git ~/python-javabridge
+          git clone https://github.com/CellProfiler/python-bioformats.git ~/python-bioformats
+          git clone https://github.com/CellProfiler/centrosome.git ~/centrosome
+          git clone https://github.com/CellProfiler/core.git ~/core
+          # git clone https://github.com/CellProfiler/CellProfiler.git ~/cellprofiler
```
bethac07, line 44: Should be able to remove this now, yes? woohoo!

bethac07 (approved):
A couple minor things you may want to clean up (or not, if they're notes to future you!), but otherwise, yay, thanks for getting this working!

.github/workflows/test.yml
bethac07: Never mind, I lost track of what repo I was in, I am mixed up lol

.github/workflows/test.yml
```diff
@@ -0,0 +1,76 @@
...
+          LC_ALL: "en_US.UTF-8"
+        if: startsWith(matrix.os, 'macos')
+        run: |
+          brew install mysql
+          mysql.server start
+          mysql --host=$CP_MYSQL_TEST_HOST --user=$CP_MYSQL_TEST_USER --execute="CREATE DATABASE $CP_MYSQL_TEST_DB;" --skip-password
+          # mysql --host=$CP_MYSQL_TEST_HOST --user=$CP_MYSQL_TEST_USER --execute="CREATE DATABASE $CP_MYSQL_TEST_DB;" --password=$CP_MYSQL_TEST_PASSWORD
+      - name: Install CP
+        run: |
+          pip install -e ~/python-javabridge
```
bethac07, line 63: I don't think it's a _bad_ idea to build all of these from source, but I think if we wanted to shave some time then using PyPI wheels for most of these should typically work.

gnodar01:
> I'd err on the side of building 4.2.2 on 10.15

I'm in general agreement, but on the last run I tried with `macos-10.15`, it was [automatically terminated](https://github.com/CellProfiler/CellProfiler/actions/runs/2712961459) because of the deprecation, stating:

> This is a scheduled macOS-10.15 brownout.

So we have to make sure we run the 4.2.2 release workflow outside of the brownout periods (posted in that [same deprecation issue](https://github.com/actions/virtual-environments/issues/5583)).

.github/workflows/test.yml
gnodar01: Nope, will remove

.github/workflows/test.yml
gnodar01: No, at least for now. I had some trouble around [this line](https://github.com/CellProfiler/CellProfiler/blob/master/tests/modules/test_exporttodatabase.py#L114), but I decided it was outside of the scope of this PR, so I'll remove it for now.

.github/workflows/test.yml
gnodar01: Will do.

.github/workflows/test.yml
gnodar01: Okay. I'll keep core from source, since we always want to have the latest, and will remove the others.

bethac07:
Shall we?

gnodar01:
Note for future self, and others:

For some reason the "Cache pip" step, [this run](https://github.com/CellProfiler/CellProfiler/runs/7459555397?check_suite_focus=true) reports:

> Cache not found for input keys: macOS-pip-a6443ba1eee7b09cafd18f4b377dac37bec6ff0e3c4f60e9be4b42dd723a816d

Even though only a few minutes before [the previous successful run](https://github.com/CellProfiler/CellProfiler/runs/7458784769?check_suite_focus=true) reported:

> Cache restored from key: macOS-pip-a6443ba1eee7b09cafd18f4b377dac37bec6ff0e3c4f60e9be4b42dd723a816d

and in the post cache step:

> Cache hit occurred on the primary key macOS-pip-a6443ba1eee7b09cafd18f4b377dac37bec6ff0e3c4f60e9be4b42dd723a816d, not saving cache.

The GH CLI is also reporting a cache exists with that key, so it seemed a bit weird that there was a cache miss. I'm reasonably certain that it's because I was not including the python version in the key generation, so the python 3.8 cache save was overwriting the python 3.9 cache contents, and vice versa.

gnodar01:
> If you forced me to pick between matrixing these and Python versions, I'd definitely pick Python versions.

One minor thing to note about this, as long as we're using `wxPython==4.1.0`, cache misses are going to lead to slow installs on python 3.9, mac and windows included, since only `4.1.1` has a python 3.9 wheel on PyPI. 

Not really a big deal, but one option is to host our own macos 11 + python 3.9 + wxPython 4.1.0 wheel...

PR #4605, opened 2022-08-01
Add widget_inspector option

gnodar01:
- add option to set widget_inspector, enabling it as item under "Test" menu
- needs corresponding core changes, in CellProfiler/core#111

### DIFF
--- a/cellprofiler/__main__.py
+++ b/cellprofiler/__main__.py
@@ -44,6 +44,7 @@ from cellprofiler_core.preferences import set_omero_server
 from cellprofiler_core.preferences import set_omero_user
 from cellprofiler_core.preferences import set_plugin_directory
 from cellprofiler_core.preferences import set_temporary_directory
+from cellprofiler_core.preferences import set_widget_inspector
 from cellprofiler_core.utilities.core.workspace import is_workspace_file
 from cellprofiler_core.utilities.hdf5_dict import HDF5FileList
 from cellprofiler_core.utilities.java import start_java, stop_java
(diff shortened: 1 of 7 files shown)


bethac07:
Fun! I like it :)

It should definitely be documented - maybe in the Help -> Other Features area where we also document things like the debug shell?

Do we want to it be additionally visible in the preferences menu, aka so you could set the preference on the fly without needing to reboot CellProfiler to call it with the flag?

gnodar01:
@bethac07 I've added it to preferences, and it's now dynamically injected as a menu item when it's enabled in preferences.

[Line 507](https://github.com/CellProfiler/CellProfiler/pull/4605/files#diff-e55f83479727f5c54025ec3b040129fee91362b217e79676aca40661bed62262R507) in `_preferences_dialog` was the "hacky" bit I was talking about. To slightly less dirty things I can think of: 1) I could add a getter for `__menu_debug` in `CPFrame`, 2) I could move the entirety of `_set_widget_inspector` into `CPFrame`. Thoughts?

bethac07:
I think we could think about refactoring out into something nice if there ever becomes a second preference that we want to use to dynamically update a menu item, but in the absence of having an idea for another use case, the existing code seems totally within-bounds to me.

gnodar01:
I refactored it so that `cpframe` has methods for injecting and removing menu items. There's a lot that could be done to make it more robust and generalized, but given we don't have a second use case in mind right now something, something premature abstraction = bad.

bethac07:
Sounds reasonable to me! It would be awesome to add documentation before pulling, just so that we know we won't forget to do it later - here is a recent [documentation PR](https://github.com/CellProfiler/CellProfiler/pull/4570/files) which demos adding a piece of documentation, though in this case I think we want it in the "Other Features" help section.

bethac07 (approved):

PR #4622, opened 2022-08-15
Make pip upgrade and setuptools and wheel install on Windows

bethac07:


.github/workflows/release.yml
```diff
@@ -50,6 +50,7 @@ jobs:
           JDK_HOME: C:\hostedtoolcache\windows\Java_Temurin-Hotspot_jdk\11.0.16-8\x64
         if: ${{ matrix.os == 'windows-2019' }}
         run: |
+          c:\hostedtoolcache\windows\python\3.8.10\x64\python.exe -m pip install --upgrade pip setuptools wheel
```
gnodar01, line 53: Why not just use `python` directly? This might cause trouble if we every use a different python version in the matrix, and forget to change it.

bethac07: agreed, but that's what the actual logs recommend to do - we can try that as a first pass though.

gnodar01 (approved):

PR #4634, opened 2022-08-30
Adding threshold to cellprofiler library

callum-jpg:
The beginnings of adding threshold to CellProfiler library has raised some questions. Currently, the only functions added to cellprofiler.library.functions.image_processing are `get_threshold_robust_background` and `get_threshold`. `get_threshold` is used within the Threshold module as a general function to read a users input settings and return a desired threshold (eg global or adaptive). `get_threshold` within the Threshold module passes instance attributes between methods (eg. adaptive requires `get_global_threshold`, `get_local_threshold`, then `apply_threshold`). To achieve this in cellprofiler library (and to avoid passing lots of settings between functions), I created `library_utils`, which contains a class that basically carries out all this logic for settings passed to `get_threshold`. However, this approach raises some issues:

- `get_threshold` represents a complex function that could be further split up into `get_global_threshold` and `get_adaptive_threshold`, for example, allowing for more granular control
- Do functions like `_correct_local_threshold` deserve their own library function, or should they be baked into `get_adaptive_threshold`?
- Ultimately, it should be cellprofiler.library.modules._threshold that contains the full equivalent of`get_threshold`. But how will we control all this logic across multiple distinct functions? Classes, or just repetitively passing the arguments to all functions that are required?
- `get_threshold` also currently takes in either *manual* or *measurement* threshold options that are not actually relevant for someone using this function in a cellprofiler library context. Instead, these scenarios should just be accounted for in the Threshold module itself, rather than in any library function, which is currently the case.
- `get_threshold` using threshold_operation names like "robust background", which should really be changed to "robust_background"
- The return of `get_threshold` is inconsistent between global and adaptive thresholds (eg. a global threshold is a single value, whereas for adaptive it is an array of image.shape containing lots of 'windows' of threshold values). Additionally, adaptive returns `th_guide` whereas global returns `None` (see below example).

```
# Minimal code to try out get_threshold
from cellprofiler.library.functions.image_processing import get_threshold
import skimage
import numpy as np

image = skimage.data.human_mitosis()
image = skimage.img_as_float(image)

# th_corrected is the threshold that has been constrained by the
# threshold_correction_factor. For adaptive, this is limited 
# to between [0.7, 1.5] of the global threshold
th_corrected, th_original, th_guide = get_threshold(
    image=image, 
    mask=mask,
    threshold_correction_factor=0.9, 
    threshold_operation="robust background",
    threshold_scope="adaptive", 
    adaptive_window_size=25
    )

### Plotting will only work for adaptive thresholds
### This is because adaptive returns an array of "windows"
### of local thresholds
fig, ax = plt.subplots(1, 3)

ax[0].set_title("input image")
ax[0].imshow(image, cmap="gray")

ax[1].set_title("th_corrected")
ax[1].imshow(th_corrected, cmap="gray")

ax[2].set_title("th_original")
ax[2].imshow(th_original, cmap="gray")


```


cellprofiler/library/functions/library_utils.py
```diff
@@ -0,0 +1,296 @@
+from re import S
+import numpy
+import skimage
+import scipy
+import centrosome
+import cellprofiler.library as cp
```
bethac07, line 6: We've generally decided against aliasing in most places in the code, to increase readability, though that's something we can certainly readdress

bethac07:
Thanks for summarizing all of this! 

My thoughts in no particular order
- I think, especially since we hope to do more typing of things, that having the two sub-functions that return really different output structures (a float vs an array) should be in two different functions, so I'd support a `get_global_threshold` vs `get_local_threshold` distinction. 
-- All of that is assuming, of course, that we want to ever even return the threshold at all, as opposed to the thresholded image (which is what the current Threshold module produces). I think we probably do, along with an `apply_threshold` function that then does the right thing depending on the input type. 
-- We could consider whether we then want something like a `threshold_image` function that runs `get_global/local_threshold` (depending on args) followed by `apply_threshold`. 
- >Do functions like _correct_local_threshold deserve their own library function, or should they be baked into get_adaptive_threshold

Not sure, honestly, I think we probably need to take these case-by-case. In this case I think I'd say "no", though while we're exposing functionality, we could consider if there's ever a case where we definitely DON'T want to do this and if so, do we make running the correction an option.
-

callum-jpg:
I suppose a `threshold_image` function would be very similar to the from cellprofiler.library.modules function for `threshold`, which would take in all of the args and return the thresholded image

callum-jpg:
The potentially tricky part about not baking `_correct_local_threshold` ([link](https://github.com/CellProfiler/CellProfiler/blob/c6a6236dd487d94ffea493aef1d9e6e1a625b998/cellprofiler/modules/threshold.py#L1089)) into the `get_adaptive_threshold` is that the latter function involves `RectBivariateSpline`, which is used to 'smooth' out the blockiness of the adaptive threshold array. This results in negative numbers in the returned threshold. Within *threshold.py*, these negatives are handled by the aforementioned `_correct_local_threshold`, in which the negative numbers are trimmed between either the provided threshold min/max or 0.7-1.5x the global threshold. While this seems like an easy thing to add in at the end of `get_adaptive_threshold`, it's not the complete picture. Before min/max trimming, the threshold correction factor is applied to the original (ie containing negative numbers) threshold array.

So, for `get_adaptive_threshold` in cellprofiler.library to return a threshold that can be immediately applied to an image (and behave as in CellProfiler), the function will require the args threshold min/max, threshold guides (eg. 0.7-1.5x global), a threshold correction factor, in addition to the other args like image, threshold_method etc.

callum-jpg:
These latest commits add discrete functions for global and adaptive thresholding, with a third functions that applies a threshold. 

Here's a script for testing global. Also includes some external thresholding functions that you can use (same method works for `adaptive_threshold` too:
```
from cellprofiler.library.functions.image_processing import (
    get_adaptive_threshold, 
    get_global_threshold, 
    apply_threshold
    )
import skimage
import numpy as np

image = skimage.data.human_mitosis()
image = skimage.img_as_float(image)

th = "otsu"
# th = skimage.filters.threshold_yen
# th = skimage.filters.threshold_mean
# th = skimage.filters.threshold_isodata

threshold = get_global_threshold(
    image, 
    threshold_method=th,
    threshold_min=0,
    threshold_max=1,
    threshold_correction_factor=1,
    assign_middle_to_foreground="foreground",
    log_transform=False
)

# _ is the sigma calculated based on smoothing
thresh_img, _ = apply_threshold(
    image, 
    threshold, 
    smoothing=0
)

fig, ax = plt.subplots(1, 2, figsize=(10,10))
ax[0].imshow(image, cmap="gray")
ax[1].imshow(thresh_img, cmap="gray")
```

And here's one for testing 3D adaptive:
```
image = skimage.data.brain()
# Change to x, y, z
image = np.swapaxes(image, 0, 2)
image = skimage.img_as_float(image)

threshold = get_adaptive_threshold(
    image, 
    threshold_method="otsu", 
    window_size=20,
    threshold_correction_factor=1,
    log_transform=False, 
    volumetric=True
    )

thresh_img, _ = apply_threshold(image, threshold)

fig, ax = plt.subplots(5, 3, figsize=(10, 15))
for i, (ax_i, ax_j, ax_k) in enumerate(ax):
    ax_i.imshow(image[:,:,i], cmap="gray")
    ax_j.imshow(threshold[:,:,i], cmap="gray")
    ax_k.imshow(thresh_img[:,:,i], cmap="gray")
```

cellprofiler/library/functions/image_processing.py
```diff
@@ -197,7 +217,7 @@ def get_adaptive_threshold(
             j0 = int(j * increment[1])
             j1 = int((j + 1) * increment[1])
             block = image[i0:i1, j0:j1]
-            block = block[~numpy.isnan(block)]
+            block = block[~numpy.logical_not(block)]
```
callum-jpg, line 220: I've altered the adaptive thresholding of CellProfiler slightly.

The reason for changing `numpy.isnan` to `numpy.logical_not` is to make how CellProfiler deals with masks more similar to that in skimage, where masked pixels are False.

Additionally, this change is because `get_adaptive_threshold` makes a call to `get_global_threshold`. The former accepted numpy.nan pixels since they were filtered out in the line highlighted here.

However, since `get_adaptive_threshold` makes a call to `get_global_threshold`, the latter is fed NaN pixels. NaN pixels are not a valid input to skimage functions. There's a discussion about scikit-image and NaN values as input here: https://github.com/scikit-image/scikit-image/issues/1624

callum-jpg:
The tests now pass for the library implementation of threshold. One thing I noticed while doing this is that CellProfiler applies a log transformation **before** masking, which I feel should be the other way.

cellprofiler/modules/threshold.py
```diff
@@ -783,20 +785,20 @@ def run(self, workspace):
             self.x_name.value, must_be_grayscale=True
         )
         dimensions = input_image.dimensions
-        final_threshold, orig_threshold, guide_threshold = self.get_threshold(
-            input_image, workspace, automatic=False,
+
+        self.final_threshold, self.orig_threshold, self.guide_threshold, binary_image, _ = self.get_threshold(
```
bethac07, line 789: Why add these to the class? I'm not against it per se but it's not what we typically do.

cellprofiler/modules/threshold.py
```diff
@@ -815,339 +817,89 @@ def run(self, workspace):
...
-                not automatic and
-                self.threshold_operation in (TM_LI, TM_OTSU) and
-                self.log_transform
-        )
-
-        if need_transform:
-            image_data, conversion_dict = centrosome.threshold.log_transform(image.pixel_data)
-        else:
-            image_data = image.pixel_data
+    def convert_setting(self, gui_setting_str):
```
bethac07, line 826: Not necessarily something to change now, but if we end up needing to do this a lot, should we make this a higher-level helper function?

cellprofiler/modules/identifyprimaryobjects.py
```diff
@@ -15,6 +15,7 @@
 from cellprofiler_core.setting.choice import Choice
 from cellprofiler_core.setting.range import IntegerRange
 from cellprofiler_core.setting.text import Integer, Float
+from cellprofiler.library.modules import threshold as _threshold
```
bethac07, line 18: Is this used?

cellprofiler/modules/threshold.py
callum-jpg: I agree. Would you envision a function such as this in `library_utils.py`?

callum-jpg:
> Minor detail - in general, it would be nice to inherit from the library modules in IDPrimary/Secondary rather than the module-modules (aka, Threshold). I'm guessing for Primary it would be more painful to do that because we're using their measurement-recording functions, but what do we think about doing it for Secondary? Better to keep it consistent, or to do the best we can, when we can?

I originally added the threshold module to `identifyprimaryobjects` in this way to make the focus of this PR just on threshold, with minimal impact on other modules. When I get to adding `identifyprimaryobjects` to library I would like to try using only the library threshold module, rather than inheriting from the GUI module. I hope copying over the measurement recording stuff from threshold into IDprimary will be straight forward

cellprofiler/modules/threshold.py
callum-jpg: So the reason for this was tests. Originally, `test_threshold` got the final, local, guide threshold by calling the threshold class method called `get_threshold`. However, when adding threshold to library we wanted more control over which thresholding method you'd like to use, rather than calling this generalised `get_threshold` function. 

So, by adding final_threshold etc. as a class attribute it's possible to get access to these from the tests since now you do not have a `get_threshold` method. Though perhaps there's a better strategy for this.

bethac07 (approved):

PR #4665, opened 2022-11-10
Add Watershed to CellProfiler Library

callum-jpg:
The watershed analysis code has been split into 3 functions in `object_processing`: `watershed_distance`, `watershed_markers` and `watershed_advanced`. `watershed_advanced` will makes calls to both watershed_distance and watershed_markers before doing some additional processing. watershed_markers is basically just a wrapped around `skimage.segmentation.watershed`, but it keeps things consistent.

Originally, the watershed implementation would perform a skimage watershed on an RGB image and then convert the RGB image if multichannel to grayscale, [as seen here](https://github.com/CellProfiler/CellProfiler/blob/768c9b7ff2698c53d8079b34b7dac56910999d2d/cellprofiler/modules/watershed.py#L524). I've changed this so that conversion to grayscale happens outside of the image analysis code since I felt like there was no reasonable way to check if an image was 'multichannel' from within the library watershed function since multichannel is a cellprofiler class attribute. This change currently seems OK, but I'll test some more.


callum-jpg:
Currently, the watershed module has been split up into 3 functions for addition to cellprofiler library. However, should the watershed functionality be further split up?

The watershed module in the CellProfiler GUI offers some declumping methods when the advanced setting is selected. Currently, these options are available in the CellProfiler Library watershed_advanced function, but does it make sense to break these out into their own functions? For example, the resulting functions could be something like: `watershed_intensity_declump` and `watershed_shape_declump`.

However, the difficulty with something like this is that following declumping in watershed_advanced, there are a lot of prerequisite steps prior to applying `skimage.segmentation.watershed`, such as applying smoothing, generating seeds from local seeds etc. I suppose this code could simply be duplicated so that `watershed_advanced` is split into the aforementioned `watershed_intensity_declump` and `watershed_shape_declump`. @bethac07

bethac07:
Per our conversation today - I "approve" in the sense that I think this was done correctly, but I think this module needs an overhaul (or at least, an evaluation of whether or not it needs an overhaul), and I think we're way more likely to remember to do it as part of this process than later. So let's hold off on pulling until we do that.

gnodar01:
The refactor sounds good to me. I'm heavily in favor of doing it in this PR. In general, I'm in favor of having all "moving a module to library" PRs paired with "refactor the module if necessary" changes.

cellprofiler/modules/watershed.py
```diff
@@ -407,179 +411,59 @@ def run(self, workspace):
...
-
-            # Dilate seeds based on settings
-            seeds = skimage.morphology.binary_dilation(seeds, self.structuring_element.value)
-
-            # get the number of objects from the distance-based or marker-based watershed run above
-            number_objects = skimage.measure.label(y_data, return_num=True)[1]
-
-            seeds_dtype = (numpy.uint16 if number_objects < numpy.iinfo(numpy.uint16).max else numpy.uint32)
-
-            # NOTE: Not my work, the comments below are courtesy of Ray
```
bethac07, line 551: I haven't tested it, but it's probably worth doing so to see if this "make things negative, then back to positive again" trick is still useful/helpful with the 2023 versions of these algorithms, and if so, taking what's in ~rows 553-579 and moving it into the new function.

callum-jpg:
The refactor in this PR has removed some code from the original watershed module implementation, detailed here: https://github.com/CellProfiler/CellProfiler/blob/fb45aa1c5d50fe88ae640a9be2ea0ebf111c1abf/cellprofiler/modules/watershed.py#L553

Testing with and without this FIFO method doesn't impact segmentation performance when assessed in 2D and 3D. Perhaps the FIFO code was useful in the original centrosome implementation of watershed before the skimage version was used?
SCR-20230224-d4s

<details>
<summary>Code to compare FIFO watershed + plotting</summary>

```python
from typing import Literal
import skimage
import numpy 
import centrosome.cpmorphology
import mahotas
import scipy

def watershed_fifo(
    input_image,
    watershed_method: Literal["intensity", "distance", "markers"] = "distance",
    declump_method: Literal["shape", "intensity", None] = "shape",
    local_maxima_method: Literal["local", "regional"] = "local",
    intensity_image=None,
    markers_image=None,
    max_seeds: int = -1,
    downsample: int = 1,
    min_distance: int = 1,
    footprint: int = 8,
    connectivity: int = 1,
    compactness: int = 0,
    exclude_border: int = True,
    watershed_line: int = False,
    gaussian_sigma: int = 1,
    structuring_element: Literal[
        "ball", "cube", "diamond", "disk", "octahedron", "square", "star"
    ] = "disk",
    structuring_element_size: int = 1,
    fifo_labs=False,
):

    # Check inputs
    if input_image.dtype != bool:
        warnings.warn(
            "Watershed expects a thresholded image as input. Did you mean to use a boolean array?"
        )

    if (
        watershed_method.casefold() == "intensity"
        or declump_method.casefold() == "intensity"
    ) and intensity_image is None:
        raise ValueError(
            f"Intensity-based methods require an intensity image to be provided"
        )

    if watershed_method.casefold() == "markers" and markers_image is None:
        raise ValueError(
            "Markers watershed method require a markers image to be provided"
        )

    # Create and check structuring element for seed dilation
    strel = getattr(skimage.morphology, structuring_element.casefold())(
        structuring_element_size
    )

    if strel.ndim != input_image.ndim:
        raise ValueError(
            "Structuring element does not match object dimensions: "
            "{} != {}".format(strel.ndim, input_image.ndim)
        )

    if input_image.ndim > 2:
        footprint = numpy.ones((footprint, footprint, footprint))
    else:
        footprint = numpy.ones((footprint, footprint))

    # Downsample input image
    if downsample > 1:
        input_shape = input_image.shape
        if input_image.ndim > 2:
            # Only scale x and y
            factors = (1, downsample, downsample)
        else:
            factors = (downsample, downsample)

        input_image = skimage.transform.downscale_local_mean(input_image, factors)

    smoothed_input_image = skimage.filters.gaussian(input_image, sigma=gaussian_sigma)

    # Calculate distance transform
    if declump_method.casefold() == "shape":
        # Holes in thresholded objects can negatively impact shape declumping, so fill them
        # Keep the original input_image to use as a mask later (and thus reverse hole-filling)
        input_image_filled = skimage.morphology.remove_small_holes(input_image)
        distance = scipy.ndimage.distance_transform_edt(input_image_filled)
    else:
        distance = scipy.ndimage.distance_transform_edt(smoothed_input_image)

    # Generate alternative input to the watershed based on declumping
    if declump_method.casefold() == "shape":
        # Invert the distance transform of the input image.
        # The peaks of the distance tranform become the troughs and
        # this image is given as input to watershed
        watershed_input_image = -distance
        # Move to positive realm
        watershed_input_image = watershed_input_image - watershed_input_image.min()
    elif declump_method.casefold() == "intensity":
        # Convert pixel intensity peaks to troughs and
        # use this as the image input in watershed
        watershed_input_image = 1 - watershed_input_image
    else:
        # No declumping
        watershed_input_image = input_image

    # Generate markers
    if watershed_method.casefold() == "distance":
        # Find maxima in the distance transform
        if local_maxima_method.casefold() == "local":
            seed_coords = skimage.feature.peak_local_max(
                distance,
                min_distance=min_distance,
                footprint=footprint,
                num_peaks=max_seeds if max_seeds != -1 else numpy.inf,
            )
            seeds = numpy.zeros(distance.shape, dtype=bool)
            seeds[tuple(seed_coords.T)] = True
        elif local_maxima_method.casefold() == "regional":
            seeds = mahotas.regmax(distance, footprint)
        else:
            raise NotImplementedError(
                f"local_maxima_method {local_maxima_method} is not supported."
            )

    elif watershed_method.casefold() == "intensity":
        # Find markers based on intensity of the intensity image
        if local_maxima_method.casefold() == "local":
            seed_coords = skimage.feature.peak_local_max(
                intensity_image,
                min_distance=min_distance,
                footprint=footprint,
                num_peaks=max_seeds if max_seeds != -1 else numpy.inf,
            )
            seeds = numpy.zeros(distance.shape, dtype=bool)
            seeds[tuple(seed_coords.T)] = True
        elif local_maxima_method.casefold() == "regional":
            seeds = mahotas.regmax(distance, footprint)
        else:
            raise NotImplementedError(
                f"local_maxima_method {local_maxima_method} is not supported."
            )
    elif watershed_method.casefold() == "markers":
        # The user has provided their own markers
        seeds = markers_image
    else:
        raise NotImplementedError

    # Seed dilation
    seeds = skimage.morphology.binary_dilation(seeds, strel)

    number_objects = skimage.measure.label(seeds, return_num=True)[1]

    seeds_dtype = numpy.uint16 if number_objects < numpy.iinfo(numpy.uint16).max else numpy.uint32

    # Label seeds
    # markers = scipy.ndimage.label(seeds)[0]
    seeds = scipy.ndimage.label(seeds)[0]

    # Run inverted labels
    if fifo_labs:
        markers = numpy.zeros_like(seeds, dtype=seeds_dtype)
        markers[seeds > 0] = -seeds[seeds > 0]
    else:
        markers = seeds

    # Run watershed
    watershed_image = skimage.segmentation.watershed(
        watershed_input_image,
        markers=markers,
        mask=input_image != 0,
        connectivity=connectivity,
        compactness=compactness,
        watershed_line=watershed_line,
    )

    if fifo_labs:
        # Copy the location of the "background"
        zeros = numpy.where(watershed_image == 0)
        # Re-shift all of the labels into the positive realm
        watershed_image += numpy.abs(numpy.min(watershed_image)) + 1
        
        # Re-apply the background
        watershed_image[zeros] = 0
        # watershed_image = scipy.ndimage.label(watershed_image)[0]
        watershed_image = skimage.measure.label(watershed_image)

    # Reverse downsampling
    if downsample > 1:
        watershed_image = skimage.transform.resize(
            watershed_image, input_shape, mode="edge", order=0, preserve_range=True
        )
        watershed_image = numpy.rint(watershed_image).astype(numpy.uint16)

    if exclude_border:
        watershed_image = skimage.segmentation.clear_border(watershed_image)
    
    return watershed_image


import skimage
import numpy

image = skimage.img_as_float(skimage.data.human_mitosis()[:128, :128])
thresh = skimage.filters.threshold_otsu(image)
th_image = image > thresh
th_image = scipy.ndimage.label(th_image)[0]

fifo_input = numpy.zeros_like(th_image)
fifo_input[th_image > 0] = -th_image[th_image > 0]

fig, ax = plt.subplots(1, 3, figsize=(15, 5))
ax[0].imshow(image)
ax[1].imshow(th_image)
ax[2].imshow(fifo_input)

ax[0].set_title("Input Image")
ax[1].set_title("Labelled Input")
ax[2].set_title("FIFO (inverted) Input")
```
</details>

callum-jpg:
Explanation of seed generation for watershed in this PR:

There are currently two method for seed generation: local or regional maxima (names subject to change). These seeds can be calculated from either a distance transform of the thresholded input image, or from a user provided intensity image.

However, what is the difference between local and regional maxima? Behind the scenes, it's a call to either skimage `peak_local_max` for local or mahotas `regmax` for regional. Both maxima determining functions are passed a **footprint** argument.

So, what does regional provide that local doesn't? Ultimately, regional is a great starting point for typical blob-like objects. Since the `mahotas.regmax` can be seen as a more *automatic* seed generation method, a user doesn't have to play around with the extra settings that come with the local seed generation method. The downside: it's slower than local.

The local seed method provides users with more granular control. It is possible to achieve similar, or even better, seed generation with local, but it does require more playing around with. `mahotas.regmax` performs a better clustering of maxima in distance transforms.

Here's an example of regional and local seeds and the watershed segmentation they produce for multiple footprint values
output


<details>
<summary>Here's the code for generating the above plot</summary>

```python
import skimage
import mahotas
import scipy
from cellprofiler.library.modules import watershed
import numpy as np

def label_shuffle(labels):
    unique = np.unique(labels)
    # Shuffle everything except the background at 0th
    np.random.shuffle(unique[1:])
    output = np.zeros_like(labels)
    for i, lab in enumerate(np.unique(labels)):
        output = np.where(labels == lab, unique[i], output)
    return output

# img = skimage.data.human_mitosis()[:64, :64]
# img = skimage.data.binary_blobs(256, blob_size_fraction=0.25, n_dim=2, seed=42)
# th_img = img > skimage.filters.threshold_otsu(img)

img = skimage.data.cells3d()[35, 1, ...]
denoised = scipy.ndimage.median_filter(img, size=3)
th_img = denoised > skimage.filters.threshold_li(denoised)
th_img = skimage.morphology.remove_small_holes(th_img, 20**3)

dist = scipy.ndimage.distance_transform_edt(th_img)

fp = [4, 8, 16, 32]

fig, ax = plt.subplots(len(fp) + 1, 4, figsize=(15, 20))

ax[0, 0].set_title(f"Input image")
ax[0, 1].set_title(f"Thresholded input")
ax[0, 2].set_title(f"Distance transform")
ax[0, 0].imshow(img)
ax[0, 1].imshow(th_img)
# ax[0, 1].axis("off")
ax[0, 2].imshow(dist)
ax[0, 3].axis("off")

for i, footprint in enumerate(fp):

    reg_ws, reg_seeds = watershed(
        th_img, seed_method="regional", footprint=footprint, downsample=1, return_seeds=True
    )
    reg_ws, reg_seeds = label_shuffle(reg_ws), label_shuffle(reg_seeds)

    local_ws, local_seeds = watershed(
        th_img, seed_method="local", footprint=footprint, downsample=1, return_seeds=True
    )
    local_ws, local_seeds = label_shuffle(local_ws), label_shuffle(local_seeds)

    ax[i+1, 0].set_title(f"regional watershed - footprint: {footprint}")
    ax[i+1, 1].set_title(f"regional seeds - footprint: {footprint}")
    ax[i+1, 0].imshow(reg_ws, cmap="jet")
    ax[i+1, 1].imshow(reg_seeds, cmap="jet")

    ax[i+1, 2].set_title(f"local watershed - footprint: {footprint}")
    ax[i+1, 3].set_title(f"local seeds - footprint: {footprint}")
    ax[i+1, 2].imshow(local_ws, cmap="jet")
    ax[i+1, 3].imshow(local_seeds, cmap="jet")

fig.tight_layout()
```
</details>

However, it's important to note that there is something extra going on behind the scenes in regional maxima. As you can see in the above image, some seed objects are labelled identically. This is a result of `seeds, _ = mahotas.label(seeds, numpy.ones((16, 16)))`, which was in the original watershed non-advanced module. This further helps cluster close regional maxima together to have the same label. Whereas in local maxima, this would be up to the user to define with the min_distance argument, for example.

So, what could regional and local be renamed to? Basic and advanced, respectively?

bethac07:
Thanks @callum-jpg !
>However, it's important to note that there is something extra going on behind the scenes in regional maxima. As you can see in the above image, some seed objects are labelled identically. This is a result of seeds, _ = mahotas.label(seeds, numpy.ones((16, 16))), which was in the original watershed non-advanced module. This further helps cluster close regional maxima together to have the same label. Whereas in local maxima, this would be up to the user to define with the min_distance argument, for example.

What does the above look like but without doing the `seeds, _ = mahotas.label(seeds, numpy.ones((16, 16)))` step in regional? 
Should we do it in both kinds of seed finding?
Should the 16 be parameterized?

bethac07:
(maybe a lightly smoothed version of the intensity image, that is after all what we do in IDPrimary)

bethac07:
Do those look less barfy with the image then being used for watershed, in addition to seed finding?

bethac07:
Cool, much less barfy, just wanted to make sure. Will add a big long comment later but for now just wanted to say thank you for all the work generating these, it was super helpful!

cellprofiler/library/functions/object_processing.py
```diff
@@ -221,6 +229,180 @@ def segment_objects(labels_x, labels_y, dimensions):
...
+    )
+
+    if strel.ndim != input_image.ndim:
+        raise ValueError(
+            "Structuring element does not match object dimensions: "
+            "{} != {}".format(strel.ndim, input_image.ndim)
+        )
+
+    if input_image.ndim == 3:
+        footprint = numpy.ones((footprint, footprint, footprint))
```
bethac07, line 280: Let's not overwrite a setting value with another thing with the same name, please!

bethac07:
Alright, promised longer comment:
- If we're going to offer a "no declumping" option, let's have it behave the same way as IDPrimary, which is not generate seeds at all and literally just run label on the thresholded image. Does that make sense?
- With the exception of edges, the equivalent mahotas and skimage functions also seem more or less equivalent - @gnodar01 and I just went down a bit of a rabbit hole, and in fact in this image, when skimage has edge exclusion turned off, at 65,356 possible pixels, the two functions literally only differed at one in terms of "is this thing a peak or not" using a footprint of 16 (and at a footprint of 4, it's still only 124). So I definitely don't think we NEED to offer both, let's just go with what is fastest, which is per this testing mahotas (the difference gets even starker for lots of centers - at a size of 4, it's 30x faster).
  -  If we think min_intensity or num_peaks (or any of the other special skimage settings, like the ability to pass in other labels and find a specified number of peaks per other set of labels or things like that) is valuable, we can offer both, maybe as "fast" and "configurable" or something like that.
- I think therefore the only thing that is a "setting", aside from whether or not to offer "fast vs configurable" is whether or not to run the extra labeling-similar-things-together. My guess is that functionally, this is going to end up doing the same thing as min_distance was doing in the skimage function (though since skimage says it has to be 2x that value, at a 0.5x multiplier). I think we probably do, and I think we want to call it min_distance, but let's just run it using the `seeds, _ = mahotas.label(seeds, numpy.ones((min_distance, min_distance)))` code. I DO think we want it configurable, because while 16 might be a good setting for nuclei, I suspect it's going to harm more than help in ie the ExampleSpeckles pipeline (or any other comparable pipeline with a lot of spots closer together).

What do you think @callum-jpg ?

[PeakFindingNotebook.pdf](https://github.com/CellProfiler/CellProfiler/files/11044577/PeakFindingNotebook.pdf)

callum-jpg:
1. For declumping, I 100% agree that it should behave similarly to IDPrimary
2. I think the comparison of regional vs local is unfair in the attached notebook since it was performed on an intensity image. In images like this, the "jump" between regions of pixels is too great, so regional performs like local. However, in the case of a distance transform, `mahotas.regmax` will be able to track the gradually changing pixels until it reaches background, allowing for a much wider neighbourhood to be considered rather than being constrained to within the footprint as in `peak_local_max`. The speed comparison between regional and local also changes markedly when a distance transform is used - in this scenario regional is slower, which is to be expected since it surveys a wider area.
3. I am not sure about the label footprint method. There are multiple ways within the module already to change the seed objects through maxima suppression methods like min_distance and binary dilation. Would an increased footprint in label be the same as an increased footprint in seed dilation? Though, I do think the former would be quicker.

bethac07:
>The speed comparison between regional and local also changes markedly when a distance transform is used - in this scenario regional is slower, which is to be expected since it surveys a wider area.

Thanks, on mobile at the moment but will check this ASAP.

>However, in the case of a distance transform, mahotas.regmax will be able to track the gradually changing pixels until it reaches background, allowing for a much wider neighbourhood to be considered rather than being constrained to within the footprint as in peak_local_max

How do you square that with [this experiment](https://github.com/CellProfiler/CellProfiler/pull/4665#issuecomment-1479734311 )? The output seems very similar across a range of sizes.

bethac07:
Here's the notebook for distance - you're right that mahotas is quite a bit slower here, but at a footprint of 16, the results are identical, and at a footprint of 32, skimage has 34 pixels of seeds to mahotas's 33. 

At inappropriately small (for this data set) footprints of size 4 and 8, skimage does start picking up more seeds than I think we want, but this is largely obviated by setting `threshold_rel` to something like 0.2 or 0.3 to keep it from just picking up weird crud. 

So we definitely can't just call mahotas "fast"; more thought is going to be needed here.

One last thing re: the current implementation - this is from the ExampleSpeckles example, showing why we definitely don't want a hardcoded value of 16 if we do a post-grouping step - it is taking 3 actual disparate thresholded objects and lumping them into one, which isn't ideal!

image


[PeakFindingNotebook_Distance.pdf](https://github.com/CellProfiler/CellProfiler/files/11045344/PeakFindingNotebook_Distance.pdf)

callum-jpg:
*"How do you square that with https://github.com/CellProfiler/CellProfiler/pull/4665#issuecomment-1479734311?"*
It's definitely harder to reconcile this from that image alone. Since every regional maxima that we propose is can be determined by the local method, it makes sense that we see them converging with certain parameters. Ultimately, I think that having more seed generation methods is good. Personally, when playing with the `skimage.data.cells3d()[:,1, ...]` in the GUI, you can get towards a good segmentation much quicker in 3D using regional than with local (this is using the same image in the linked plot, but the plot was in 2D only). Using local can be frustrating in this scenario - you change one parameter and it offsets another. Attached is the pipeline and image to recreate this.

[3d_watershed.zip](https://github.com/CellProfiler/CellProfiler/files/11045645/3d_watershed.zip)

As for naming, I felt like regional and local could be named as basic and advanced, respectively. Though, maybe it'd would be strange to have another parameter called advanced available inside the watershed advanced.

bethac07:
>Personally, when playing with the skimage.data.cells3d()[:,1, ...] in the GUI, you can get towards a good segmentation much quicker in 3D using regional than with local (this is using the same image in the linked plot, but the plot was in 2D only). Using local can be frustrating in this scenario - you change one parameter and it offsets another. Attached is the pipeline and image to recreate this.

Right, but is this with regional still having the additional grouping step that local doesn't have, or have you made local code changes not pushed here where you've tested removing that grouping step and still found that? The point I'm trying to make is that I think that most of the differences that you're experiencing in the behavior between "regional" and "local" CellProfiler modes to me seem likely to be driven by _that additional grouping step_, not by the actual choice of `mahotas.regmax` vs `skimage.peak_local_max`, because the notebooks and your tests above show that those two algorithms head to head come up with very similar if not identical results when set in the appropriate regimes. I'm fine offering more options, I just haven't seen much that convinces me that the `mahotas.regmax` vs `skimage.peak_local_max` algorithms produce differentiable enough results to be worth making users have to choose between them. Is that more clear?

callum-jpg:
The above pipeline was tested without the 16 footprint labelling of proposed seeds - sorry I didn't include this information. The same `scipy.ndimage.label` was used for both.

I took a look at how the number of proposed seeds changes between regional and local for both 2D and 3D, in addition to the sum of the bitwise_xor. At higher footprint values, their seed proposal aligns (regional becomes considerably slower). Though, what you can see with regional is that it's a lot less volatile than local, which can be nice for a naive user. However, with changes to other parameters in local you could achieve the same performance at the same footprint values.

**2D image** with `skimage.data.cells3d()[35, 1, ...]`
2d_comparison

**3D image** with `skimage.data.cells3d()[:,1, ...]`
3d_comparison

If I had to choose a single maxima method, I'd choose local. Currently in the module, if advanced mode is off, regional is the method that is used with the user able to change the footprint only. With local, you could have something similar, only allowing the user to change the footprint. When in advanced mode, then they could then change the other parameters that local has.

callum-jpg:
I performed some speed tests for regional vs local on either a distance transform or an intensity image. As was seen in the notebooks above, regional is faster for an intensity image whereas local is faster for a distance transform. (also, 

| image                                      | image_type         | local_time | regional_time | footprint             |
|--------------------------------------------|--------------------|------------|---------------|-----------------------|
| skimage.data.cells3d()[:,1,...]_(3D)        | intensity          | 235 ms     | 168 ms        | numpy.ones((8, 8, 8)) |
| skimage.data.human_mitosis()_(2D)           | intensity          | 1.58 s     | 16.6 ms       | numpy.ones((8, 8))    |
| skimage.data.gravel()_(2D)                  | intensity          | 45.6 ms    | 5.37 ms       | numpy.ones((8, 8))    |
| skimage.data.lily()_(2D, converted to gray) | intensity          | 108 ms     | 16.2 ms       | numpy.ones((8, 8))    |
| skimage.data.cells3d()[:,1,...]_(3D)        | distance_transform | 152 ms     | 15.8 s        | numpy.ones((8, 8, 8)) |
| skimage.data.human_mitosis()_(2D)           | distance_transform | 12.4 ms    | 123 ms        | numpy.ones((8, 8))    |
| skimage.data.gravel()_(2D)                  | distance_transform | 75 ms      | 7.13 ms       | numpy.ones((8, 8))    |
| skimage.data.lily()_(2D, converted to gray) | distance_transform | 226 ms     | 20.8 ms       | numpy.ones((8, 8))    |


If we assume that the most common watershed method that's used is distance, then I think local makes sense. Also, some the test times were kinda wild when compared to the notebooks above and I don't know why - I think this is likely more specific to my system than anything.

callum-jpg:
Comparison of regional vs local on elongated bacteria. This aim of this test was to determine if regional can trace out much longer, non-cell-like objects outwith the provided footprint.

 **Regional at a footprint of 8**
SCR-20230324-gq2

**Local at a footprint of 8**
SCR-20230324-fwl

**Regional at a footprint of 64**
SCR-20230324-gr6

**Local at a footprint of 64**
SCR-20230324-grg


Here's the image and pipeline used: [elongated_bacteria.zip](https://github.com/CellProfiler/CellProfiler/files/11064197/elongated_bacteria.zip)

gnodar01 (approved):

PR #4671, opened 2022-11-17
adding a padding module

Mas515:
This module is a work in progress. The image dimensions are not updated so the next module fails. A use for this module is for example to pad a 3D z stack with 1 z slice at the top and 1 z slice at the bottom to allow the use of exclude object touching border without excluding objects present in the first and last z slices.

### DIFF
--- a/cellprofiler/modules/__init__.py
+++ b/cellprofiler/modules/__init__.py
@@ -67,6 +67,7 @@ builtin_modules = {
     "opening": "Opening",
     "overlayobjects": "OverlayObjects",
     "overlayoutlines": "OverlayOutlines",
+    "padimage":"PadImage",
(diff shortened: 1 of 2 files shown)


ErinWeisbart:
Can now crop image to original size after padding which I believe is fully working for 2D but I don't think 3D cropping in Z is behaving as expected. 

I changed the input of pad parameters to only allow padding on one side of each axis because that made crop handling simple (since if someone is padding both sides of an axis then crop would crop them right back off or get into some tricky behavior involving centering where the smaller crop was lost and bigger crop partially lost which just seemed really messy). I suppose we could make the crop logic a lot fancier, allowing selection of justification or centering for crop, if someone wants to argue we should do that....

PR #4678, opened 2022-11-29
Fill readers

gnodar01:
Requires CellProfiler/core#130, invokes ~`fill_readers()`~ `activate_readers(check_config=True)` **after** headless mode is determined. This is especially important when running tests (which use headless mode) in github actions since the `get_config()` function in CellProfiler/core/cellprofiler_core/preferences/__init__.py will import `wx` if it does not detect headless mode, and therefore cause the tests to exit prematurely because a framework build of python is required.


gnodar01:
@DavidStirling Do you think there's any important reason **not** to delay invoking `fill_readers()`?

If so, we could alternatively install `python.app` in the github workflow, when macos is used, and invoke pytest with it. That will get around the issue with running tests in github actions, but still seems wrong to me.

PR #4681, opened 2022-11-30
Scyjava Refactor

gnodar01:
Require CellProfiler/core#133


cellprofiler/gui/omerologin.py
```diff
@@ -2,7 +2,9 @@
 """omerologin - dialog box to capture login credentials for Omero
 """
 
-import bioformats.formatreader
+#TODO: this file uselss until CellProfiler/CellProfiler#4684 is resolved
```
bethac07, line 5: Could we therefore just comment the whole thing out for now?

cellprofiler/knime_bridge.py
```diff
@@ -54,6 +53,8 @@ def ZmqFrame(data=""):
 CLEAN_PIPELINE_REQ_1 = "clean-pipeline-request-1"
 CLEAN_PIPELINE_REPLY_1 = "clean-pipeline-reply-1"
 
+LOGGER = logging.getLogger(__name__)
+
 
 class KnimeBridgeServer(threading.Thread):
```
bethac07, line 59: I'm tempted to say let's just comment all this out until we deal with #4689 but that doesn't necessarily have to happen here. But since this removes the javabridge functionality without replacing it with anything, I feel like we can pretty confidently say if the bridge wasn't ALREADY broken (which I'm pretty sure it was), it's definitely likely to be now.

environment.yml
```diff
@@ -0,0 +1,28 @@
+# run: conda env create
```
bethac07, line 1: FWIW - we've always avoided putting a conda env in the main repo (we have one currently in the wiki only) because we haven't wanted to commit to keeping such a thing up to date. We have enough conda users in the group now that it's probably fine, but I wanted to at least flag this here.

bethac07 (approved):
As in the companion PR, I've made some small notes here, but otherwise, fantastic!

PR #4694, opened 2022-12-12
Adding FillObjects to library

callum-jpg:
Adding FillObjects to the library. This one works in a way that is slightly different to the other modules added so far due to `FillObjects` class inheriting from the `ObjectProcessing`. Basically, the library function needs to be passed as a lambda function to the `self.function` attribute and allows us to keep everything that `ObjectProcessing` does.


bethac07 (approved):
Nice!

PR #4705, opened 2022-12-19
Add savecroppedobjects to library

callum-jpg:


bethac07:
Can we have this in maybe a file_processing section of library/functions instead? It's not technically image processing, I think that will help keep us more organized.

Otherwise, looks in pretty good shape!

bethac07:
Need to check if all the test failures are due to this; https://github.com/CellProfiler/core/pull/139#issuecomment-1358228792 - presumably, we need to fix that and then rebase this again

callum-jpg:
In NumPy 1.24 `np.bool` has been removed https://github.com/numpy/numpy/releases/tag/v1.24.0, which is used in centrosome

gnodar01:
https://github.com/CellProfiler/core/pull/140

bethac07:
@callum-jpg please do rebase onto master once #4710 is pulled, thanks!

Issue #4693, opened 2022-12-09
Pipeline upgrade fails for pipelines previously using standard metadata conventions


bethac07:
Previously, CellProfiler used "C" for channel by default, T for timepoint, Z for Z plane. If you used the standard constructors in other places (like in NamesAndTypes or the SaveImages modules where setting to use a specific piece of metadata in subfolder creation), T and Z now seem to be the same, but "C" is now "Channel", so pipelines like the [linked one] (3D example 1 from the InputModules tutorial) break.

We should either go back to C, or provide an upgrade path for this - probably since it touches a lot of different modules, it will need to be a bump of the pipeline version, OR it will need to bump every module that might read or write metadata (so all the input and output ones, plus potentially some others like CalculateMath).

bethac07:
This is a jamboree-branch issue, not master

PR #4706, opened 2022-12-20
Update metadata conventions

gnodar01:
PR addressing #4693, but also other general pipeline refactoring work.

Companion to PR CellProfiler/core#139.

Note that this builds off, and will be merged into, the not yet merged scyjava branch.


gnodar01:
Resolves #4693

PR #4707, opened 2022-12-21
Library enhanceedges

callum-jpg:


cellprofiler/library/functions/image_processing.py
```diff
@@ -17,3 +20,57 @@ def medial_axis(image):
     return skimage.morphology.medial_axis(image)
 
 
+def enhance_edges_sobel(image, mask=None, direction="all"):
+    if direction.casefold() == "all":
+        output_pixels = centrosome.filter.sobel(image, mask)
```
bethac07, line 25: Scipy has a Sobel filter, have we done any testing to compare results and/or performance?

cellprofiler/library/functions/image_processing.py
```diff
@@ -17,3 +20,57 @@ def medial_axis(image):
...
+
+def enhance_edges_log(image, mask=None, sigma=2.0):
+    size = int(sigma * 4) + 1
+    output_pixels = centrosome.filter.laplacian_of_gaussian(image, mask, size, sigma)
+    return output_pixels
+
+
+def enhance_edges_prewitt(image, mask=None, direction="all"):
+    if direction.casefold() == "all":
+        output_pixels = centrosome.filter.prewitt(image, mask)
```
bethac07, line 43: Same as above re: Scipy

cellprofiler/library/functions/image_processing.py
```diff
@@ -17,3 +20,57 @@ def medial_axis(image):
...
+    auto_low_threshold=True,
+    sigma=1.0,
+    low_threshold=0.1,
+    manual_threshold=0.2,
+    threshold_adjustment_factor=1.0,
+):
+
+    if auto_threshold or auto_low_threshold:
+        sobel_image = centrosome.filter.sobel(image)
+        low, high = centrosome.otsu.otsu3(sobel_image[mask])
```
bethac07, line 66: Definitely let's move to the skimage version (and/or our own!)

cellprofiler/library/functions/image_processing.py
```diff
@@ -17,3 +20,57 @@ def medial_axis(image):
...
+        if auto_threshold:
+            high_th = high * threshold_adjustment_factor
+        if auto_low_threshold:
+            low_th = low * threshold_adjustment_factor
+    else:
+        low_th = low_threshold
+        high_th = manual_threshold
+
+    output_pixels = centrosome.filter.canny(image, mask, sigma, low_th, high_th)
```
bethac07, line 75: Same comments as above, except this one is in skimage rather than scipy

bethac07:
Need to check if all the test failures are due to this; https://github.com/CellProfiler/core/pull/139#issuecomment-1358228792 - presumably, we need to fix that and then rebase this again

gnodar01:
> Need to check if all the test failures are due to this; [CellProfiler/core#139 (comment)](https://github.com/CellProfiler/core/pull/139#issuecomment-1358228792) - presumably, we need to fix that and then rebase this again

https://github.com/CellProfiler/core/pull/140

bethac07:
@callum-jpg please do rebase onto master once #4710 is pulled, thanks!

callum-jpg:
@bethac07 

Scipy is noticeably faster for both the Sobel and Prewitt filters (~7.6 μs vs ~52 μs), but SciPy by default does not handle masks, but skimage does. However, the skimage + mask implementation is slightly slower (~34 μs).

However, it seems that the way in which skimage and centrosome sobel filters handle masks (or the lack thereof) is different.
SCR-20230105-n27

Centrosome applies `binary_erosion` to inputs, regardless if a mask was provided (see [here](https://github.com/CellProfiler/centrosome/blob/0894ce58f6ff9b380335b26ca23a48d7906d1331/centrosome/filter.py#L580)), whereas the skimage implementation will only apply the binary erosion if a mask is present. This leads to non-masked images having the border pixels set as 0 in the centrosome implementation, as shown above. You can recreate the centrosome result by providing an entirely True mask to `skimage.filters.sobel`. 

The above is also true for the Prewitt filter. 

```
from centrosome.filter import sobel, prewitt
import scipy
import numpy as np
import skimage

rand_arr = np.random.random((10, 10))

fig, ax = plt.subplots(1, 4, figsize=(15, 5))

ax[0].imshow(rand_arr)
# sqrt of the sum of squares for h and vertical (axis 0 or 1) filters. 
ax[1].imshow(
    np.sqrt(scipy.ndimage.sobel(rand_arr, axis=0) ** 2 + scipy.ndimage.sobel(rand_arr, axis=1) ** 2))
ax[2].imshow(skimage.filters.sobel(rand_arr))
# ax[2].imshow(skimage.filters.sobel(rand_arr, mask=np.ones_like(rand_arr).astype(bool)))
ax[3].imshow(sobel(rand_arr))

ax[0].set_title("Input")
ax[1].set_title("Scipy Sobel")
ax[2].set_title("Skimage Sobel")
ax[3].set_title("Centrosome Sobel")
```

callum-jpg:
In the EnhanceEdges GUI module, there is a "Calculate Gaussian's sigma automatically" setting, which defaults to True. Do we want this value to default to true in CellProfiler library? In the GUI, it's not really that automatic either (perhaps it was intended to be at some point). Currently, it just sets the sigma to 1.0 for Canny or 2.0 for Laplacian of Gaussian algorithms

callum-jpg:
Here is a more biologically relevant comparison of the different sobel edge detecting implementations:
SCR-20230208-nos

In the above example you can see in the first row the comparison between the SciPy and skimage sobel filters. In the second row, a mask is passed to the skimage sobel filter, which makes a call to [this function](https://github.com/scikit-image/scikit-image/blob/00177e14097237ef20ed3141ed454bc81b308f82/skimage/filters/edges.py#L54) which then erodes masks. This creates a result similar to the centrosome implementation with the difference that skimage only applies the erosion if a mask is provided. The third row looks at smaller masks - you can see some mild pixel differences between the two images here.

Code to generate the above plot:
<details>
  <summary>Expand</summary>
    
    from centrosome.filter import sobel
    import scipy
    import numpy as np
    import skimage
    
    image = skimage.data.human_mitosis()[0:50, 0:50]
    image = skimage.img_as_float(image)
    mask = np.ones_like(image).astype(bool)
    small_mask = np.zeros_like(image, dtype=bool)
    small_mask[20:40, 30:50] = True
    
    # sqrt of the sum of squares for h and vertical (axis 0 or 1) filters. 
    scipy_sobel = np.sqrt(scipy.ndimage.sobel(image, axis=0) ** 2 + scipy.ndimage.sobel(image, axis=1) ** 2)
    skimage_sobel = skimage.filters.sobel(image)
    skimage_sobel_mask = skimage.filters.sobel(image, mask=mask)
    centrosome_sobel = sobel(image)
    centrosome_sobel_mask = sobel(image, mask=mask)
    
    skimage_sobel_small_mask = skimage.filters.sobel(image, mask=small_mask)
    skimage_sobel_manual_mask = skimage_sobel.copy()
    skimage_sobel_manual_mask[~small_mask] = 0
    
    fig, ax = plt.subplots(3, 3, figsize=(12, 12))
    
    ax[0][0].imshow(image)
    ax[0][1].imshow(scipy_sobel)
    ax[0][2].imshow(skimage_sobel)
    ax[1][0].imshow(skimage_sobel_mask)
    ax[1][1].imshow(centrosome_sobel)
    ax[1][2].imshow(centrosome_sobel_mask)
    ax[2][0].imshow(skimage_sobel_small_mask)
    ax[2][1].imshow(skimage_sobel_manual_mask)
    ax[2][2].set_axis_off()
    
    ax[0][0].set_title("Input image")
    ax[0][1].set_title("Scipy Sobel")
    ax[0][2].set_title("Skimage Sobel")
    ax[1][0].set_title("Skimage Sobel + mask")
    ax[1][1].set_title("Centrosome Sobel no mask")
    ax[1][2].set_title("Centrosome Sobel + mask")
    ax[2][0].set_title("Skimage Sobel - small mask")
    ax[2][1].set_title("Skimage Sobel - manual small mask")
    
</details>

bethac07:
@callum-jpg Can you add just a comment here to the PR documenting the answers chosen to the questions raised in the PR, for making it easy to see what happened for future us? Then ready to merge I think

callum-jpg:
@bethac07 At this stage, no changes have been made in terms of what edge filters we use. 

In this PR we considered switching from using the centrosome filters to those that are in skimage. We considered using the skimage function as it's slightly faster (~20us for Sobel) and would have more support. 

However, how centrosome and skimage handle masks is different. Centrosome erodes the a 1px border around the whole image with and without a mask applied. Skimage only does this when a mask is provided.

Perhaps it would be best to switch over to skimage edge filter in another PR.

bethac07:
>Perhaps it would be best to switch over to skimage edge filter in another PR.
Awesome, can you make a GH issue? We then in theory are ok to pull, though I suspect we may want to wait until the skimage 0.20 branch is pulled to minimize numbers of rebases.

bethac07 (approved):

PR #4710, opened 2023-01-05
Remove deprecated numpy aliases

gnodar01:
Complement to CellProfiler/core#140


gnodar01:
Ignore weirdly named commit comment above about recompiling centrosome. I copied and pasted the wrong thing.

bethac07 (approved):
Approved, assuming once the wx wheel FINALLY builds that the tests pass

PR #4716, opened 2023-01-24
Check for whether Channels are a measurement

bethac07:
Caused an error in some cases when the user was specify which channels to write into a properties file


bethac07:
I think there's a deeper issue here.

PR #4727, opened 2023-02-02
Don't check perimeter when area is 0

bethac07:
Companion to #4726 , if we decide there not to handle all 4 cases, just one, change that here too.


bethac07:
First close-and-reopen was because I thought you couldn't change the target branch (you can!), second was to trigger the tests, which after the change to the target branch just noped out.

PR #4739, opened 2023-03-02
Add Opening and Closing to library

callum-jpg:


cellprofiler/modules/opening.py
```diff
@@ -16,10 +16,10 @@
 
 """
 
-import numpy
-import skimage.morphology
 from cellprofiler_core.module import ImageProcessing
+from cellprofiler_core.setting import Binary
```
bethac07, line 20: Is this needed or accidental?

callum-jpg: Not required. I had played around with adding an optional yes/no to process planewise, but I thought it'd be easier to keep the workflow the same by inferring if closing/opening should be applied planewise based on the input structuring element. Ie. a 2d disk on a 3d image will be applied planewise whereas a ball will be applied in 3d

bethac07 (approved):

PR #4749, opened 2023-03-13
Move mysqlclient to an optional dep

JacobHayes:
The mysqlclient package is "tricky" to install (requires a compiler + mysql and python dev headers) and is only used for the ExportToDatabase functionality, which already guards against a missing `mysqlclient` installation:
https://github.com/CellProfiler/CellProfiler/blob/4bd869eef9b347a92e731f78b2269bf604ca4d27/cellprofiler/modules/exporttodatabase.py#L155-L163

Since the package is tricky to install and isn't _required_ at runtime, this PR proposes we move it to an optional dependency (installed via `pip3 install "CellProfiler[mysql]"`), easing installation for users who don't use mysql.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,7 @@ setuptools.setup(
     extras_require={
         "build": ["black", "pre-commit", "pyinstaller", "twine"],
         "docs": ["Sphinx>=3.1.1", "sphinx-rtd-theme>=0.5.0"],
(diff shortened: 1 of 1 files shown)


JacobHayes:
Closing in favor of https://github.com/CellProfiler/CellProfiler/pull/4750 (hope that one is acceptable)

Issue #4432, opened 2021-07-15
Matplotlib version requirement


vkhodygo:
**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04 VM
 - CellProfiler Version 4.2.0
 - source
 - pypi

**Describe the issue**
When I follow the installation instructions I get an error saying that my `matplotlib` version is not correct.
I have to install `matplotlib==3.1.3` after that manually. 

**Error messages**
Generic pip error message saying that my version is incompatible. 

**Additional context**
Both `cellprofiler` and `centrosome` have their `matplotlib`versions specified. On a side note here, there is no need to have `matplotlib` in the `cellprofiler` requirements list since `centrosome` is a dependency of `cellprofiler` and already requires `matplotlib` installation. That would reduce maintenance work a bit.

iosonofabio:
Also, I have it working now with `matplotlib 3.4.3` and it works just fine

gnodar01:
Closing as completed (hopefully)

> On a side note here, there is no need to have matplotlib in the cellprofiler requirements list since centrosome is a dependency of cellprofiler and already requires matplotlib installation.

We do, since we want to have the ability to do version constraining differently between the two.

PR #4750, opened 2023-03-13
Loosen matplotlib pin and add gui and mysql extras

JacobHayes:
The `mysql` and GUI deps are trickier to install (notably, `mysqlclient` and `wxPython`), but are not needed for sqlite exports or headless mode. This moves those packages to extras, easing the headless installation.

Additionally, `matplotlib` 3.1.3 was released in Feb 2020 and does not provide wheels for newer python versions, making installation difficult. [`cellprofiler-core` uses `>=3.1.3`](https://github.com/CellProfiler/core/blob/be7ee8f77a0e5320095db0942f335a4e2ea664e6/setup.py#L31) and other users seem to have [no issues with newer versions](https://github.com/CellProfiler/CellProfiler/issues/4432#issuecomment-919716542), so this relaxes the constraint to `>=3.1.3`.

This might reduce some of the other GH issues trying to install things (assuming they're not using GUI stuff)?


psobolewskiPhD:
Gentle bump:
The mysqlclient pin is the only thing blocking pretty easy installation with a conda environment.yml for example on macOS.
conda-forge has mysqlclient back to 2.0.3 but not back to 1.4.x
So either making that an optional dependency or loosening the pin would be really helpful!
❤️

gnodar01:
@bethac07 What do you think about this PR? I've re-done this branch in PR #4825 so that it's rebased on master and includes #4815 and #4818, but figure we should keep the discussion here.

Re. the comment by @psobolewskiPhD, mysql has been updated to `~=2.0.0` in #4815, which allows for `2.0.3`.
Re. changing to `matplotlib>=3.1.3`, that should be fine.

However adding `gui` and `mysql` extras, and moving currently required deps to them may not be ideal. It's certainly nice to be able to have a way to install cellprofiler from PYPI without them, but the majority of people installing CellProfiler from PYPI want those extras, and it would require those people to now have to always run `pip install cellprofiler[guid,mysql]` instead of just `pip install cellprofiler`.

I've tried to think of some alternatives, but due to the state of python packaging there is currently no way to have an *excludes* extra, such that we can do `pip install cellprofiler[no_gui, no_mysql]`, and there's also no way to specify a *default* extra such that `pip install cellprofiler` would implicitly include `[gui, mysql]` unless otherwise specified by an explicit extra tag. There are open issues by people asking for both of those things, as well as a [talk on the subject](https://speakerdeck.com/uranusjr/lets-fix-extras-in-core-metadata-3-dot-0) at [PyCON Packaging Summit](https://us.pycon.org/2022/events/packaging-summit/) last year, but they're not possible at the moment (AFAICT). 

Maybe we could wrap this in the CI/CD monorepo refactor so that we have 3 PYPI packages, `cellprofiler-core`, `cellprofiler-headless`, and `cellprofiler`, with `cellprofiler` depending on both of the prior ones, and also `mysql` and `gui` deps?

bethac07:
Yeah, while I'm sympathetic to wanting to just install headless (having just this weekend spent a couple of hours messing with linux dependency versions then waiting literally overnight for the wxPython wheel to build on a system), I'm hesitant to move away from `pip install cellprofiler` not giving you literally everything the main CellProfiler program needs to run, especially as we're seeing more interest in not-super-computational people doing it to get plugin access. We could make it [so at least there's an `all` option](https://stackoverflow.com/a/64791668), and maybe `pip install cellprofiler[all]` isn't too bad, but again, I worry about complicating things for the people who need things to be simplest. 

I think I like your strategy of making a `cellprofiler-headless` package as a good least-worst option, but don't forget, there's also going to be `cellprofiler-library` there too. I do wonder how, from a practical point, we'd want to split up the code in such a package - will `cellprofiler-headless` and `cellprofiler` literally package up all the same exact files, just with different dependency settings, and how easy is it to do that, practically? Or will we literally abstract out all the GUI files, and how complex does that make our new monorepo? Not saying no to anything here, just trying to make sure we've fully framed out the problem before we decide.

gnodar01:
> I do wonder how, from a practical point, we'd want to split up the code in such a package - will cellprofiler-headless and cellprofiler literally package up all the same exact files, just with different dependency settings, and how easy is it to do that, practically? Or will we literally abstract out all the GUI files, and how complex does that make our new monorepo?

I am thinking

```
CellProfiler/
|
+--core/
|      |
|      +--setup.py <- builds `cellprofiler-core` 
|
+--library/
|      |
|      +--setup.py <- builds `cellprofiler-library`
|
+--cellprofiler/
|      |
|      +--headless_setup.py <- builds `cellprofiler-headless`
|      |
|      +--gui_setup.py <- builds `cellprofiler`
|
+--build_all.sh
```

Where `headless_setup.py` has:

```
...
    extras_require={
       "mysql": ["mysqlclient~=2.0.0"],
    },
    install_requires=[
        "boto3~=1.28.41",
        "cellprofiler-core==5.0.0",
        "centrosome==1.2.2",
        "h5py~=3.6.0",
        "imageio~=2.31.3",
        "inflect~=7.0.0",
        "joblib~=1.3.2",
        "mahotas~=1.4.13",
        "matplotlib>=3.1.3",
        "numpy~=1.24.4",
        "Pillow~=10.0.0",
        "pyzmq~=22.3.0",
        "requests~=2.31.0",
        "scikit-image~=0.20.0",
        "scikit-learn~=1.3.0",
        "scipy>=1.9.1,<1.11",
        "tifffile>=2022.4.8,<2022.4.22",
    ],
...
```

Whereas `gui_setup.py` has:

```
    extras_require={
        "build": ["black", "pre-commit", "pyinstaller", "twine"],
        "docs": ["Sphinx>=3.1.1", "sphinx-rtd-theme>=0.5.0"],
        "test": ["pytest~=7.4.1", "pooch>=1.3.0", "six~=1.16.0"],
    },
    install_requires=[
        "Jinja2~=3.1.2",
        "docutils==0.15.2",
        "scyjava>=1.9.1",
        "sentry-sdk==0.18.0",
        "wxPython==4.2.0",
        "cellprofiler-headless==5.0.0; extra == 'mysql'"
    ],
```

bethac07:
Sold! Do we want to close this PR then, in favor of the eventual monorepo PR?

Issue #4747, opened 2023-03-04
Finalize grammar for library functions/modules


bethac07:
We have quite a bit of mixed `thisfunction` and `this_function`, let's decide on one (or at least consistently where we used one vs the other) and then go through and clean up.

callum-jpg:
I think the module level functions should be named in the style of CellProfiler GUI, but without the capitalization of words to avoid confusion with python classes. As for the individual functions (ie. functions that are found within modules, like `shrink_to_point`), I think they should be named with standard python conventions: lowercase word or words separated by underscores.

callum-jpg:
To add to this how do we want to pass certain method options to library that contain spaces? For example, threshold has a method called "Minimum Cross-Entropy", should this be converted to `minimum_cross_entropy` or `minimumcrossentropy`? Currently, the former is what happens and I added this helper function to threshold (shown below), but it'd be much better living somewhere accessible for all modules since I'm running into the same space-removal problem in other modules. So, it would be great to standardize this grammar too.

```python
    def convert_setting(self, gui_setting_str):
        """
        Convert GUI setting strings to something cellprofiler
        library compatible. That is, remove spaces and hyphens.
        """
        rep_list = ((" ", "_"), ("-", "_"))
        converted_str = gui_setting_str
        for replacement in rep_list:
            converted_str = converted_str.replace(*replacement)
        return converted_str
```

PR #4758, opened 2023-03-28
Add MeasureImageOverlap to library

callum-jpg:
This is the first measurement module to be added to library and as a result has raised some questions:

- What should a measurement module return when used from library?
Here, the module returns a dictionary that can be easily added with calls to the dictionary like `data[FTR_F_FACTOR]`. This data can be easily added to measurements with the adaptive column names based on input images using: `m.add_image_measurement(self.measurement_name(FTR_F_FACTOR), data[FTR_F_FACTOR])`

- Where should the Objects class live?
In some of the functionality in MeasureImageOverlap there is a call to the ijv format of a labelmap (specifically in `compute_earth_movers_distance`), which assumes the label map is an instance of `cellprofiler_core.object.Objects()`. While CellProfiler Library could contain functionality that converts dense label maps to ijv formats, this is already provided by the Objects class in core. So, to solve this and minimally alter the actual image analysis code in this module, Library now possesses the Objects class. This is just a duplicate of the core code, but perhaps in time Library becomes a dependency of core (rather than the inverse if this change wasn't the case). I can also see the objects class being somewhat useful by itself if a user wants to easily switch between dense and ijv formats for their own labels. However, library will only add labels to the Objects class if the analysis functionality requires it

Additionally, this PR has added the `convert_setting` to utils of library. All this function does is to convert GUI setting strings with hyphens and spaces to underscores only. Functions like this are already present in threshold and expand or shrink objects, so it makes sense to have it in one place. This change is in part depending upon the decision in #4747


gnodar01:
Latest change requires Cellprofiler/core#142

gnodar01:
Tests should pass once CellProfiler/core#142 is merged

gnodar01:
Looks like we'll have to hold off on merging this PR until after factoring out Library into its own package. Since `Objects` and `Segmentation` classes really should live in library, but pieces of core need them, we need library to be a dependency of core (obviously without having cellprofiler be a dependency of core).

We should make it a priority to move to a monorepo structure sooner rather than later.

gnodar01:
I think simply moving `Objects` and `Segmentation` classes to library and pulling the PR won't be so simple. I was resisting the (strong) urge to refactor it all until now, when I noticed axis ordering is not respected between the two classes, see [here](https://github.com/CellProfiler/CellProfiler/blob/main/src/subpackages/core/cellprofiler_core/object/_segmentation.py#L63) and [here](https://github.com/CellProfiler/CellProfiler/blob/main/src/subpackages/core/cellprofiler_core/object/_objects.py#L91). I guess it's eventually reconciled *somewhere*, otherwise things wouldn't work as expected, but it at least needs cleaning.

In the spirit of refactoring, I'm thinking about making it a general rule that library is purely functional. I'm not totally set on it yet, but it would mean that the functions related to creating the dense & sparse label data structures (numpy arrays and records), transforming them from one to the other, and performing calculations on them (e.g. `make_ijv_outlines`) live in library, whereas the `Object` and `Segmentation` classes stay in core, where they maintain state, do cacheing, etc. That would however necessitate the refactoring of downstream functions that have also been moved to library, e.g. `mesurement.py -> calculate_earth_movers_distance` to no longer use the objects, "losing out" on stateful manipulations, cacheing, etc. For that particular example, no big deal, but thinking in the future I'm not sure how much this rule would get in our way.

bethac07:
... oof.  

... ooof.

... oooof.

I agree in principle with your proposed distinctions - functions live in the library, where of course core classes can import them when they need, but that classes are only in core. I think there actually won't be so, so, so many, because Image and Object/Segmentation should be the only classes that do behavior that we then want to be in the library, but those are extremely famous last words.

callum-jpg:
I really like adding the label processing functions like this to library.

With that, what type of objects should be passed to library functions (ie. in what format)? Are the added functions able to reliably detect, validate and coerce different label maps (eg. 2D, 3D, 2D+overlapping) into the IJV format and back, or are there particular detection/conversion routes that are not possible/dodgy? Should there be an extra argument in the function that indicates overlapping objects are to be expected? 

To sidestep the above: for modules that do return overlapping objects, what format should they be in? Will it be unique to that module, or should all object returning modules return them in the same form for consistency (eg. all modules use and produce dense matrices)?

gnodar01:
> With that, what type of objects should be passed to library functions (ie. in what format)?

It really depends on the use case. 

`sparse` is less memory intense than `dense`, so it should be used where possible, but it's a pretty limited format with just a 1-d array of tuples, so you can't do matrix math, fancy indexing, and the like. If that stuff is needed, `dense` can be used.

`labels` is just dense, but with no support for `c`, `z`, or `t`, and no overlapping, but `label_set` can be used for overlapping.

`ijv` is a silly format that is around to maintain any existing functionality that relies on it.

> Are the added functions able to reliably detect, validate and coerce different label maps (eg. 2D, 3D, 2D+overlapping) into the IJV format and back, or are there particular detection/conversion routes that are not possible/dodgy?

Every conversion can reliably happen between all the formats, and if something goes wrong it's a bug that should be fixed. The only caveat is a single `labels` matrix doesn't support overlap, so `label_set` should be used.

> Should there be an extra argument in the function that indicates overlapping objects are to be expected?

I don't think so, no. Up to the user to specify the format/converters that will work for them.

> To sidestep the above: for modules that do return overlapping objects, what format should they be in?

`sparse` if possible, `dense` if necessary. If `scipy.ndimage.label` returns a `labels` format, so sometimes that format is necessary, but it's easy to convert between `dense` and `labels` since that is just involves axis expanding/shrinking. Converting between the two sparse formats (`sparse`, `ijv`) to the two dense formats (`dense`, `labels`/`label_set`) is a bit more expensive.

> Will it be unique to that module, or should all object returning modules return them in the same form for consistency (eg. all modules use and produce dense matrices)?

If possible, only `sparse` would be used everywhere unless another format is absolutely required, but I haven't done a deep dive to see how plausible that would be.

gnodar01:
`make_ijv_outlines`, `relate_histogram`, `histogram_from_labels`, `histogram_from_ijv`, `center_of_mass` logic should also probably be abstracted out from `Objects` and into `library`, though not sure if the best place to put it is in `cellprofiler_library.functions.segmentation` or `cellprofiler_library.functions.object_processing` or some other place.

src/subpackages/core/cellprofiler_core/object/_segmentation.py
```diff
@@ -10,19 +14,34 @@ class Segmentation:
...
-        shape - the 5-D shape of the imaging site if sparse.
+        dense - A 6-d array composed of one or more 5-d integer labelings of
+        each hyper-voxel. The dimension order is labeling, c, t, z, y, x.
+        Typically, a 2-D non-overlapping segmentation has dimensions of
+        1, 1, 1, 1, y, x.
+
+        sparse - a labeling stored in a record data type with each column
+                 having a name of "c", "t", "z", "y", "x" or "label".
+                 The "label" column is the object number, starting with 1.
+                 When "c", "t", "z" are absent, this is the same as an ijv
```
gnodar01, line 29: "same as" should be "similar to" or "directly convertible to/from"

src/subpackages/core/cellprofiler_core/utilities/hdf5_dict.py
```diff
@@ -1457,13 +1459,15 @@ class HDF5ObjectSet(object):
...
+    A dense 6-d array composed of one or more 5-d integer labelings of
     each pixel. The dimension order is labeling, c, t, z, y, x. Typically,
     a 2-D non-overlapping segmentation has dimensions of 1, 1, 1, 1, y, x.
 
-    The i, j, v labeling of the pixels. The labeling is stored in a record
-    data type with each column having a name of "c", "t", "z", "y", "x" or
-    "label". The "label" column is the object number, starting with 1.
+    Sparse labeling stored in a record data type with each column having a name
+    of "c", "t", "z", "y", "x" or "label". The "label" column is the object
+    number, starting with 1. When "c", "t", "z" are absent, this is the same as
```
gnodar01, line 1468: "same as" is can be misconstrued. scipy does use ijv format, but wraps it in a special object. so should be "interoperable with" or "representable as", etc.

src/subpackages/core/cellprofiler_core/utilities/hdf5_dict.py
```diff
@@ -1475,12 +1479,12 @@ class HDF5ObjectSet(object):
...
-    AXIS_T = "t"
-    AXIS_Z = "z"
-    AXIS_Y = "y"
-    AXIS_X = "x"
+    AXIS_LABELS = SPARSE_FIELD.label.value
+    AXIS_C = SPARSE_FIELD.c.value
+    AXIS_T = SPARSE_FIELD.t.value
+    AXIS_Z = SPARSE_FIELD.z.value
+    AXIS_Y = SPARSE_FIELD.y.value
+    AXIS_X = SPARSE_FIELD.x.value
```
gnodar01, line 1487: Technically, these are not referenced anywhere in code, but are referenced in some docstrings. Left here for posterity to lower confusion, but really should update the docstrings.

PR #4771, opened 2023-04-27
Add ability to run planewise to UnmixColors

bethac07:
I haven't fully tested this out yet (hence draft), and I think the case it's covering is mostly a corner case, but AFAICT, there's no reason we shouldn't be able to support UnmixColors being 3D by running it planewise.

### DIFF
--- a/cellprofiler/modules/unmixcolors.py
+++ b/cellprofiler/modules/unmixcolors.py
@@ -386,7 +386,7 @@ blue absorbance values from the image.
             workspace.display_data.input_image = input_pixels
             workspace.display_data.outputs = {}
         for output in self.outputs:
-            if not image.volumetric:
+            if not input_image.volumetric:
                 image = self.run_on_output(input_pixels, output)
             else:
                 image = numpy.zeros_like(input_pixels)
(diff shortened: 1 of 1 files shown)


gnodar01:
@bethac07 When estimate custom absorbances from an image, do you think it would make sense to be able to provide a 2D **or** 3D image? If so, would we want to estimate a single set or RGB absorbance values from across the full volume, or estimate Z number of RGB absorbance values, and applying each plane-wise to the input image. My intuition is no, none of that would be useful, but I want to double check.

bethac07:
>When estimating custom absorbances from an image, do you think it would make sense to be able to provide a 2D or 3D image? 

Good question! When folks are doing that, in general, we want them to choose a tiny little crop that's just the one stain on its own, so for a tiny little crop, I think it's fine to have it 2D only.

In practice, this is covering a pretty corner case, because in general, one wouldn't do "true" 3D imaging of histology slides most of the time, because the whole point is that these sections and dyes are opaque, so you won't be able to image deeply or much. I think this is mostly not going to be used, or only used in the cases where someone has pre-aligned slides outside of CellProfiler. But we can see!

gnodar01 (approved):

PR #4781, opened 2023-05-24
Make module search fuzzy

gnodar01:
Module search uses exact substring matching, which is a bit too strict. We'd like to allow for spelling mistakes and skipping characters but still display sensible results ranked according to closest match.

For instance when searching for "IdentifyTertiaryObjects"

typing: `"zidentifytertiaryobjects"` should still work even with the leading `"z"` character

likewise typing `"idtertiary"` should work, where you type the leading `"id"`, skip over `"entify"`, followed by `"tertiary"`

This PR uses Damerau-Levenshtein distance similarity for fuzzy search. There's **a lot** of other choices, and a whole research field in this subject (eg driven by sequence alignment in genomics) as well as a bunch of custom open source solutions used in various tools (eg `fzf` or VSCode's file search). I was not at all rigorous in my choice, and I chose Damerau-Levenshtein simply because it seemed to match best for the small set of example queries I tried.

I also limited the results to `10` because it seemed "just right" to me. We could make this bigger/smaller, or a variable list by using some thresholding heuristic on the distance score.

### DIFF
--- a/cellprofiler/gui/addmoduleframe.py
+++ b/cellprofiler/gui/addmoduleframe.py
@@ -1,4 +1,8 @@
 import wx
+from rapidfuzz.utils import default_process
+from rapidfuzz.process import extract
+from rapidfuzz.distance import DamerauLevenshtein
+
 from cellprofiler_core.utilities.core.modules import (
     get_module_names,
     instantiate_module,
@@ -246,13 +250,22 @@ class AddModuleFrame(wx.Frame):
         ):
             self.__module_categories_list_box.Select(-1)
             self.__on_category_selected(None)
+
         keys = list(self.__module_dict["All"].keys())
-        keys = [
-            key for key in keys if self.search_text.GetValue().lower() in key.lower()
-        ]
+
+        # results in [('Module1', score1, idx1), ('Module2', score2, idx2), ...]
+        # where idx is the index of the result in `keys`
+        top_scorers = [m[0] for m in extract(
+            self.search_text.GetValue(),
+            keys,
+            processor=default_process,
+            scorer=DamerauLevenshtein.similarity,
+            limit=10
+        )]
+
         self.__module_list_box.Clear()
-        self.__module_list_box.AppendItems(sorted(keys))
-        if len(keys) > 0:
+        self.__module_list_box.AppendItems(top_scorers)
+        if len(top_scorers) > 0:
             self.__module_list_box.Select(0)
         else:
             self.__module_list_box.AppendItems("No matching modules")
--- a/cellprofiler/utilities/rules.py
+++ b/cellprofiler/utilities/rules.py
@@ -1,7 +1,8 @@
 """rules - code for parsing and applying rules from CPA
 """
 
-from difflib import get_close_matches
+from difflib import SequenceMatcher
+from rapidfuzz import process
 import re
 
 from cellprofiler_core.module import Module
@@ -87,13 +88,30 @@ class Rules(Module):
 
         @staticmethod
         def return_fuzzy_measurement_name(measurements,object_name,feature_name,full,allow_fuzzy,fuzzy_value=FUZZY_FLOAT):
+            def standard_ratio(query, candidate, **kwargs):
+                s = kwargs["SequenceMatcher"]
+                s.set_seq1(candidate)
+                return s.ratio()
+
             measurement_list = [f"{col[0]}_{col[1]}" for col in measurements]
             if allow_fuzzy:
                 cutoff = fuzzy_value
             else:
                 cutoff = 1
-            closest_match = get_close_matches('_'.join((object_name,feature_name)),measurement_list,1,cutoff)
-            if len(closest_match) == 0:
+
+            query = '_'.join((object_name,feature_name))
+            # closest_match = get_close_matches(query,measurement_list,1,cutoff)
+            s = SequenceMatcher(b=query)
+            closest_match = process.extractOne(
+                query,
+                measurement_list,
+                processor=None,
+                scorer=standard_ratio,
+                score_cutoff=cutoff,
+                scorer_kwargs={"SequenceMatcher": s}
+            )
+
+            if closest_match == None or len(closest_match) == 0:
                 return ''
             else:
                 if full:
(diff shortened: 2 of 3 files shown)


gnodar01:
`RapidFuzz` is certainly more performant in runtime, but the difference against `difflib` is not noticeable with the small amount of text being searched through.

The most basic use of `RapidFuzz` falls back to `difflib` which seems to use a modified version of Ratcliff-Obershelp distance measure. So I did try both the `RapidFuzz` fallback of `difflib` and tried using `difflib` directly without `RapidFuzz`.

In terms of results, the specific rankings returned by the two are qualitatively very different. I didn't like the results of Ratcliff-Obershelp-style (`difflib`) distance measures personally, so I wanted to try other distance measures and I settled with Damerau-Levenshtein because it seemed to perform better, subjectively.

---

Using `difflib` directly, you are required to pass in a threshold for the score to filter the results.

So for instance, when trying to get `IdentifyTertiaryObjects` to be close to the top:

```
In [28]: get_close_matches('idtert', keys, 10, 0.4)
Out[28]: 
['FilterObjects',
 'MedianFilter',
 'FlipAndRotate',
 'DilateObjects',
 'DisplayScatterPlot',
 'Watershed',
 'Tile',
 'SplitOrMergeObjects',
 'GaussianFilter']
```

Didn't even pass the (pretty low) threshold, much less near the top.

```
In [30]: get_close_matches('zidterti', keys, 10, 0.4)
Out[30]: ['FilterObjects', 'MedianFilter']
```

Misspelling means threshold of `0.4` is no good. When we drop it to `0.3`, the one we want is third in the list:

```
In [32]: get_close_matches('zidterti', keys, 10, 0.3)
Out[32]: 
['FilterObjects',
 'MedianFilter',
 'IdentifyTertiaryObjects',
 'DisplayScatterPlot',
 'FlipAndRotate',
 'DilateObjects',
 'SplitOrMergeObjects',
 'MakeProjection',
 'GaussianFilter',
 'Watershed']
```

It's hard to know what to hardcode the threshold to.

We can write our own variation of `get_close_matches` which still uses `difflib` but returns a static number of results rather than filtering, but I still don't personally like the ranking it returns.

Using the same queries above with Damerau-Levenshtein:

```
In [36]: process.extract('idtert', keys, processor=utils.default_process, scorer=distance.DamerauLevenshtein.similarity, limit=5)
Out[36]: 
[('IdentifySecondaryObjects', 6, 40),
 ('IdentifyTertiaryObjects', 6, 41),
 ('DisplayScatterPlot', 5, 20),
 ('FilterObjects', 5, 30),
 ('IdentifyDeadWorms', 5, 36)]
```

```
In [37]: process.extract('zidterti', keys, processor=utils.default_process, scorer=distance.DamerauLevenshtein.similarity, limit=5)
Out[37]: 
[('IdentifyTertiaryObjects', 6, 41),
 ('MeasureObjectIntensityDistribution', 6, 58),
 ('FilterObjects', 5, 30),
 ('IdentifyObjectsInGrid', 5, 37),
 ('IdentifySecondaryObjects', 5, 40)]
```

For a handful of other examples I tried, it just seems to consistently do better in terms of results. I didn't do a lot of digging into the algorithms themselves. I'm sure there are mathematical justifications for using one or the other algorithm, but going by intuition I preferred results from Damerau-Levenshtein most/all of the time.

gnodar01:
Actions are failing because MySQL is failing to install, not because of the code in the PR.

gnodar01:
Hmm, do you mean change the existing code to use `RapidFuzz` but keep the existing behavior (ie matching fuzzy string using the same ranking system as before), or do you mean change the existing code to use `RapidFuzz` and take advantage of alternative distance scores.

The former is straightforward, but to exactly mimic existing behavior, it's still using `difflib` just with extra steps. Below I've commented out `difflib`'s `get_close_matches` function, and instead implemented my own scorer, `standard_ratio` and passed it to `RapidFuzz`. The custom scorer uses `SequenceMatcher` which is imported from `difflib`, so that I can use `difflib`'s `ratio` metric (since the standard `ratio` metric is slightly different in `RapidFuzz`).

```python
from difflib import SequenceMatcher
from rapidfuzz import process

...

def return_fuzzy_measurement_name(measurements,object_name,feature_name,full,allow_fuzzy,fuzzy_value=FUZZY_FLOAT):
    def standard_ratio(query, candidate, **kwargs):
        s = kwargs["SequenceMatcher"]
        s.set_seq1(candidate)
        s.set_seq2(query)
        return s.ratio()

    measurement_list = [f"{col[0]}_{col[1]}" for col in measurements]
    if allow_fuzzy:
        cutoff = fuzzy_value
    else:
        cutoff = 1

    query = '_'.join((object_name,feature_name))
    # closest_match = get_close_matches(query,measurement_list,1,cutoff)
    s = SequenceMatcher()
    closest_match = process.extractOne(
        query,
        measurement_list,
        processor=None,
        scorer=standard_ratio,
        score_cutoff=cutoff,
        scorer_kwargs={"SequenceMatcher": s}
    )

    if closest_match == None or len(closest_match) == 0:
        return ''
    else:
        if full:
            return closest_match[0]
        else:
            return closest_match[0][len(object_name)+1:]
```

The latter is certainly doable, and would do away with the use of `difflib` altogether, but would require you and the image analysts to weight in with opinions on what distance metric would be most suitable in the `Rules` class. In some cases, different distance metrics would result in different top scorers (depending on how far apart the query and candidate strings tend to be when using `FilterObjects` and the like).

gnodar01:
Notebook if you want to play around with the different search algorithms: https://drive.google.com/file/d/10RHov8cCR4pqjSNIRhho8rP6QDRzJphL/view?usp=sharing

gnodar01:
@bethac07 Do you think this is good to go? I did end up refactoring `rules.py` to use rapidfuzz, while keeping identical behavior as before.

bethac07 (approved):

bethac07:
Yup! I was being a little nitpicky, because I think the current method doesn't surface "correct" items fully as I would expect (you can see below that even several letters into Identify, it's still got non-Identify things as the "top hit"), but since we're showing a top 10, I think it's fine, and definitely helpful for the case of typos, which wins out. 

image

PR #4783, opened 2023-06-13
Add Identify Primary, Secondary and Tertiary to CellProfiler Library

callum-jpg:
Outstanding todo:

- [ ] Add object processing functionality to `functions.object_processing` so that `cellprofiler.library.modules` is essentially just an API for `functions`. (This is implemented in principle for IDPrimary in the commented out code, but it was causing a circular dependency problem with threshold)
- [ ] Remove kwargs from threshold
- [ ] Add a `return_cp_output` arg to threshold. This will mean a user can use CellProfiler's threshold elsewhere and not have to deal with the return of unnecessary variables


src/subpackages/library/cellprofiler_library/functions/object_processing.py
```diff
@@ -451,3 +458,630 @@ def fill_convex_hulls(labels):
...
+        )
+        assert binary_maxima_image.shape[0] == image.shape[0]
+        assert binary_maxima_image.shape[1] == image.shape[1]
+
+    # Erode blobs of touching maxima to a single point
+    shrunk_image = centrosome.cpmorphology.binary_shrink(binary_maxima_image)
+    return shrunk_image
+
+
+def smooth_image(image, mask=None, filter_size=None, min_obj_size=10):
```
bethac07, line 520: This is image-in,image-out, it should go in image-processing

src/subpackages/library/cellprofiler_library/functions/object_processing.py
```diff
@@ -451,3 +458,630 @@ def fill_convex_hulls(labels):
                 cmask
             ] = label
     return output
+
+
+# Rename to get_maxima_from_foreground?
+def get_maxima(
```
bethac07, line 464: This is image-in,image-out, it should go in image-processing

bethac07:
So, big picture question - these currently break the convention of "functionality goes into library/functions, library/modules are just logic wrappers". These modules are the special-est of special cases, so I'm not saying we can't break the rules for them, but I'd be lying if I didn't say that IPO was one of my visions of "user just sits down with the library and stuff works".

If the concern is that it makes object processing too big or messy, what about having instead a thing for `segmentation` alongside `image_processing`, `object_processing`, etc? It breaks the CellProfiler user-facing convention, but it respects the underlying module class convention, so I can live with that.

gnodar01:
> So, big picture question - these currently break the convention of "functionality goes into library/functions, library/modules are just logic wrappers". These modules are the special-est of special cases, so I'm not saying we can't break the rules for them, but I'd be lying if I didn't say that IPO was one of my visions of "user just sits down with the library and stuff works".
> 
> If the concern is that it makes object processing too big or messy, what about having instead a thing for `segmentation` alongside `image_processing`, `object_processing`, etc? It breaks the CellProfiler user-facing convention, but it respects the underlying module class convention, so I can live with that.

Agreed. We already have a `functions/segmentation.py`, and I wouldn't even mind splitting functions between that and `object_processing` as long as it matches, semantically.

PR #4784, opened 2023-06-13
3D IdentifyPrimaryObjects

callum-jpg:


gnodar01:
@callum-jpg Would you mind adding a list of TODOs left on this before it's ready to move out of DRAFT stage?

callum-jpg:
@gnodar01 Of course. I was planning to take a look at this again within the next few weeks to finish it up, but ultimately I think the backbone of it is there. The real standout items that are wrong include: 
- The GUI module documentation/setting names. These need to be updated to better reflect the new options (eg. downsampling) 
- Wording needs changed in order to highlight that this is for 3D. 
- From an image analysis perspective, the usability needs to be checked to. Does it behave "similarly" enough to IDPrimary? Should better defaults be selected?

As an aside, I think IDtertiary would be simple to add in 3D, but secondary (eg propagation) might be harder. Though, maybe for secondary different methods could be used. Unsure if there is a imminent need for secondary and tertiary though. 

I'll put the todo on my todo.

Toodeloo

callum-jpg:
The biggest problem I see with this module is this: what is the volume of a cell? In IDPri2D, we assume cells are sort of circular, but what should be assumed in 3D? Spherical is the initial thought, but a lot of bioimages may not actually have a z-depth for the entire cell, sort of leading to a cylinder/ellipsoid shape (ie. top and bottom of cell are chopped off). If this is the case, maybe we should request the user defines the depth of the cell (or just assume it's the min_size or max_size if not provided).

Ultimately I think what this module is trying to do is Threshold and Watershed, which already exist in cellprofiler(library). Though, there is some extra functionality built in to this module (eg size filter). I like the this module since it means using one module rather than 2 or more (eg. median filter before threshold) and it feels a little more beginner friendly than watershed. So, maybe it should just be a clever stitching of existing library functions? (threshold -> watershed -> filter_on_size_3d(tbc))

If doing the above, you could determine an ideal `min_distance` arg value for watershed based on the user defined expected cell dimensions.

I think the above larger questions need to be answered before addressing the following todos for the PR as it currently stands:
- It's slow. Currently we have `self.low_res_maxima`, but could this be replaced with a tuneable downsampling (eg. like in watershed) that makes things much quicker.
- fill_holes is broken
	- `centrosome.cpmorphology.fill_labeled_holes` doesn't work on 3D, so I used `skimage.morphology.remove_small_holes`, which is less automatic and requires a definable max fill area threshold. 
		- With an accepted "min volume" of a cell, we could set the threshold to this, though maybe there's a weird edge case here
- `centrosome.cpmorphology.binary_shrink` doesn't work for 3D, so it's just not used here. Might not be needed though
- Structuring element is fixed as 3D in some places. This could be changed to be adaptive (eg 2d or 3d), allowing for this method to be used on 2D too
	- If so, name of module should be adapted. A 3D only module seems weird
- Add outlines and maxima to output figure plot
	- There's already code for this in watershed

Overall, I think IDPri3D would be amazing if it could segment `skimage.data.cells3d()[:,1,...]` intuitively

PR #4797, opened 2023-08-04
Bump centrosome

gnodar01:


gnodar01:
Pairs with CellProfiler/core#146 and CellProfiler/centrosome#116

bethac07:
@gnodar01 2 absolute value tests failing for Earth Mover Distance - I feel like we've seen that several times in the past when we change sklearn package versions, typically. I think we've just adjusted the test values at those points.

bethac07 (approved):

PR #4801, opened 2023-08-07
Tighten some pins, loosen others

bethac07:


bethac07:
Was passing before I moved the core requirement to 4.2.6, should pass again once https://github.com/CellProfiler/core/pull/149 pulls

gnodar01 (approved):

PR #4638, opened 2022-09-02
Dark mode support

gnodar01:
Fix for #4632 and #4627

PR #4803, opened 2023-08-08
Recreate #4638 in 4.2.x

bethac07:


bethac07:
@gnodar01 it wasn't clear to me how much the two "fix the welcome screen commits" were independent from the rest of that PR - it seemed like "not fully", so I went the opposite direction and started with the full PR and then fixed error messages until it opened. AFAICT, there's no landmines here, but can use the pair of eyes who actually did the original PR to give a second look.

gnodar01 (approved):

Issue #4717, opened 2023-01-24
ExportToDatabase issue when writing particular channels to a properties file


bethac07:
**Describe the bug**
H/T @sugan89!

When using ExportToDatabase in a situation where the user wants to specify which channels are used in the properties file and/or which color they're assigned (aka where `Include information for all images, using default values?` is No), in certain circumstances, ExportToDatabase failed with the stack trace below. It turns out it's due to not checking in that part of the code path whether the `"ExportToDb", "ExportToDb_Channels"` measurement was made - that measurement was added in #4396 . As far as I can tell, this measurement IS correctly written into the measurement tables during the run, but in the post_run (where the properties file is made, it's done. No idea why. I _assume_ we tested at the time that #4396 worked, but I can't get it to work in a couple of test cases in my hands.

**To Reproduce**
Run the QC tutorial at tutorials.cellprofiler.org

**Screenshots**
![image](https://user-images.githubusercontent.com/6721515/214383405-adeb7cb2-907e-45f9-b479-c58b610425b8.png)

**Desktop (please complete the following information):**
OSX 12.6, 4.2.4&4.2.5 built, 4.2.x source branch; Windows 4.2.5 built

**Additional context**
```
Failed to complete post_run processing for module ExportToDatabase.
Traceback (most recent call last):
  File "/Users/bcimini/Documents/GitHub/CellProfiler/core/cellprofiler_core/pipeline/_pipeline.py", line 1535, in post_run
    module.post_run(workspace)
  File "/Users/bcimini/Documents/GitHub/CellProfiler/CellProfiler/cellprofiler/modules/exporttodatabase.py", line 2857, in post_run
    self.write_properties_file(workspace)
  File "/Users/bcimini/Documents/GitHub/CellProfiler/CellProfiler/cellprofiler/modules/exporttodatabase.py", line 4148, in write_properties_file
    all_properties = self.get_property_file_text(workspace)
  File "/Users/bcimini/Documents/GitHub/CellProfiler/CellProfiler/cellprofiler/modules/exporttodatabase.py", line 4421, in get_property_file_text
    channels_list = workspace.measurements.get_measurement("ExportToDb", "ExportToDb_Channels")
  File "/Users/bcimini/Documents/GitHub/CellProfiler/core/cellprofiler_core/measurement/_measurements.py", line 854, in get_measurement
    vals = self.hdf5_dict[object_name, feature_name, image_set_number]
[...]

PR #4805, opened 2023-08-10
Replace measurement-based ExportToDB state transmission with shared_state implementation

DavidStirling:
Requires cellprofiler/core#147

Fixes #4717

Seeing people posting errors revolving around the faux measurements we used to fix #3991 makes me sad. 

To summarise, we needed to make a note of how many channels were in each image the user wanted to be listed in their CPA properties file. The problem was that we make that file after the run ends and workers shut down. At that point we don't have access to image providers and so needed to make a note of how many channels were in an image during the run. This was done by creating 'hidden' measurements to record that information, but some naughty modules still tried to grab those measures. There were also some rare settings combinations which would trip up the measurement generator.

Measurements were never an ideal solution to passing info between pipeline phases. With module state now working properly we can use that instead for a much cleaner fix.

PR #4827, opened 2023-09-19
Replace measurement-based ExportToDB state transmission with shared_state implementation

gnodar01:
Seems in addition to the other fix mentioned in #4805, it also fixes the issue #4695 solved. I've rebased this on master and also reverted the changes made there.

### DIFF
--- a/cellprofiler/modules/exporttodatabase.py
+++ b/cellprofiler/modules/exporttodatabase.py
@@ -168,6 +168,8 @@ except Exception:
 #
 ##############################################
 D_MEASUREMENT_COLUMNS = "MeasurementColumns"
+D_PROPERTIES_IMAGES = "PropertiesImages"
+D_PROPERTIES_CHANNELS = "PropertiesChannels"
 
 """The column name for the image number column"""
 C_IMAGE_NUMBER = "ImageNumber"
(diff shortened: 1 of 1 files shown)


gnodar01:
This supersedes #4805, so pasting @DavidStirling to reduce potential confusion:

Requires https://github.com/CellProfiler/core/pull/147

Fixes https://github.com/CellProfiler/CellProfiler/issues/4717

Seeing people posting errors revolving around the faux measurements we used to fix https://github.com/CellProfiler/CellProfiler/issues/3991 makes me sad.

To summarise, we needed to make a note of how many channels were in each image the user wanted to be listed in their CPA properties file. The problem was that we make that file after the run ends and workers shut down. At that point we don't have access to image providers and so needed to make a note of how many channels were in an image during the run. This was done by creating 'hidden' measurements to record that information, but some naughty modules still tried to grab those measures. There were also some rare settings combinations which would trip up the measurement generator.

Measurements were never an ideal solution to passing info between pipeline phases. With module state now working properly we can use that instead for a much cleaner fix.

gnodar01:
@bethac07 requesting you as reviewer since I (re)constructed this PR.

LGTM, but worth you taking a very quick look here, and in the corresponding CellProfiler/core#147, in case there is something in the ancient knowledge that I would overlook but catches your eye.

bethac07:
LGTM also but should probably test it out in practice - should be able to do in the background today, if you haven't already. 

> in case there is something in the ancient knowledge that I would overlook but catches your eye

Cracks about my age? Harsh but fair ;)

bethac07:
@DavidStirling , do you happen to remember, what data were we using to test this property file functionality that it actually did what it was supposed to? I just tried the PR and built 4.2.6 on a color image (built off of the UntangleWormsBrightfield example, but then simplified just in case that was why the pipeline was cranky) and in both cases, it's giving an identical properties file but it's wrong (says that channels_per_image is 1, image_channel_colors are gray, no blend modes, etc).

bethac07:
Nevermind, error was between keyboard and chair on my side- I was accidentally not running this PR (forgot I had CellProfiler aliased on terminal). 4.2.6 (built or source, can now confirm) is NOT doing the right thing, so I must have screwed it up in #4695 , but this does!

bethac07 (approved):

bethac07:
#4695 was I guess just pulled, so probably NOT what screwed it up, but something did along the way in one of these commits. In any case, probably not worth tracking down at this stage!

image

PR #4828, opened 2023-09-26
Fix ClassifyObjects error messages

DavidStirling:
Looks like a typo from the implementation of fuzzy matching, unfortunately this prevents the tooltip from telling the user which specific measure is missing.


gnodar01:
Bug introduced in #4575

bethac07 (approved):
Whoops, thanks!

PR #4835, opened 2023-10-12
Dev container for CellProfiler

callum-jpg:
Here's a Docker container that works with VSCode's Dev Containers, but uses locally cloned repos for core and cellprofiler inside the container. This means you can make changes to CellProfiler and core, but all of the required dependencies are all isolated inside the docker container.

The script `prepare_env.sh` will install the repos so you can access them inside the container. This script also reactivates the micromamba/conda environment that contains all of the dependencies. This script is run when issuing the `docker run` command.

Building and running the docker container assumes the following folder structure:
```
cellprofiler_repos/
├── CellProfiler/
│   ├── docker/
│   │   └── dev/
│   │       ├── Dockerfile
│   │       └── prepare_env.sh
│   ├── setup.py
│   └── ...
└── core/
    ├── setup.py
    └── ...
```

Commands for building/running while inside the `cellprofiler_repos` directory:
```
docker buildx build --platform linux/amd64 -t cellprofiler/cellprofiler_dev:latest CellProfiler/docker/dev

docker run -it --rm -v $(PWD):/workspace --name cellprofiler_dev cellprofiler/cellprofiler_dev:latest sh CellProfiler/docker/dev/prepare_env.sh
```

Once the docker container is running, you can connect to it using VSCode's "Dev Containers" by selecting "Attach to Running Container..." and then selecting "cellprofiler_dev". If you need access to jupyter from VSCode, install the python and jupyter vscode extensions.

`docker images` says the container is 2.79GB on my system. Using cellprofiler in headless mode works too since the container is interactive, but you could easily detach it with `-d` too.


bethac07:
Fun! We were just chatting about CellProfiler and Dockers yesterday, it must have summoned you ;) .

So good news/bad news - @gnodar01 has been finishing up work on monorepo and we are [planning to merge it likely next week](https://forum.image.sc/t/restructuring-of-github-repositories/87263). So your current structure is not going to be the structure next week, so this will presumably need to be tweaked to match. Apologies!

callum-jpg:
I can't emphasise enough how excited I am for the monorepo. Once the monorepo has been merged, I'll tweak this PR

gnodar01:
@callum-jpg We've monorepo-ized. I've rebase this on the new default branch, `main`, but haven't updated anything else. I think this might be better off in the `distribution` directory, either `distribution/docker/dev` or `distribution/dev/docker`.

gnodar01 (commented):
@callum-jpg We've monorepo-ized. I've rebase this on the new default branch, main, but haven't updated anything else. I think this might be better off in the distribution directory, either distribution/docker/dev or distribution/dev/docker.

docker/dev/prepare_env.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+# Install local core and cellprofiler repos
+pip install -e core/
```
bethac07, line 4: I think/hope if you `pip install -e` the monorepo right now you get all the necessary subpackages, maybe? Have we checked that @gnodar01 , or do you need to pip install the core and frontend subfolders rn?

gnodar01: You currently have to do (from the root of the monorepo):

```
pip install -e src/subpackages/library
pip install -e src/subpackages/core
pip install -e src/frontend
```

I haven't decided on the ideal way to do a top level "install all in developer mode" script. I suppose the easiest way would just be to put a `requirements.txt` at the top level, with:

```requirements.txt
-e src/subpackages/library
-e src/subpackages/core
-e src/frontend
```

and then just run `pip install -r requirements.txt` at the top level.

I wouldn't mind doing that in this PR, with the caveat that it may be subject to change in the near-ish term.

callum-jpg:
Ok, so I've updated the container to reflect the new monorepo structure. As a surprise I've also added another container I think is useful. Long story short: it allows you to use the CellProfiler GUI inside the Docker container using the browser. This is possible using Xpra.

Here's how it looks: 
cp_xpra

I think this is cool from a development standpoint, but I think there's potential here too if a user has issues installing CellProfiler on their own system for whatever reason. Could also host remotely. In these scenarios, the `dev_gui/Dockerfile` can be changed to install CellProfiler from PyPI.

While both containers work in a basic sense, here are a few roadblocks I wasn't able to get past that might make the experience a little better:
- Currently the `prepare_env.sh` script is used AFTER docker run to install the local CellProfiler repo with pip in editable mode. Is it possible to have this be automatically triggered on docker run? eg. with `SHELL` or something
- For the GUI docker I tried to add a start menu option and a desktop to make it feel a little more like a normal desktop but this didn't work as I'd hoped. Code I tried can be found [here](https://github.com/CellProfiler/CellProfiler/blob/690a8145e8a1f7b8c30bed21a8e03b405b93a212/distribution/docker/dev_gui/Dockerfile#L109)
- For some reason using wxPython==4.2.1 works in the headless container but does not in the GUI container. In the GUI one, it returns the error `pkg_resources.DistributionNotFound The 'wxPython==4.2.0' distribution was not found and is required by CellProfiler`. 4.2.0 works, but a lot of GUI elements have missing icons and the terminal shows lots and lots of lovely warnings
- The GUI container could inherit from the dev container, as seen [here](https://github.com/CellProfiler/CellProfiler/blob/690a8145e8a1f7b8c30bed21a8e03b405b93a212/distribution/docker/dev_gui/Dockerfile#L72C31-L72C31)

PR #4846, opened 2023-12-07
Template module for running segmentation tasks from Docker [WIP]

bethac07:


bethac07:
Closing in favor of #4893

PR #4847, opened 2023-12-15
Move to Pyproject

gnodar01:
Still uses setuptools, but updates for compliance with PEPs 517, 518, 621, 660. Switches from `setup.py` to `pyproject.toml` for all packages.


src/frontend/pyproject.toml
```diff
@@ -0,0 +1,118 @@
+[build-system]
+# needs a version that supports PEPs 517, 518, 621, 660
+# https://setuptools.pypa.io/en/latest/history.html#v61-0-0
+requires = ["setuptools>=61.0.0"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "CellpPofiler"
+version = "5.0.0.dev1"
+authors = [
```
bethac07, line 10: Several of these emails won't work anymore, do we want to remove? I notice Anne's also doesn't have one, so guessing we don't NEED to have them

src/frontend/pyproject.toml
```diff
@@ -0,0 +1,118 @@
...
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Topic :: Scientific/Engineering :: Bio-Informatics",
+    "Topic :: Scientific/Engineering :: Image Recognition",
+    "Topic :: Scientific/Engineering :: Image Processing",
+    "Topic :: Scientific/Engineering",
+]
+dependencies = [
+    "cellprofiler-library==5.0.0",
```
bethac07, line 56: For this and below, if we really want to keep everything in sync, can we pull the version directly?

bethac07:
Leaving way-too-early-to-worry-about-addressing-yet comments, mostly because I currently have a goldfish brain and don't want to forget entirely. Ignore me until you're ready to not.

src/frontend/pyproject.toml
gnodar01: Not until the build system refactor, I don't think. We can dynamically version the package that a particular `pyproject.toml` refers to (see latest changes, and use of `setuptools-scm`), but we can't dynamically link a dependencies version.

gnodar01:
Taking out of draft status, but still not fully ready for review. Need to run some tests with github actions.

.github/workflows/release.yml
```diff
@@ -4,125 +4,141 @@ jobs:
...
-      - name: Upload Windows Artifact
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        uses: actions/upload-release-asset@v1
-        with:
-          asset_content_type: application/exe
-          asset_name: CellProfiler-Windows-4.2.4.exe
-          asset_path: /home/runner/work/CellProfiler/CellProfiler/CellProfiler-Windows-4.2.4.exe
-          upload_url: ${{ steps.create_release.outputs.upload_url }}
+# TODO re-enable, make some condition for this to run
```
bethac07, line 89: This just makes the release accessible to try to download once GH has built it, it doesn't send it outside of GH. Should be safe to keep this enabled once you're done debugging without a separate condition - after this PR, hopefully will not be too many times where we want to test that it builds an executable, but we don't want to then take that executable and play with it/distribute it. Having auto upload to somewhere helpful like our AWS bucket is a good NEXT step, but that's not this bit.

bethac07:
Sorry, couldn't help myself (but, to be honest, didn't try that hard - my love-hate relationship with getting CellProfiler to build is well documented and probably a(nother) sign that something is deeply wrong with me). I hope "debugging the CellProfiler Windows build and [adding a bunch of Powershell stuff to it](https://github.com/CellProfiler/CellProfiler/pull/4847/files/c77a4e7d632c56a0f1879e16f99054d019e008ea..a37345040bf4dfbcfd682f4ee1f34af8799e4a4a)" wasn't your lifelong hope and dream that I took away.

Windows now builds-ish, at least it completes, though there are a bunch of missing hidden imports it says it can't find that seem very important, and Mac is failing because pyinstaller is now trying to start the GUI when it imports the GUI, which does not go well for it. The current pyprojects don't pin pyinstaller back; I'm reasonably sure we've never successfully built beyond 4.10 in [our own builds](https://github.com/CellProfiler/CellProfiler/blob/4.2.x/.github/workflows/release.yml#L34), and it looks like [neither have our friends](https://github.com/glencoesoftware/CellProfiler/blob/compat-5-0-ngff/.github/workflows/releasev2.yml#L43). It looks like the unpin happened [here](https://github.com/CellProfiler/CellProfiler/commit/31d5299db10afd2047ef57d9c61f33368e1a8ab7#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34L46) on September 22nd, and the release action shows no runs between August and a couple of days ago, so that checks out. 
 
I would be tempted at this point to see if restoring the pyinstaller pin fixes things, and making pyinstaller 6 compatibility an issue for 2024-Nodar-and-or-Beth so this chunk of work can get wrapped up in a nice little Christmas bow, but your call, I'll stop messing around in your sandbox now. :)

gnodar01:
TODO:
[ ] - check windows build works

bethac07 (approved):
The one unresolved comment I have is not a showstopper, merely a "making sure this line looks the way you want, not sure if this comment was intended as documentation or just a leftover." Otherwise, super exciting!!!

distribution/macos/Makefile
```diff
@@ -1,7 +1,17 @@
+# make CP_VERSION=5.0.0.dev32 CP_BRANCH="" CP_DIR=/Users/ngogober/Developer/CellProfiler/CellProfiler
```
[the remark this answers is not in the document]
gnodar01: Indeed, it was intended as documentation. I'll add to it so that it's more instructive.

gnodar01:
@bethac07 The build was indeed broken on Windows, but luckily the fix wasn't crazy and now it runs. There is one thing I'm not sure on re. Windows however: was it always the case that when you launch CellProfiler.exe the console appears with the "Starting CellProfiler" message? Or is that something I introduced. I do know [how to disable it](https://pyinstaller.org/en/v6.3.0/usage.html#cmdoption-w), just not sure if I should.

bethac07:
Yes, that was always the case, and it's definitely preferred - that way users can give us stack traces, can see what's going on during analysis mode, etc. I remember spending ~10 hours or so way back in the day trying to figure out a way to get Mac to also give us a comparable view-only console, and then just giving up (spent about a minute jsut now looking for the GH issue for that that I 65% think we may have had, also gave up). It's like, 15thousandth on the to-do list, especially because in a pinch, we can just run the internal executable from Mac to get such a console, but it's definitely feature, not bug.

PR #4852, opened 2024-01-11
Add MeasureObjectSizeShape to Library

callum-jpg:
Decisions/questions that need to be addressed before merge:
- Input objects are expected to be in the dense format, but is this the best choice?
	- I chose dense since it feels the most intuitive (to me), while still supporting overlapping etc. Converting to dense should be easy with the Library segmentation `..._convet_to_dense` functions.
- I added a dense property to [_objects.py](https://github.com/callum-jpg/CellProfiler/blob/59a786fe53f909ffec55737296abec5b137ec9f7/src/subpackages/core/cellprofiler_core/object/_objects.py#L165) for use in places [like this](https://github.com/callum-jpg/CellProfiler/blob/59a786fe53f909ffec55737296abec5b137ec9f7/src/frontend/cellprofiler/modules/measureobjectsizeshape.py#L351)
- Where should feature names be stored?
	- I like the [Enum stuff](https://github.com/CellProfiler/CellProfiler/blob/main/src/subpackages/library/cellprofiler_library/opts/measureimageoverlap.py) that was implemented in #4758 and wanted to use [something similar here](https://github.com/callum-jpg/CellProfiler/blob/59a786fe53f909ffec55737296abec5b137ec9f7/src/subpackages/library/cellprofiler_library/opts/objectsizeshapefeatures.py)
		- ie. define feature names once and they can be imported into the frontend and in library, where they are needed
	- Keeping them in the frontend module would require library imports from frontend, which is not what we want
	- Better name than `opts`? `fts`? (assuming opts is reserved for 'options' to arguments in CellProfiler GUI and library)
- I think it's worthwhile having more tests that ensure a particular value is consistent over time
	- ie. a test for correct values for all possible features given a single object

I'm going to be using this library implementation for a project, which will hopefully highlight any bugs to be fixed.


src/frontend/cellprofiler/modules/measureobjectsizeshape.py
```diff
@@ -826,51 +370,13 @@ def display(self, workspace, figure):
...
-        objects = workspace.get_objects(object_name)
-        if len(objects.indices) > 0:
-            data = objects.fn_of_ones_label_and_index(function)
-        else:
-            data = numpy.zeros((0,))
-        self.record_measurement(workspace, object_name, feature_name, data)
-
     def record_measurement(self, workspace, object_name, feature_name, result):
         """Record the result of a measurement in the workspace's measurements"""
         data = centrosome.cpmorphology.fixup_scipy_ndimage_result(result)
```
sspathak, line 375: Should this be in library? It's a "mere" fixup, but it's also a morphological operation, and users of library may find it convinient to have via library.

sspathak (changes requested):
Has conflicts with main that need to be resolved

sspathak:
> Where should feature names be stored

We discussed and decided on the enum approach as being the best option.

> Better name than opts? fts? (assuming opts is reserved for 'options' to arguments in CellProfiler GUI and library)

Agreed, we need to have a separate namespace like `fts`. We'll make these changes on a separate PR though.

(This and above comments made by @gnodar01 )

sspathak:
We rebased the contents of this PR onto `main`, so we'll need to close this one and open a new one.

PR #4861, opened 2024-01-24
Fix core tests

### LINKED ISSUE #4859: CI Bug - Core tests run infinitely except for when you look at them hard enough
May or may not be reproducable, depending on wether this bug appearing depends on randomness of Github actions (which vm GH happens to allocate, where it is run, when it happens to run, etc.)

In the most recent ~four~ six (and counting) runs of the `test.yml` workflow, tests for frontend and library pass as usual, but tests for core run without outputting anything for an indefinite amount of time. [Here](https://github.com/CellProfiler/CellProfiler/actions/runs/7629460281) there were 3 attempts. [The first](https://github.com/CellProfiler/CellProfiler/actions/runs/7629460281/attempts/1) was manually cancelled after 30 minutes. [The second](https://github.com/CellProfiler/CellProfiler/actions/runs/7629460281/attempts/2) was manually cancelled after 1 hour and 40 minutes. On [attempt 3](https://github.com/CellProfiler/CellProfiler/actions/runs/7629460281) I re-ran failed (cancelled) jobs with "enable debug logging" toggled on:

Screenshot 2024-01-23 at 7 10 44 PM

The debug information showed nothing of particular note, but the tests ran successfully, and took just over 3 minutes to do so:

Screenshot 2024-01-23 at 7 13 39 PM

I assumed it was just coincidence, and merged PR #4847 where those tests were being launched from. After merge, tests run again since master is being pushed to.

[Again](https://github.com/CellProfiler/CellProfiler/actions/runs/7631010120/attempts/1), debug info was off, and core tests ran indefinitely; this time for 4 hours 15 minutes, at which point I cancelled.

Each of the times it ran indefinitely, the output was sparse:

Screenshot 2024-01-23 at 7 16 38 PM

The output just shows the command for invoking the tests, the initial pytest output right before the first test begins, and then nothing. It doesn't seem to hang on a particular test, but rather hangs before the first even begins.

Running tests locally, nothing seems to be amiss. They run perfectly fine.

~As I'm typing this,~ I've re-run the most recent cancelled tests with "enable debug logging" toggled on, and [again](https://github.com/CellProfiler/CellProfiler/actions/runs/7631010120) it ran to completion in 3 minutes.

Should note, looking through the outputs of previous steps, nothing about the installation process seems amiss (the usual culprit of long-running tasks being pip install something). It gets **to** the test step quick enough, just hangs **on** the test step.

gnodar01:
As yet unsolved, but attempts to fix #4859

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -125,7 +125,7 @@ jobs:
           pip install -e src/subpackages/core[test]
       - name: Run Tests
         run: |
-          pytest -k 'not test_load_objects and not test_load_single_object' ${{ github.workspace }}/tests/core
+          pytest --tb=long --show-capture=all -vvv --color=yes --code-highlight=yes --timeout=300 -k 'not test_load_objects and not test_load_single_object' ${{ github.workspace }}/tests/core
 
   test-library:
     name: Install and test Library
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -126,7 +126,7 @@ jobs:
       - name: Run Tests
         timeout-minutes: 10
         run: |
-          pytest --tb=long --show-capture=all -vvv --color=yes --code-highlight=yes --timeout=300 -k 'not test_load_objects and not test_load_single_object' ${{ github.workspace }}/tests/core
+          pytest --tb=long --show-capture=all -vvv --color=yes --code-highlight=yes --timeout=300 -k 'not test_load_objects and not test_load_single_object and test_04_05_exception' ${{ github.workspace }}/tests/core/
 
   test-library:
     name: Install and test Library
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -126,7 +126,7 @@ jobs:
       - name: Run Tests
         timeout-minutes: 10
         run: |
-          pytest --tb=long --show-capture=all -vvv --color=yes --code-highlight=yes --timeout=300 -k 'not test_load_objects and not test_load_single_object and test_04_05_exception' ${{ github.workspace }}/tests/core/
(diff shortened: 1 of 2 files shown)


gnodar01:
So I put in a global timer, did not toggle on "enable debug logging", and expected it to hit the timeout and cause it to prematurely fail. Of course it ran to completion perfectly.

gnodar01:
We don't know if this fixes it, but merging anyway.

bethac07 (approved):

gnodar01:
On [Attempt 1 of this run](https://github.com/CellProfiler/CellProfiler/actions/runs/7821952627/job/21339989664) the increased verbosity invocation of `pytest` showed that `tests/core/analysis/test_analysis.py::TestAnalysis::test_04_05_exception` fails. It "failed" in the sense that we set a timeout that causes it to fail. Without the timeout, it is more likely that it just runs forever. On attempt 2, everything passed with no hangs.

The timeout also allowed us to get to the rest of tests.

`test_analysis` has a bunch of `zmq` and multi threading stuff.

What we don't know yet:
- when it does hang indefinitely, is it always `test_04_05_exception` that hangs, or is it just *some* test in `test_analysis.py`
  - because of the workers being spawned, it may be some *interaction* between tests that causes the hanging, so running just `test_04_05_exception` may not cause the hang, but running all of them causes at least one of the to hang... sometimes
- why does it not hang sometimes
  - it didn't happen to run when we ran only that test
  - even when all tests are enabled, sometimes it passes without hangs
  - it *seems* that logging out information increases the chance it doesn't hang which may imply that its a timing issue


We will merge this PR, but not resolve the issue. We'll use the increase verbosity to monitor any future runs, and gather information such as "is it always the same test that hangs" and "is there any common patterns between successful/unsuccessful runs.

gnodar01:
Update `tests/core/analysis/test_analysis.py::TestAnalysis::test_04_04_01_display` failed on [this run](https://github.com/CellProfiler/CellProfiler/actions/runs/7822612651/job/21342007966)

PR #4865, opened 2024-02-06
Bump dockefile for python 3.9

### LINKED ISSUE #4864: Python 3.9 not available in Ubuntu 18.04
Hi all,

Just a heads up, as far as I can tell, Python 3.9 isn't available in Ubuntu 18.04 LTS, which means https://github.com/CellProfiler/CellProfiler/commit/ac4361ae5131baf1b4ad50b0b7d4edc37b3856ca fails during the build. I don't see any CI/CD tests associated with the deployment and the latest version on DockerHub is still using 3.8, so I wanted to report it here. 

Changing the first line to `FROM amd64/ubuntu:20.04` seems to fix it (has to be amd64 arch; there isn't a wxPython wheel for arm64 and building the wheel from source fails), along with a few other changes:

- `ARG DEBIAN_FRONTEND=noninteractive` to get past some of the new installation logic 
- wxPython path to `RUN pip install https://extras.wxpython.org/wxPython4/extras/linux/gtk3/ubuntu-20.04/wxPython-4.2.1-cp39-cp39-linux_x86_64.whl` 

18.04 has reached its EOL anyway, so probably best to update. I can PR if you want.

gnodar01:
ubuntu 18 does not come with python 3.9 and has reached EOL, update dockerfile to use ubuntu 20

credit to @wheelern

resolves #4864

### DIFF
--- a/distribution/docker/Dockerfile
+++ b/distribution/docker/Dockerfile
@@ -1,4 +1,8 @@
-FROM ubuntu:18.04
+# must be am64, wxPython wheel for arm64 is not available
+FROM amd64/ubuntu:20.04
+
+# bypass apt installation frontend interaction
+ARG DEBIAN_FRONTEND=noninteractive
 
 ARG cp_version=4.2.3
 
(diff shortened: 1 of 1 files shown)


bethac07:
Thanks Nic and Nodar!

bethac07 (approved):

PR #4867, opened 2024-02-07
Build patch

gnodar01:
[Latest](https://github.com/CellProfiler/CellProfiler/actions/runs/7815198162) build fails because `mysqlclient` fails to build (on MacOS). 

problem:
- there are no wheels `mysqlclient` for mac, instead `pip` builds the wheel
- this has generally worked fine, but recently, for `mysql >= 8.3.0`, the MySQL team killed
  some APIs that have been deprecated for a while now, causing the `mysqlclient` build to fail

solution:
- we can pin the `brew install` of `mysql` to `< 8.3.0`, or we can bump the
 ` pip install` of `mysqlclient` to `>= 2.2.3` which as of 2/4/24 has a [patch](https://github.com/PyMySQL/mysqlclient/issues/688)
- later seems more "responsible"


bethac07 (approved):
Sounds fair to me, assuming tests pass!

gnodar01:
> Sounds fair to me, assuming tests pass!

They did not, but do now other than the #4859 issue.

PR #4893, opened 2024-04-26
Add segmentation-plugin-that-requires-external-dependencies template

bethac07:
These will likely usually, though need not be, DL segmentation things we want to run in inference. I don't see that happening _less_ often anytime soon!


gnodar01:
Tests failed because python installed for x64 architecture, but the default macos runner is now arm. Fix will be in main branch once #4894 is merged.

bethac07:
@gnodar01 ok to pull this?

gnodar01 (approved):

PR #4897, opened 2024-04-30
Fix project descriptions

bethac07:
There was a typo in the toml file for library, and the relpaths/non-https paths in the READMEs were causing strange links on PyPI. Nothing life-critical just tidying up.


bethac07:
(I'm open to reverting the relpath changes if we don't want them, but I'd argue it "maybe switching your branch if you're wandering around the website in a different branch" is a less-worse than "links looking weird and broken-y on PyPI". But the typo and the https additions on actual websites are def needed!)

gnodar01 (approved):

PR #4909, opened 2024-07-23
Enable 3D display in MeasureImageOverlap

### LINKED ISSUE #4890: MeasureImageOverlap's viewer is not set for 3D
The module runs, but you just get a view like this. Current 4.2.6 build.

image

bethac07:
Resolves #4890 

TLDR, we reshape the image to make the measurement easier, and then we never put it back.


gnodar01:
I just merged the new centrosome version. Please rebase on `main` so that tests pass and merging is no longer blocked.

PR #4920, opened 2024-07-25
Fix ClassifyObjects two-measurement display in analysis mode

### LINKED ISSUE #4817: ClassifyObjects: TypeError: Invalid shape (1, 1024, 1024) for image data
**Describe the bug**
When using ClassifyObjects with a pair of measurements it runs fine in test mode but throws an error when running analyze

**To Reproduce**
Run the attached pipeline in test and analyze mode

**Expected behavior**
I expect analyze and testmode to behave the same and not throw errors

**Screenshots**
Analyze
image

Test
image

 - OS: Windows 11
 - CellProfiler Version: 2.4.6
 - Downloaded from the website or installed from source? Downloaded from website
 - If the latter, how (cloned, installed from pypi, etc)? -

**Additional context**
Stacktrace:
  File "cellprofiler\gui\pipelinecontroller.py", line 2992, in module_display_request
  File "cellprofiler\modules\classifyobjects.py", line 978, in display
  File "cellprofiler\modules\classifyobjects.py", line 1112, in display_two_measurements
  File "cellprofiler\gui\utilities\figure.py", line 90, in wrapper
  File "cellprofiler\gui\figure\_figure.py", line 1886, in subplot_imshow_labels
  File "cellprofiler\gui\utilities\figure.py", line 90, in wrapper
  File "cellprofiler\gui\figure\_figure.py", line 1643, in subplot_imshow
  File "matplotlib\__init__.py", line 1599, in inner
  File "matplotlib\cbook\deprecation.py", line 369, in wrapper
  File "matplotlib\cbook\deprecation.py", line 369, in wrapper
  File "matplotlib\axes\_axes.py", line 5679, in imshow
  File "matplotlib\image.py", line 689, in set_data
TypeError: Invalid shape (1, 1024, 1024) for image data

Included are some images and the pipeline
[cellprofiler pipeline.zip](https://github.com/CellProfiler/CellProfiler/files/12535882/cellprofiler.pipeline.zip)

bethac07:
Resolves #4817 

I don't know why the labels were tuple wrapped here - as far as I can tell, nowhere else in the codebase are they treated that way (there are only 6 modules that use the `workspace.display_data.labels` syntax anywhere, it seems to be only in older modules with weird display needs). I suspect this was just an older syntax, that since we're deep in a corner here (only in ClassifyObjects, only in "classifying in two objects" mode, only when running in analysis and not test mode, where folks shouldn't really have eyes open at all _anyway_), it was just never fixed.

I don't fully understand why the array was being differently shaped in analysis vs test mode, it seems like analysis enforces some assumptions about the shape/type that test does not... I am hoping we will never need to/benefit from digging into this further but will post breadcrumbs below. 

<details>

Before patch, printing `workspace.display_data.labels` in test mode:

```
(array([[0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       ...,
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0]], dtype=int8),)
```

in analysis mode:

```
[array([[0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       ...,
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0],
       [0, 0, 0, ..., 0, 0, 0]], dtype=int8)]
```

After:

test:

```
[[0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]
 ...
 [0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]]
```

analysis:

```
[[0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]
 ...
 [0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]
 [0 0 0 ... 0 0 0]]
```

</details>


gnodar01:
I could not resist. Obviously... But perhaps to the good, because the underlying problem here may surface again in the future, whenever we place tuple data into `workspace` and expect (require) it to stay a tuple rather than being converted to a list.

In the `display_two_measurements` method you find the following line:

```python
labels = object_codes[mapping[workspace.display_data.labels]]
```

As you showed, in test mode the `labels` numpy ndarray is wrapped in a tuple, `(...)`, and in analysis mode it's wrapped in a native python list `[...]`.

These two different formats cause different behaviors in numpy's fancy indexing system, as simplified here:

``` python
import numpy as np

idxs = np.arange(100).reshape((10,10))

data = np.random.random(100)

data[[idxs]].shape
# (1, 10, 10)

data[(idxs,)].shape
# (10, 10)
```

This is why the current fix in this PR, of removing the tuple around `objects.segmented`, works. In the above example `data[idxs]` works the same as `data[(idxs,)]` (but not the same as `data[[idxs]]`). In the tuple case, NumPy sees `(idxs,)` as a single-element tuple containing the array indices. This is still advanced indexing, but now it is interpreted as indexing with a single 2D array of shape `(10, 10)`. This is identical to the no tuple AND no list case. In the list case however, NumPy sees `[idxs]`, flattens it to a shape of `(100,)`, and then treats it as an advanced index, but adds an extra dimension because of the double brackets, `data[[idxs]]`.

Why does Analysis mode get square brackets whereas Test mode gets parenthesis? Because in the code, under the `run_two_measurements` method, we have `workspace.display_data.labels = (objects.segmented,)`. So it's originally defined to be in a tuple. Test mode runs this directly. You have the following chain of calls:

For running the module,

`PipelineController.do_step -> Pipeline.run_module(module, workspace_model) -> run_two_measurements -> workspace.display_data.labels = (objects.segmented,)`

and later for displaying the contents,

`PipelineController.do_step -> module.display(workspace_model, figure) -> display_two_measurements -> labels = object_codes[mapping[workspace.display_data.labels]]`

However in Analysis mode, a worker runs the pipeline for its given image set. The run path looks similar:

`Worker.do_job -> current_pipeline.run_image_set -> Pipeline.run_module(module, workspace_model) -> runt_two_measurements -> workspace.display_data.labels = (objects.segmented,)`

However, we have quite a different path to get to display.

After `Pipeline.run_module` returns, the next step in `current_pipeline.run_image_set` is calling `Worker.display_handler(module, workspace.display_data, image_set_number)`.  That creates a `Display` object, which inherits from `AnalysisRequest` (itself inheriting from `Communicable`), which is a class for sending requests from the worker to the main process. Next `Worker.send(...,display_data_dict=display_data.__dict__,...)` is called (`labels` is in this dict, but still wrapped in a tuple at this point). This serializes the data and sends it over `work_socket` to the `request_queue` of the `Runner.jobserver` daemon of the main thread. The `jobserver` dequeues it and deserializes it.

At this point the damage is done, but just for completion, the remaining steps are as follows. The `jobserver` posts the event to the listener, which in this case is `PipelineController.analysis_event_handler`. That calls `wx.CallAfter(self.module_display_request, evt)` (`evt` contains the `display_data.__dict__`, which contains the `labels` and also contains `module_num`). This runs `PipelineController.module_display_request` which gets the module via the `module_num` from the `Pipeline` object, and calls `module.display(self.__workspace, fig)`, which as previously calls `display_two_measurements -> labels = object_codes[mapping[workspace.display_data.labels]]`, except now with `labels` wrapped in square brackets.

The serialization process is where the issue occurs. It happens through `Communicable.send`, which takes the contents of the `display_data_dict`, and puts them through `cellprofiler_core.utilities.zmq.json_encode` (still our code, not pyzmq's). This calls `make_sendable_dictionary` recursively calls itself on the `display_data_dict`, and then calls the following when it encounters `labels`:

```python
        elif isinstance(v, (list, tuple)):
            v = make_sendable_sequence(v)
```

`make_sendable_sequence` in turn has:

```python
    result = []
    for v in l:
        if isinstance(v, (list, tuple)):
            result.append(make_sendable_sequence(v))
    ...
    return tuple(result)
```

This is problematic, but not the main problem. You have the `tuple` (`(...)`) being replaced with the `list` (`[...]`). Both list contents, and tuple contents, are being appended to `result`, which is defined to be a list. But then on returning it its cast into a tuple (always). The real problem follows.

Once `make_sendable_dictionary` finishes its recursive serialization it returns to `json_encode`. From there we have the following:

```python
    buffers = []
    encoder = make_CP_encoder(buffers)
    json_str = json.dumps(sendable_dict, default=encoder)
```

where `make_cp_encoder` does this:

```python
def make_CP_encoder(buffers):
    def encoder(data, buffers=buffers):
    ...
    return encoder
```

This is used as a custom encoder for `json.dumps`. The relevant content of the custom `encoder` function is this:

```python
        ...
        if isinstance(data, memoryview):
            # arbitrary data
            idx = len(buffers)
            buffers.append(data)
            return {"__buffer__": True, "idx": idx}
```

Misc (non-numpy) data like tuples and lists just get appended to `buffers`, which is a list.

The solution is to modify the custom json encoder to account for tuples.

I mean the real solution is don't abuse json in this way, but you know...

gnodar01:
I tried out the results of the commit above with and without the tuples, as set in: `workspace.display_data.labels = (objects.segmented,)`. Both now work, i.e. it is not converted to a list in Analysis mode. I think it's worth keeping in your changes @bethac07, but we can send tuple data and have it stay tuple data now, so that's nice.

bethac07:
X-D

bethac07:
Nice detective work tho :-D

gnodar01 (approved):

PR #4923, opened 2024-07-25
Catch case where in local thresholding, image is mixture of nan and 0

bethac07:
@gnodar01 , this is the least-messy way I could think of to do that (because stuff like `unique` has weird behavior with nan's present, but `nan`'s also evaluate to true, so...), but if you can think of something more elegant and performant (likely), let me know and I'm happy to switch the implementation. Will hold off making a PR to main until we decide how to efficiently implement.

### DIFF
--- a/cellprofiler/modules/threshold.py
+++ b/cellprofiler/modules/threshold.py
@@ -932,6 +932,10 @@ staining.
         elif numpy.all(image_data == image_data[0]):
             local_threshold = numpy.full_like(image_data, image_data[0])
 
+        elif numpy.unique(numpy.where(numpy.isnan(image_data),0,image_data)).shape == (1,):
+            #test if all values are a mixture of 0 or nan
+            local_threshold = numpy.zeros_like(image_data)
+
         elif self.threshold_operation == TM_LI:
             local_threshold = self._run_local_threshold(
                 image_data,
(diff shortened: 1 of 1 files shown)


cellprofiler/modules/threshold.py
```diff
@@ -932,6 +932,10 @@ def get_local_threshold(self, image, mask, volumetric):
         elif numpy.all(image_data == image_data[0]):
             local_threshold = numpy.full_like(image_data, image_data[0])
 
+        elif numpy.unique(numpy.where(numpy.isnan(image_data),0,image_data)).shape == (1,):
```
gnodar01, line 935: `np.all((x == 0) | np.isnan(x))` avoids writing a new `image_data`-sized array/matrix.

gnodar01 (changes requested):

cellprofiler/modules/threshold.py
gnodar01: Or at least it writes 2 x `image_data` shaped arrays, but the dtype is `bool`

bethac07: See, I knew you'd have a cleaner idea

gnodar01 (approved):

bethac07:
I did not break any local tests, but nor did I write any new ones here for this case; I left the companion 5 PR in draft until I do, because this seems like a case worth catching

gnodar01 (approved):

PR #4924, opened 2024-07-25
Add  local thresholding catch for all nans or a mixture of 0s and nans

### LINKED ISSUE #4862: IdentifyPrmaryObjects: Error when using adaptive thresholding on completely masked image
**Describe the bug**
Running the attached pipeline (a variation of the welcome screen example pipeline) in which the entire image is masked out yields the error message below:
> Failed to run module IdentifyPrimaryObjects
> Traceback (most recent call last):
>   File "cellprofiler\gui\pipelinecontroller.py", line 3390, in do_step
>   File "cellprofiler_core\pipeline\_pipeline.py", line 1341, in run_module
>   File "cellprofiler\modules\identifyprimaryobjects.py", line 1076, in run
>   File "cellprofiler\modules\identifyprimaryobjects.py", line 1213, in _threshold_image
>   File "cellprofiler\modules\threshold.py", line 879, in get_threshold
>   File "cellprofiler\modules\threshold.py", line 940, in get_local_threshold
>   File "<__array_function__ internals>", line 180, in amin
>   File "numpy\core\fromnumeric.py", line 2918, in amin
>   File "numpy\core\fromnumeric.py", line 86, in _wrapreduction
> ValueError: zero-size array to reduction operation minimum which has no identity`

Running version 4.2.4 on windows.
[pipeline.zip](https://github.com/CellProfiler/CellProfiler/files/14126075/pipeline.zip)

bethac07:
Resolves #4862 

Leaving as draft for now because we probably want a test here

### DIFF
--- a/src/subpackages/library/cellprofiler_library/functions/image_processing.py
+++ b/src/subpackages/library/cellprofiler_library/functions/image_processing.py
@@ -246,10 +246,9 @@ def get_adaptive_threshold(
 
     thresh_out = None
 
-    if len(image) == 0 or numpy.all(image == numpy.nan):
-        thresh_out = numpy.zeros_like(image)
-    elif numpy.all((image == 0) | numpy.isnan(image)):
-        #test if all values are a mixture of 0 or nan
+    # test if all values are a mixture of 0 or nan
+    # also True if image.size == 0
+    if  numpy.all((image == 0) | numpy.isnan(image)):
         thresh_out = numpy.zeros_like(image)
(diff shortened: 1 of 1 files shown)


gnodar01:
I actually can't trigger the same issue (ie the error message) being reported in #4862 (even after removing modifications made here). May have been "fixed" in the move to library.

bethac07:
Ok! Fine to close if we don't need it, but I also think reasonably harmless to keep in (but you tell me, as you're thinking about large image, if you think the performance hit isn't worth it).

PR #4926, opened 2024-07-31
Pixi

gnodar01:
This PR adds a basic [pixi](https://pixi.sh/) setup.

One thing that would be useful to add, that will not be added in this PR, are `java` and `mysql` installations via `conda-forge`. However for now, assuming you've done the classic `mysql` and `java` setups, `pixi install -e dev` will do an editable install of all 3 cellprofiler packages, equally well across platforms. `pixi run -e dev cp` will run cellprofiler. Alternatively `pixi shell -e dev` will open up a shell with the environment activated (analogous to `conda activate cp-dev`).

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -6,13 +6,15 @@ to enable biologists without training in computer vision \
 or programming to quantitatively measure phenotypes from \
 thousands of images automatically."""
 channels = ["conda-forge"]
-platforms = ["osx-64"]
+platforms = ["win-64", "linux-64", "osx-64", "osx-arm64"]
 
 [tasks]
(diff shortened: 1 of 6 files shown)


pixi.toml
```diff
@@ -0,0 +1,34 @@
+[project]
+name = "CellProfiler"
+description = """
+CellProfiler is a free open-source software designed \
+to enable biologists without training in computer vision \
+or programming to quantitatively measure phenotypes from \
+thousands of images automatically."""
+channels = ["conda-forge"]
+platforms = ["win-64", "linux-64", "osx-64", "osx-arm64"]
```
bethac07, line 9: Did we/should we check on all these places? LMK if you need an arm tester

bethac07: NVM!

bethac07 (approved):

PR #4927, opened 2024-08-01
Core Tests

gnodar01:
Addresses #4859 (maybe). Logging statements were added throughout in order to track the source of the error. Most logging statements, within individual tests, did nothing to alter the sporadic failures. However, after putting logging statements within the `__exit__` method of `FakeWorker`, several dozen runs of testing did not run into the error.

This may have been coincidence. For instance, It could be the case, that although the macos version is specified, different VMs provided by GitHub cause the error to occur, or not occur. Given that the multiple successful runs were conducted over a period of three days (mostly during the weekend), there may have been a string of lucky VM allocations.

However, if the simple act of logging within `__exit__` did indeed fix the problem, it points to what has always been the prime suspect: a race condition. Specifically, each test sets up a context in which `FakeWorker` is instantiated. Once the test cases pass, the work is cancelled, the `uinittest.TestCase`'s `teardown` is called, the sockets are terminated, queues are closed, and the context is exited, causing `__exit__` to be invoked. There is a strong suspicion that during this tear down process, given that multiple things are happening concurrently, there is an order of events (only occurring sometimes) such that the `__exits__`'s invocation of `self.join()` hangs indefinitely, because the `FakeWorker`'s thread has already closed or is otherwise blocked. 

If that is the case, the act of logging, which requires calls to `logging` and system writes, slows things down just enough to avoid the race condition. This conclusion is wild speculation, but I have nothing else to go off of at the moment unless and until tests start failing again, and I can actually read the output of the logging statements.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,7 +53,8 @@ jobs:
         env:
           LC_ALL: "en_US.UTF-8"
         run: |
-          brew install mysql
+          brew install mysql@8.4
+          brew link mysql@8.4 --force
           mysql.server start
           mysql --host=$CP_MYSQL_TEST_HOST --user=$CP_MYSQL_TEST_USER --execute="CREATE DATABASE $CP_MYSQL_TEST_DB;" --skip-password
       - name: Install cellprofiler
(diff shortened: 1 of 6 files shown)


gnodar01:
Is draft, but not put into draft because I want tests to launch. DO NOT MERGE.

gnodar01:
This PR also solved an issue with frontend tests, because the latest version of homebrew's mysql is version 9, which breaks things. We now explicitly install version `8.4`.

PR #4929, opened 2024-08-05
fix label assignment when different objects are neighbors

bethac07:


bethac07:
@gnodar01 if we do decide to do a 4.2.8 next week (or a future week), please do merge, might as well get that bug patch in there

PR #4934, opened 2024-08-13
Fix worker halting

gnodar01:
The problem:

From a debug log of running CP on a pipeline which encounters the issue of a hanging post-processing step, we have this as our final Progress before the hang:

```
[CP - DEBUG] cellprofiler_core.analysis._runner::interface: 👺 sending Progress - [(89, 'FinishedWaitingMeasurements'), (90, 'FinishedWaitingMeasurements'), (92, 'FinishedWaitingMeasurements'), (93, 'FinishedWaitingMeasurements'), (95, 'FinishedWaitingMeasurements'), (96, 'FinishedWaitingMeasurements'), (97, 'FinishedWaitingMeasurements'), (99, 'FinishedWaitingMeasurements'), (101, 'FinishedWaitingMeasurements')]

[CP - INFO] cellprofiler.gui.pipelinecontroller::analysis_event_handler: Progress, Counter({'Done': 231, 'FinishedWaitingMeasurements': 9})
```

I ran the pipeline on `240` image sets, so it wasn't the last few that hung, it was a bunch in the middle, almost sequentially.

Also from the logs, a successfully completed (progress `Done`) image set looks like this:

```
[Worker 2 (76933)] cellprofiler_core.worker._worker::do_job: 👺 job is: [240]
[Worker 2 (76933)] cellprofiler_core.worker._worker::do_job: 👺 worker sending MeasurementsReport for job [240]

[CP - DEBUG] cellprofiler_core.analysis._runner::jobserver: Received measurements report for [240]

[CP - DEBUG] cellprofiler_core.analysis._runner::queue_received_measurements: 👺 QUEUE OP: jobserver starting to put measurements in queue for interface, img_set_nums [240]; is not full
[CP - DEBUG] cellprofiler_core.analysis._runner::queue_received_measurements: 👺 QUEUE OP: jobserver done putting measurements in queue for interface, img_set_nums [240]

[CP - DEBUG] cellprofiler_core.analysis._runner::jobserver: Acknowledged measurements report for [240]

[Worker 2 (76933)] cellprofiler_core.worker._worker::do_job: 👺 worker received MeasurementsReport ACK for job [240]

[Worker 2 (76933)] cellprofiler_core.worker._worker::do_job: 👺 worker finished with job [240]

[CP - DEBUG] cellprofiler_core.analysis._runner::interface: 👺 deleting measurements for [240] as it has be consolidated
```

Whereas an unsuccessful one (stuck in `FinishedWaitingMeasurements` status), looks like this:

```
[Worker 5 (77016)] cellprofiler_core.worker._worker::do_job: 👺 job is: [101]
[Worker 5 (77016)] cellprofiler_core.worker._worker::do_job: 👺 worker sending MeasurementsReport for job [101]

[Worker 5 (77016)] cellprofiler_core.worker._worker::do_job: 👺 worker finished with job [101]
```

Once stuck, the worker blocks until analysis is terminated. So because we have 9 stuck above, those workers are not used for the remainder of jobs in the pipeline. That's pretty bad.

From the logs, the Worker does its job correctly. It gets a job, runs the pipeline, and sends back a `req` of type `MeasurementsReport`. It is the task of `Boundary.spin` to receive this over a socket, and send it to the upward queue for `interface` and `jobserver` to process. For the case of `MeasurementsReport`, it is the job of `interface` to consolidate it into the aggregate measurements, and then send back and `ACK` to the `Worker` so that it can stop blocking and go about its business.

The `interface` and `jobserver` never receive these hung jobs in their queues. In the log it appears `Boundary.spin` never even receives it.

### DIFF
--- a/src/subpackages/core/cellprofiler_core/worker/_worker.py
+++ b/src/subpackages/core/cellprofiler_core/worker/_worker.py
@@ -402,6 +402,16 @@ class Worker:
         req.send_only(work_socket)
         response = None
         while response is None:
+            # we timeout here because we have noticed in cases of piplines with
+            # many image sets (e.g. 240) and many workers (e.g. 15),
+            # that for some small number of image sets (e.g 5), Workers will
+            # successfully send MeasurementReports, but the Boundary.spin
+            # thread never receives them. It is difficult to diagnose,
+            # and difficult to fix "properly". Instead we notice we haven't
+            # received an ACK for a while, and assume we're hung, and the
+            # calling function of send is welcome to retry.
+            # See https://github.com/CellProfiler/CellProfiler/pull/4934
+            # for more details.
             poll_res = poller.poll(2000)
             if len(poll_res) == 0:
                 LOGGER.debug("Worker timeout on polling for response")
--- a/src/subpackages/core/cellprofiler_core/worker/_worker.py
+++ b/src/subpackages/core/cellprofiler_core/worker/_worker.py
@@ -280,18 +280,21 @@ class Worker:
                     return
 
                 if worker_runs_post_group:
-                    last_workspace.interaction_handler = self.interaction_handler
-                    last_workspace.cancel_handler = self.cancel_handler
-                    last_workspace.post_group_display_handler = (
-                        self.post_group_display_handler
-                    )
-                    # There might be an exception in this call, but it will be
-                    # handled elsewhere, and there's nothing we can do for it
-                    # here.
-                    current_pipeline.post_group(
-                        last_workspace, current_measurements.get_grouping_keys()
-                    )
-                    del last_workspace
+                    if not last_workspace is None:
+                        last_workspace.interaction_handler = self.interaction_handler
+                        last_workspace.cancel_handler = self.cancel_handler
+                        last_workspace.post_group_display_handler = (
+                            self.post_group_display_handler
+                        )
+                        # There might be an exception in this call, but it will be
+                        # handled elsewhere, and there's nothing we can do for it
+                        # here.
+                        current_pipeline.post_group(
+                            last_workspace, current_measurements.get_grouping_keys()
+                        )
+                        del last_workspace
+                    else:
+                        LOGGER.error("No workspace from last image set, cannot run post group")
 
             # send measurements back to server
             req = MeasurementsReport(
--- a/src/subpackages/core/cellprofiler_core/worker/_worker.py
+++ b/src/subpackages/core/cellprofiler_core/worker/_worker.py
@@ -25,6 +25,7 @@ from ..constants.worker import NOTIFY_STOP
 from ..constants.worker import all_measurements
 from ..measurement import Measurements
 from ..utilities.measurement import load_measurements_from_buffer
+from ..utilities.zmq import PollTimeoutException
 from ..pipeline import CancelledException
 from ..preferences import get_awt_headless
 from ..preferences import set_preferences_from_dict
@@ -39,7 +40,7 @@ class Worker:
 
     """
 
-    def __init__(self, context, analysis_id, work_request_address, keepalive_address, with_stop_run_loop=True):
+    def __init__(self, context: zmq.Context, analysis_id, work_request_address, keepalive_address, with_stop_run_loop=True):
 
         self.context = context
         self.work_request_address = work_request_address
@@ -303,7 +304,17 @@ class Worker:
                 buf=current_measurements.file_contents(),
                 image_set_numbers=image_set_numbers,
             )
-            rep = self.send(req)
+            while True:
+                try:
+                    rep = self.send(req)
+                    break
+                except PollTimeoutException:
+                    LOGGER.debug(f"Worker timeout on sending MeasurementsReport; reconstructing socket and retry for job {str(job.image_set_numbers)}")
+                    self.work_socket.close(linger=0)
+                    self.work_socket = self.context.socket(zmq.REQ)
+                    self.work_socket.set_hwm(2000)
+                    self.work_socket.connect(self.work_request_address)
+                    continue
 
         except CancelledException:
             # Main thread received shutdown signal
@@ -391,7 +402,11 @@ class Worker:
         req.send_only(work_socket)
         response = None
         while response is None:
-            for socket, state in poller.poll():
+            poll_res = poller.poll(2000)
+            if len(poll_res) == 0:
+                LOGGER.debug("Worker timeout on polling for response")
+                raise PollTimeoutException
+            for socket, state in poll_res:
                 if state != zmq.POLLIN:
                     continue
                 elif socket == self.keepalive_socket:
--- a/src/subpackages/core/cellprofiler_core/worker/_worker.py
+++ b/src/subpackages/core/cellprofiler_core/worker/_worker.py
@@ -61,6 +61,7 @@ class Worker:
 
         # Setup the work server socket
         self.work_socket = self.context.socket(zmq.REQ)
+        self.work_socket.set_hwm(2000)
         self.work_socket.connect(self.work_request_address)
 
         # Establish a connection to the keepalive socket
--- a/src/subpackages/core/cellprofiler_core/worker/_worker.py
+++ b/src/subpackages/core/cellprofiler_core/worker/_worker.py
@@ -306,7 +306,7 @@ class Worker:
             )
             while True:
                 try:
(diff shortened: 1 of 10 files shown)


gnodar01:
The problem:

From a debug log of running CP on a pipeline which encounters the issue of a hanging post-processing step, we have this as our final Progress before the hang:

```
[CP - DEBUG] cellprofiler_core.analysis._runner::interface: 👺 sending Progress - [(89, 'FinishedWaitingMeasurements'), (90, 'FinishedWaitingMeasurements'), (92, 'FinishedWaitingMeasurements'), (93, 'FinishedWaitingMeasurements'), (95, 'FinishedWaitingMeasurements'), (96, 'FinishedWaitingMeasurements'), (97, 'FinishedWaitingMeasurements'), (99, 'FinishedWaitingMeasurements'), (101, 'FinishedWaitingMeasurements')]

[CP - INFO] cellprofiler.gui.pipelinecontroller::analysis_event_handler: Progress, Counter({'Done': 231, 'FinishedWaitingMeasurements': 9})
```

I ran the pipeline on `240` image sets, so it wasn't the last few that hung, it was a bunch in the middle, almost sequentially.

Also from the logs, a successfully completed (progress `Done`) image set looks like this:

```
[Worker 2 (76933)] cellprofiler_core.worker._worker::do_job: 👺 job is: [240]
[Worker 2 (76933)] cellprofiler_core.worker._worker::do_job: 👺 worker sending MeasurementsReport for job [240]

[CP - DEBUG] cellprofiler_core.analysis._runner::jobserver: Received measurements report for [240]

[CP - DEBUG] cellprofiler_core.analysis._runner::queue_received_measurements: 👺 QUEUE OP: jobserver starting to put measurements in queue for interface, img_set_nums [240]; is not full
[CP - DEBUG] cellprofiler_core.analysis._runner::queue_received_measurements: 👺 QUEUE OP: jobserver done putting measurements in queue for interface, img_set_nums [240]

[CP - DEBUG] cellprofiler_core.analysis._runner::jobserver: Acknowledged measurements report for [240]

[Worker 2 (76933)] cellprofiler_core.worker._worker::do_job: 👺 worker received MeasurementsReport ACK for job [240]

[Worker 2 (76933)] cellprofiler_core.worker._worker::do_job: 👺 worker finished with job [240]

[CP - DEBUG] cellprofiler_core.analysis._runner::interface: 👺 deleting measurements for [240] as it has be consolidated
```

Whereas an unsuccessful one (stuck in `FinishedWaitingMeasurements` status), looks like this:

```
[Worker 5 (77016)] cellprofiler_core.worker._worker::do_job: 👺 job is: [101]
[Worker 5 (77016)] cellprofiler_core.worker._worker::do_job: 👺 worker sending MeasurementsReport for job [101]

[Worker 5 (77016)] cellprofiler_core.worker._worker::do_job: 👺 worker finished with job [101]
```

Once stuck, the worker blocks until analysis is terminated. So because we have 9 stuck above, those workers are not used for the remainder of jobs in the pipeline. That's pretty bad.

From the logs, the Worker does its job correctly. It gets a job, runs the pipeline, and sends back a `req` of type `MeasurementsReport`. It is the task of `Boundary.spin` to receive this over a socket, and send it to the upward queue for `interface` and `jobserver` to process. For the case of `MeasurementsReport`, it is the job of `interface` to consolidate it into the aggregate measurements, and then send back and `ACK` to the `Worker` so that it can stop blocking and go about its business.

The `interface` and `jobserver` never receive these hung jobs in their queues. In the log it appears `Boundary.spin` never even receives it.

gnodar01:
One suspect is the following

In `Runner` we have:

```
        # We use a queue size of 10 because we keep measurements in memory (as
        # their HDF5 file contents) until they get merged into the full
        # measurements set.  If at some point, this size is too limiting, we
        # should have jobserver() call load_measurements_from_buffer() rather
        # than interface() doing so.  Currently, passing measurements in this
        # way seems like it might be buggy:
        # https://github.com/h5py/h5py/issues/244
        self.received_measurements_queue = queue.Queue(maxsize=10)
```

This queue is where `MeasurementReport` objects should go. It has `maxsize=10`.

In the `jobserver` we have:

```
elif isinstance(req, anarequest.MeasurementsReport):
    self.queue_received_measurements(req.image_set_numbers, req.buf)
    req.reply(anareply.Ack())
```

where `queue_received_measurements` is:

```
def queue_received_measurements(self, image_set_numbers, measurements):
        self.received_measurements_queue.put((image_set_numbers, measurements))
```

That will obviously encounter the `maxsize` limit on that queue. However this ends up not being (totally) responsible for the issue. If we were using `put_nowait` it would drop it, which would indeed be a problem and lead to a forever hanging `pos-processing` step, because the `Worker` never receives its `ACK`. However we use `put` instead. That will simply block until the queue is below the max capacitance of `10`. 

The `interface` gets from that queue, eventually, and then the `jobserver` continues.

While it is not *THE* problem, it may well be part of the problem. If the `jobserver` is blocked, its not processing other requests. This causes a backlog which is not relieved until `interface` reduces the queue size. That period in which `jobserver` is hung, allowing an accumulation which may exacerbate the underlying problem.

gnodar01:
From [High Water Mark](https://zguide.zeromq.org/docs/chapter2/#High-Water-Marks) section of the ZMQ docs:

> When your socket reaches its HWM, it will either block or drop data depending on the socket type. PUB and ROUTER sockets will drop data if they reach their HWM, while other socket types will block.

In `Boundary.spin` we have a `ROUTER` socket:

```python
request_socket = self.zmq_context.socket(zmq.ROUTER)
request_socket.setsockopt(zmq.LINGER, 0)
request_port = request_socket.bind_to_random_port(
    self.zmq_address)
self.request_address = self.zmq_address + (":%d" % request_port)
```

In the `Worker`, we have a `work_socket` of type `REQ`, connected to that same address:

```python
self.work_socket = self.context.socket(zmq.REQ)
self.work_socket.connect(self.work_request_address)
```

The default highwater mark is `1000`, however also from the same section of the docs:

> Lastly, the HWMs are not exact; while you may get up to 1,000 messages by default, the real buffer size may be much lower (as little as half), due to the way libzmq implements its queues.

In any case, the `request_socket` in `Boundary.spin` is the thing that gets the `MeasurementsReport` from the worker, which is then placed in the upward queue for `interface` and `jobserver`. If it has reached the high water mark, it will drop `MeasurementsReport` message, because it is of type `ROUTER`. That means `jobserver` never gets it, and `interface` never consolidates it.

gnodar01:
Another candidate cause, also due to the use of `ROUTER` is this:

> If you’re using ROUTER sockets, it’s remarkably easy to lose messages by accident, by sending malformed identity frames (or forgetting to send an identity frame). In general setting the ZMQ_ROUTER_MANDATORY option on ROUTER sockets is a good idea, but do also check the return code on every send call.

> 0MQ drops messages it can't route

It could be in some rare cases, that malformed identities are made and those messages never get delivered. 

I think this is less likely, because I have this line in `Boundary.spin`:

```python
req = Communicable.recv(s, routed=True)
if isinstance(req, MeasurementsReport):
    LOGGER.debug(f"👺 Boundary - received measurements report, {req.image_set_numbers}")
```

`Commmunicable.recv` shouldn't be affected by any malformed identities, because its the one that setups up future routing. Up to the point where `recv` is called, we're just using default zmq behavior, nothing custom.

I never receive the debug message for hung jobs, so they presumably never even make it to the `request_socket` (of type `ROUTER`) in the first place.

gnodar01:
Modifying the high water mark, on either the send side or the receive side, did not seem to fix the problem.

In the end a somewhat "hacky" solution worked, which is simply to timeout and retry.

Ultimately a proper fix would likely require a larger refactor that includes thinking through and simplifying the concurrency system we currently have in place. That is a PR for another time.

src/subpackages/core/cellprofiler_core/worker/_worker.py
```diff
@@ -280,26 +282,39 @@ def do_job(self, job):
...
+                        )
+                        # There might be an exception in this call, but it will be
+                        # handled elsewhere, and there's nothing we can do for it
+                        # here.
+                        current_pipeline.post_group(
+                            last_workspace, current_measurements.get_grouping_keys()
+                        )
+                        del last_workspace
+                    else:
+                        LOGGER.error("No workspace for from last image set, cannot run post group")
```
bethac07, line 299: typo

bethac07 (approved):

PR #4944, opened 2024-09-23
Patches

gnodar01:
- fix selected readers not propagating to children
- add some CLI flags
- add profiling scripts
- fix issues with sending objects to logging, causing exceptions
- fixes debugger not working with workers
- name all threads, for debugging purposes
- add a WIP imageiov3 reader

### DIFF
--- a/docs/frontend/source/modules/license.rst
+++ b/docs/frontend/source/modules/license.rst
@@ -52,9 +52,6 @@ License: `GNU General Public License (GPL)`_
 Name: h5py 
 License: `BSD 3-Clause “New” or “Revised” license`_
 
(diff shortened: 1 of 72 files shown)


bethac07 (approved):

bethac07:
Approved at the "this all seems reasonable" level, let me know if I should also be checking it at the "tests pass on my M1" level or other levels. But ty, lgtm!

gnodar01:
Nope, didn't touch anything that propagates down to c-extension stuff. We may want to eventually run profiles on intel vs m1 vs windows, but not particularly useful right now.

Issue #2285, opened 2016-09-02
Rescale intensity relative to maximum intensity in image.


mcquin:
Currently, when loading an image, intensities are rescaled [based on the following criteria](http://forum.cellprofiler.org/t/how-to-get-intensity-row-value/3893/2):
1. A value provided by the user*.
2. The maximum intensity supported by the image, as defined in the image metadata.
3. The maximum value supported by the image data type**.

Instead of defaulting to rescaling by supported maximum intensity (options 2 or 3), CellProfiler should default to rescaling intensities by actual maximum intensity. That is, rescale intensities between 0 and 1 such that the actual minimum intensity value is mapped to 0 and the actual maximum intensity value is mapped to 1 (and all other intensities are adjusted accordingly):

``` python
image = (image - image.min())/(image.max() - image.min())
```

Rescaling intensities in this way has the following benefits:
1. It works.
2. It's consistent and not dependent on image metadata.
3. It won't rescale all values to be teeny-tiny if metadata is missing and the data type is 32-bit integer (in this case, all values would be rescaled by dividing by 65535. Ew!).
4. We won't be dependent on `python-bioformats` to load images and rescale values (this is very important for 3D work).

---

\* Kind of but not really. It's first scaled by supported maximum intensity (option 2 or 3) and then scaled again by the user specified value. This [boolean rescale](https://github.com/CellProfiler/CellProfiler/blob/7fed55ccad5953a7d8be1809a03637997ed398dd/cellprofiler/modules/loadimages.py#L3200-L3203) is not passed into the [call to read](https://github.com/CellProfiler/CellProfiler/blob/7fed55ccad5953a7d8be1809a03637997ed398dd/cellprofiler/modules/loadimages.py#L3220-L3226). Nonzero values evaluate to `True`, so the image is first [rescaled to its supported maximum intensity](https://github.com/CellProfiler/python-bioformats/blob/64bf14550522103c3248a2387b6f6ec0e68e51d6/bioformats/formatreader.py#L882-L883) and then [rescaled again according to the provided value](https://github.com/CellProfiler/CellProfiler/blob/7fed55ccad5953a7d8be1809a03637997ed398dd/cellprofiler/modules/loadimages.py#L3249-L3251).

[...]

PR #4955, opened 2024-11-15
Switch to new scaling logic on image read

### LINKED ISSUE #4954: Rescaling needs overhaul
We noticed that a particular `32` bit image is not being rescaled in CellProfiler.

The image is a `32` bit *float*, not `32` bit *uint*.

If it was `uint32` or `int32` CellProfiler would rescale it, but any `float` type is assumed to already be in the range of `0 - 1` (in this case an incorrect assumption).

It would not be difficult to add in some logic that looks to see if the `dtype` is `float`, and then further checks if the `float` values are all `0 - 1` (in which case don’t rescale it, since it is already scaled), or if the float values go beyond `0 - 1` in either direction, in which case it should be rescaled.

On top of `float` vs `int`, there are other issues with rescaling related to maintaining compatibility with python-bioformats which is no longer necessary in CP5.

Here is the current rescale logic:

```python
    def find_scale_to_match_bioformats(data):
        if data.dtype in (numpy.int8, numpy.uint8):
            return 255
        elif data.dtype in (numpy.int16, numpy.uint16):
            return 65535
        elif data.dtype == numpy.int32:
            return 2 ** 32 - 1
        elif data.dtype == numpy.uint32:
            return 2 ** 32
        else:
            return 1
```

In the reader, whether its bioformats or imageio, we then do this:

```python
        if rescale:
            imax = self.find_scale_to_match_bioformats(data)
            data = data.astype(numpy.float32) / float(imax)
```

That is all sorts of terrible. It’s just plain incorrect.

Firstly we return the max value of the unsigned type, regardless of if the image data type itself is signed or unsigned. For example if we have data that is `int8`, then its min and max values would be `-128` and `127`, but instead we return `255` and divide by that. So in the rescaled version the values could be anywhere between `-0.5019607843137255` and `0.4980392156862745`. Same problem with `16` bit signed and unsigned.

For `32` bit signed and unsigned its even worse. Firstly we return `2^32` for `uint32` when it should really be `2^32-1`. Otherwise we’re potentially dividing the max pixel value of `4_294_967_295` in the image data by `4_294_967_296` which doesn’t get us `1`. But then we return `2^32-1` for `int32`. That has the same problems as with `8` and `16` bit variants, but also has additional problems for `32` bit integers. The image data starts off as `32` bit integers, so depending of if its unsigned or signed, it will be in the range of `0 - 4_294_967_295` or `-2_147_483_648 - 2_147_483_647`. Usually it won’t go that low and that high, e.g. in the example image pixel values are `2_844 - 23_603`, but in general they can be. We then unconditionally cast that image data to `float32` before dividing:

```data = data.astype(numpy.float32) / float(imax)```

`float32` cannot exactly represent all of the integer values that `uint32` or `int32` can. That makes sense because both types have the same number of bits available. Every bit permutation in `int32` and `uint32` represents an integer, whereas `float32` has to represent a bunch of non-whole rational numbers in between the integers as well. `float32` can exactly represent all integer values in the range of `-16_777_216 - 16_777_216`, but beyond those bounds it starts to lose precision. So if the original image contained values beyond those bounds, it loses information when it’s converted from `32` bit integer to `32` bit float.

For illustration:

```data = np.array([1.0, 255.0, 256.0, 65535.0, 65536.0, 16_777_216.0, 16_777_217.0, 4_294_967_285.0], dtype=np.dtype('uint32'))
data

> array([         1,        255,        256,      65535,      65536,
         16777216,   16777217, 4294967285], dtype=uint32)

data.astype(np.float32)

> array([1.0000000e+00, 2.5500000e+02, 2.5600000e+02, 6.5535000e+04,
       6.5536000e+04, 1.6777216e+07, 1.6777216e+07, 4.2949673e+09],
      dtype=float32)
```

Then on top of that, if it’s `int32`, it’s divided by the wrong number as before. It’s also uniquely divided by the wrong number if it’s `uint32` for the reason I covered above.

Summary:

There are 3 sub-issues
1) images of type `float` are not always in the range of `0 - 1`
2) image of type `int[8|16|32]` return the wrong rescale value
3) `uint32` and `int32` return the wrong rescale value, and then lose information on cast

gnodar01:
resolves #4954


gnodar01:
@bethac07 

I've written it so it addresses the issues with #4954 without changing the original spirit of the code, which is scale by the full range allowed by the dtype, rather than the more usual approach (e.g. `skimage.exposure.rescale_intensity`) of scaling by the range of values present in the data array.

It's covered in more detail in the code comments, but the gist is: however narrow or wide of a range the original data covered with respect to the full range allowed by the dtype is preserved.

This is pretty much how we were doing things before, just with the egregious issues around casting fixed. 

What may be a problem is, both wit how we were doing it before and now, is precision loss during the casting. In the case where the image data has values taking up a relatively small portion of the full range allowed by the dtype then the rescaled version will span the same relative portion of the values `0 - 1`, which will in the worst cases lead to loss of information because of precision.

This is made intuitive by realizing, for say 32 bit images, that `uint32` has `2^32` possible values, and `float32` also has `2^32` possible values, but only `2^24 - 1` are between `0 - 1`. So there's just less space in that range, and the values representable in that range are not evenly spread.  If the data values already span a small portion as `uint32`, and we try to smush them all in the same relatively small portion within the `0 - 1` range, we have much less room than if we tried to cover the full range of `0 - 1`. Hence information may be lost, albeit probably in edge cases rather than the normal state of things.

The flip side of this is that if we do scale each image so that its values span the full range of `0 - 1`, then 1) it's quite different to how we did it before and I do not know what effect it will have on downstream image processing and measurements, and 2) inter-image differences are lost. An image that spans a tiny portion of the `uint32` and an image that spans nearly the full range of `uint32` will both span the full range of `0 - 1`. That also has feels like it has implications for downstream processing, though I don't have any concrete cases in mind.

So question: do we scale as `skimage.exposure.rescale_intensity` does (which is also more space/time efficient), or how I have it here currently (less efficient, but similar to how it was before)?

My guess is the latter, for both backwards compatibility (ish) and for the reasons above, but maybe there are factors I'm missing.

gnodar01:
Related: #2285

gnodar01:
Another assumption I have made, that I need to confirm is correct:

If we have an image of type `int[8|16|32]`, then what do we want:
- `   img_data + [ 128 | 32768 | 214748648 ] / (2**[ 8 | 16 | 32] - 1)           ` <- range of ` 0 - 1`
- `( img_data + [ 128 | 32768 | 214748648 ] / (2**[ 8 | 16 | 32] - 1) ) * 2 - 1` <- range of `-1 - 1`

The former is what I was planning to do.
The latter is what `skimage` does in both `skimage.exposure.rescale_intensity` and `skimage.util.img_as_float32` (although the two have different logic).

From docs on `skimage.util.img_as_float32`:

> The range of a floating point image is [0.0, 1.0] or [-1.0, 1.0] when converting from unsigned or signed datatypes, respectively.

On the one hand skimage's approach makes sense. Why have an image as an `int` if you don't want to preserve negative values? On the other hand the answer to that is often "because x tool does a weird thing on purpose or by accident". Also, I don't know if CellProfiler's modules always expect non-negative numbers.

One of the reasons we scale to `0-1` is mentioned in the linked thread:

> Unfortunately, some external libraries (e.g., skimage) work best for images scaled between 0 and 1. Dunno how much would be impacted by removing implicit rescaling.

But actually skimage often (almost always? always always?) works just as well with images scaled between -1 and 1 (as can be seen in the [data type conventions](https://scikit-image.org/docs/stable/user_guide/data_types.html)).
Matplotlib as well (see screenshot), though I can't say confidently about our other libraries (we don't use it, but e.g. I think OpenCV is oriented around `uint8`, or at least it was in the past).

Screenshot 2024-11-22 at 11 56 59 AM

bethac07:
We have generally always (for values of always being "for as long as I've been around") assumed a 0-1 paradigm, and I'm pretty hesitant to change that now just because we really, really don't know what will break. Negative data is also incredibly rare in microscopy (though I'm sure now someone will pop up to say it's possible in FLIM or FRET or AFM or some other case, but still, _rare_)

I think between "convert everything to 0-1" and "convert signed to -1:1 and unsigned to 0-1 (aka what skimage does)",  I've now gone back and forth a couple of times in my head but I've also realized "RescaleIntensity exists", so at the end of the day, people can change stuff however they want if things are barfing. So I'm 60-40 in favor of "We have way more confidence that everything works with 0-1, so let's enforce that and if someone _really_ wants to bring in negative data, they can do it with RescaleIntensity and consequences are on them", but I also think at the end of the day I won't be mad about "if people want to bring in signed data and process it, let's let them try" if you'd rather follow skimage's strategy.

gnodar01:
In `MeasureColocalization`, the image's `scale` property is used. It is also used in some Illastik plugins (`Predict` and `RunIlastik`).

The `scale` is the *original* image scale as reported by either file metadata (e.g. for 12-bit images) or `dtype`.

In the plugins, the image is rescaled up from `0 - 1` back to the original range to be essentially the original image that was read in (excluding any "minor" information loss).

In `MeasureColocalization` it is used for a somewhat different purpose:

```python
            if self.do_costes:
                # Orthogonal Regression for Costes' automated threshold
                scale = get_scale(first_image.scale, second_image.scale)
                if self.fast_costes == M_FASTER:
                    thr_fi_c, thr_si_c = self.bisection_costes(fi, si, scale)
                else:
                    thr_fi_c, thr_si_c = self.linear_costes(fi, si, scale)
```

```python
    def bisection_costes(self, fi, si, scale_max=255):
...
        left = 1
        right = scale_max
        mid = ((right - left) // (6/5)) + left
        lastmid = 0
        # Marks the value with the last positive R value.
        valid = 1

        while lastmid != mid:
            thr_fi_c = mid / scale_max
            thr_si_c = (a * thr_fi_c) + b
            combt = (fi < thr_fi_c) | (si < thr_si_c)
            if numpy.count_nonzero(combt) <= 2:
                # Can't run pearson with only 2 values.
                left = mid - 1
            else:
                try:
                    costReg, _ = scipy.stats.pearsonr(fi[combt], si[combt])
                    if costReg < 0:
                        left = mid - 1
                    elif costReg >= 0:
                        right = mid + 1
                        valid = mid
                except ValueError:
                    # Catch misc Pearson errors with low sample numbers
                    left = mid - 1
            lastmid = mid
            if right - left > 6:
                mid = ((right - left) // (6 / 5)) + left
            else:
                mid = ((right - left) // 2) + left

        thr_fi_c = (valid - 1) / scale_max
        thr_si_c = (a * thr_fi_c) + b

        return thr_fi_c, thr_si_c
```

```python
    def linear_costes(self, fi, si, scale_max=255):
...
        i_step = 1 / scale_max
...
        img_max = max(fi.max(), si.max())
        i = i_step * ((img_max // i_step) + 1)
...
        fi_max = fi.max()
        si_max = si.max()
...
        thr_fi_c = i
        thr_si_c = (a * i) + b
        while i > fi_max and (a * i) + b > si_max:
            i -= i_step
        while i > i_step:
            thr_fi_c = i
            thr_si_c = (a * i) + b
            combt = (fi < thr_fi_c) | (si < thr_si_c)
            try:
                # Only run pearsonr if the input has changed.
                if (positives := numpy.count_nonzero(combt)) != num_true:
                    costReg, _ = scipy.stats.pearsonr(fi[combt], si[combt])
                    num_true = positives

                if costReg <= 0:
                    break
                elif self.fast_costes.value == M_ACCURATE or i < i_step * 10:
                    i -= i_step
                elif costReg > 0.45:
                    # We're way off, step down 10x
                    i -= i_step * 10
                elif costReg > 0.35:
                    # Still far from 0, step 5x
                    i -= i_step * 5
                elif costReg > 0.25:
                    # Step 2x
                    i -= i_step * 2
                else:
                    i -= i_step
            except ValueError:
                break
        return thr_fi_c, thr_si_c
```

`Image.scale` is stored as a single number, and sometimes had a not very correct value for the same reasons stemming from issues discussed above.

Now (as of the latest commit), it is a single number, and allows you to get back to a *range* of image intensities similar but the the same as the original image. If the original image was a `dtype` of `uint`, and metadata was not used, then you get the original image back. Otherwise, there may be discrepancies. For instance if the original image was a `dtype` of `int8`, and had values between `-128 - 127`, then instead what you get is an image with values between `0 - 255`.

From a (very) shallow read of the colocalization logic, that is actually a good thing.

src/frontend/cellprofiler/gui/parametersampleframe.py
```diff
@@ -56,6 +56,11 @@
 ID_SAMPLE_BUTTON = wx.NewId()
 
 
+# TODO: we have decided to tear out parameter sampling
+# it is currently broken, and even if we fix it, it writes out
+# data to files that are then meant to be browsed in another APP that no longer exists.
```
bethac07, line 61: Well, it does exist, you link to it ;)

src/subpackages/core/cellprofiler_core/constants/modules/namesandtypes.py
```diff
@@ -17,22 +17,26 @@
...
 RESCALING_HELP_TEXT = """\
 This option determines how the image intensity should be rescaled from
 0.0 – 1.0.
 
+-  *{INTENSITY_RESCALING_BY_DATATYPE}:* Rescale the image to 0 – 1
+   by dividing by 255 or 65535, depending on the number of bits used
+   to store the image.
 -  *{INTENSITY_RESCALING_BY_METADATA}:* Rescale the image intensity
-   so that saturated values are rescaled to 1.0 by dividing all pixels
+   so that saturated values are rescaled to 0 - 1 by dividing all pixels
```
bethac07, line 39: if you're putting a min here, you want just "values", not "saturated values"

src/subpackages/core/cellprofiler_core/image/abstract_image/file/_file_image.py
```diff
@@ -104,7 +114,121 @@ def __init__(
...
+
+        The conversion also handles signed types by shifting the data first.
+
+        Finally the conversion handles three conditions present in floating
+        point images:
+        1. Floating point images which already have values inside the range 0-1.
+           In this case, it is simply cast to float32.
+        2. Floating point images which have values normalized inside the range -1-1.
+           In this case, the data is renormalized to the range 0-1, and then cast to float32.
+        2. Floating point values which *represent* signed/unsigned integers but stored as float
```
bethac07, line 209: MAX PEDANTRY - `3.`

src/subpackages/core/cellprofiler_core/image/abstract_image/file/_file_image.py
```diff
@@ -313,23 +465,25 @@ def __set_image(self):
 
         self.cache_file()
         channel_names = []
+
         if is_matlab_file(self.__filename):
             img = load_data_file(self.get_full_name(), loadmat)
-            self.scale = 1.0
+            # TODO - 4955: should we be overwriting if not None?
```
bethac07, line 471: So here's where you're going to punch me for not bringing this up earlier, mostly because I just didn't think about it because they aren't _images_; mat and npy were in general intended in CellProfiler to be a special case - they're intended for flatfield illumination correction images, which we do NOT actually want rescaled. We want the scaling as is, for reasons easier to explain in person (but can drop a link too if you'd rather). 

I'm not sure how useful the matlab reader is, because it requires the table have [a very specific name](https://github.com/CellProfiler/CellProfiler/blob/main/src/subpackages/core/cellprofiler_core/utilities/image.py#L129-L133), and I'm not sure if anything but CellProfiler 1 and CellProfiler 2 ever consistently made such a .mat file with such a named table. But for the numpy reader, when it is called by loading something as an illum correction function, we want to not rescale. 

(I think adding the ability to treat numpy arrays as normal images is a nice-to-have; I can point you at internal slack messages from the last few months where people in the lab wanted to be able to read npy files as images and we had to tell them "no, convert them to tiffs"; I think in the age of deep learning, this will happen more and more. But we do also need to be able to support the special case of "a numpy array being loaded as an illum correction image, which you should please not touch")

src/subpackages/core/cellprofiler_core/modules/namesandtypes.py
```diff
@@ -1526,15 +1539,16 @@ def add_simple_image(
...
+                metadata_rescale=metadata_rescale,
                 volume=volume,
                 spacing=spacing,
                 z=z,
                 t=t
             )
         elif load_choice == LOAD_AS_ILLUMINATION_FUNCTION:
             provider = MonochromeImage(
-                name, url, series, index, channel, False, volume=volume, spacing=spacing, z=z, t=t
+                name, url, series, index, channel, rescale_range=None, metadata_rescale=False, volume=volume, spacing=spacing, z=z, t=t
```
bethac07, line 1551: See note above about how need to ensure this keeps being treated as a special case without rescaling.

src/subpackages/core/cellprofiler_core/image/abstract_image/file/_file_image.py
bethac07: I'm super, duper sorry for not having thought about this before!

bethac07 (approved):
Nice work! I'm assuming you'll clean up the last couple little TODOs and failing tests and don't need my thoughts beyond here, sorry this became such a weird sidequest but I like the cleanups you proposed.

gnodar01:
@bethac07 

[test_load_raw](https://github.com/CellProfiler/CellProfiler/blob/main/tests/core/modules/test_namesandtypes.py#L1848-L1859)

makes a 12-bit tiff image, with values 0 - 4096, with the proper metadata set, and dtype as `uint16`

```
path = tests.core.modules.make_12_bit_image(folder, file_name, (34, 19))
```

then it runs a workspace, with just `NamesAndTypes` as the module, and with the "load as illumination function" set

```
    workspace = run_workspace(
        path, cellprofiler_core.modules.namesandtypes.LOAD_AS_ILLUMINATION_FUNCTION
    )
```

and expects all values in the resulting image data to be `< 1/16` which is `2**12 / 2**16`

```
    assert numpy.all(pixel_data <= 1.0 / 16.0)
```

[This causes](https://github.com/CellProfiler/CellProfiler/blob/main/src/subpackages/core/cellprofiler_core/modules/namesandtypes.py#L1537) `namesandtypes` to set `rescale = False`

Which means the [reader won't rescale](https://github.com/CellProfiler/CellProfiler/blob/main/src/subpackages/core/cellprofiler_core/readers/bioformats_reader.py#L211-L215), and neither will [FileImage](https://github.com/CellProfiler/CellProfiler/blob/main/src/subpackages/core/cellprofiler_core/image/abstract_image/file/_file_image.py#L325-L334) (although the `.npy` check does not get triggered, because we have a `.tiff`)

But `FileImage`, calls `Image`, which gets a `uint16` array and [does its own rescaling](https://github.com/CellProfiler/CellProfiler/blob/main/src/subpackages/core/cellprofiler_core/image/_image.py#L158-L159), not with the `scale` value returned from the reader, but by the `dtype`, so the divisor is `2**16-1`.

This is what `test_load_raw` **expects**, with its assertion that values in `pixiel_data` be less than `1/16`, since `2^12/2^16 = 1/16`.


In my version, I straight up don't rescale anywhere: in the reader, `FileImage`, or `Image`. The `Image` class still has that rescaling logic, but I cast the `uint16` to `float32` before its reached, so it never triggers the rescaling there. I can easily not do the casting and have the same results as the original. Is this desired?

bethac07:
I'm truly not sure, we're into the deep lore here, moving stuff to Slack

PR #4956, opened 2024-11-19
Add log scale option in rescaleintensity

### LINKED ISSUE #4951: Add log scaling as option to RescaleIntensity module
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
I'm curious if anyone has thought to add the log scaling originally used in the image context menu to `RescaleIntensity` or elsewhere. The purpose would be to create an image which has good contrast for say, overlaying outlines for export, etc.

**Describe alternatives you've considered**
What i've been doing thus far is re-creating in `ImageMath` the operations here: https://github.com/CellProfiler/CellProfiler/blob/908d7b517c42329e3436fff7f93874ce71e8fc56/cellprofiler/gui/artist.py#L638-L643. but this requires more modules than i'd like 🙂 and simply clipping/stretching the intensities doesn't cut it. 
image
image
image

Antsalacia:
Resolves #4951 

I have added a new option in the `RescaleIntensity` class to be able able to have higher contrast on image. The code used was the same as shared by [braymp](https://github.com/braymp) in the issue #4951.

### DIFF
--- a/src/frontend/cellprofiler/modules/rescaleintensity.py
+++ b/src/frontend/cellprofiler/modules/rescaleintensity.py
@@ -43,6 +43,7 @@ M_DIVIDE_BY_IMAGE_MAXIMUM = "Divide by the image's maximum"
 M_DIVIDE_BY_VALUE = "Divide each image by the same value"
 M_DIVIDE_BY_MEASUREMENT = "Divide each image by a previously calculated value"
 M_SCALE_BY_IMAGE_MAXIMUM = "Match the image's maximum to another image's maximum"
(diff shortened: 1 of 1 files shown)


gnodar01:
Thanks for your contribution. I see the PR is still in draft, but just to preemptively catch an often missed nuance: you will need to update the revision number, and account for any changes in `upgrade_settings`. Please read the following: https://github.com/CellProfiler/CellProfiler/wiki/Module-structure-and-data-storage-retrieval#citi

PR #4958, opened 2024-11-27
warn in correctillumapply if applying image to itself

ErinWeisbart:
My goal is for the GUI to throw the Red X if, for any image, the user has set the `input image` and the `illumination function` to the same image.

In a test where I set the `input image` and the `illumination function` to the same image, I know that the validation error is being reached because if I add a print statement after `if image_name == illum_correct_name:` it prints. However, there is no Red X. 

If I put the check in `validate_module()` (instead of `validate_module_warnings`) it behaves similarly

(I'm not sure how to pass the specific image that is raising the error as if I put `self.images.image` in the ValidationError() (instead of the `self.images` that's in this draft) it throws `Exception in cpmodule.text_valid 'list' object has no attribute 'image'`)

### DIFF
--- a/src/frontend/cellprofiler/modules/correctilluminationapply.py
+++ b/src/frontend/cellprofiler/modules/correctilluminationapply.py
@@ -366,6 +366,14 @@ somewhat empirical.
                     % image.rescale_option,
                     image.divide_or_subtract,
                 )
+        for image in self.images:
+            image_name = image.image_name.value
+            illum_correct_name = image.illum_correct_function_image_name.value
+            if image_name == illum_correct_name:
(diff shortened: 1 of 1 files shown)


ErinWeisbart:
We also checked error as a Bool:

```
    def validate_module(self,pipeline):
        error = False
        for image in self.images:
            image_name = image.image_name.value
            illum_correct_name = image.illum_correct_function_image_name.value
            if image_name == illum_correct_name:
                print('print statement')
                error = True
        if error:
            raise ValidationError(
                "You are applying an image to itself. Check your input image/s and illumination function/s.",
                self.images,
            )
```

Issue #4583, opened 2022-07-20
Failure to keep row order of measurements when IDprimary = IDsecondary


rsenft1:
**Describe the bug**
If you have a primary object (nucleus) the same size as your secondary object (cytoplasm) and then make tertiary objects from their difference (WITH option Shrink smaller object prior to subtraction set to NO), then measure these objects, there are resulting issues. For one, the parent # is wrong (0) because there are no pixels to measure parent identity in. Worse than this, the subsequent measurements for objects in that image are now offset, which is pretty bad. 

**To Reproduce**
Attaching a project file that works with the built-in fly pipeline

**Expected behavior**
Parent # should be correct regardless of object size and measurements should be nan if there are no pixels to an object, not offset the rest of the table. 

**Screenshots**
![image](https://user-images.githubusercontent.com/28116530/180003492-e1e1003a-6ed3-43be-998b-0e2df2fbe359.png)

**Desktop (please complete the following information):**
Mac from source (GitHub) and built.

[flytest.cpproj.zip](https://github.com/CellProfiler/CellProfiler/files/9150578/flytest.cpproj.zip)

PR #4990, opened 2025-04-11
Issues: #4583 - Bugfix - Inconsistencies when objects with 0 area are measured 

sspathak:
Bug fix:
When `measureobjectsizeshape` module is called with "invisible" objects (i.e. when `identifytertiaryobjects` creates an object but it doesn't exist due to its primary and secondary objects having a perfect overlap), fill the measurements for that object with `NaN` or `0`
We made the decision to disallow ghost objects. Modules need to keep track of when a ghost object could potentially generated and implement logic to ensure that labels (`objects.segmented`) contains at least one pixel of each object. 

This PR also contains commits from callum-jpg's contribution to porting `measureobjectsizeshape` to cellprofiler library. 
refer to #5005.


tests/core/core/object/test__objects.py
```diff
@@ -126,3 +128,137 @@ def test_center_of_mass(self):
...
+    child_labels[3:7, 3:7] = request.param[0]
+    child_labels[2:3, 13:17] = request.param[1]
+    child_labels[7:8, 13:17] = request.param[2]
+    child_labels[3:7, 23:27] = request.param[3]
+    child.segmented = child_labels
+
+    return objects, child
+
+@pytest.fixture(scope="function")
+def create_object_primary_overflows_equal_parts_into_other_object():
```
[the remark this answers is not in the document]
sspathak: I appreciate your concern! This test case is temporary and will likely be removed. 

I had a discussion with @gnodar01 yesterday where we concluded that the bug, that causes the parent-child relation of ghost objects to break, originates somewhere around [here: identifytertiaryobjects.py#L304](https://github.com/CellProfiler/CellProfiler/blob/9638387750dab24aded02002d7ede4479438a43a/src/frontend/cellprofiler/modules/identifytertiaryobjects.py#L304). One potential fix is to run a check to ensure that the `parent_of_child` array returned by `_objects.relate_children()` is monotonically increasing and when not, handle that error in `measuretertiaryobjects.py`. However, it may be the case that the output `parent_of_child` is always guaranteed to be a range from 1 to n where n is the total number of objects and that we need not call `secondary_objects.relate_children(tertiary_objects)` [here: identifytertiaryobjects.py#L304](https://github.com/CellProfiler/CellProfiler/blob/9638387750dab24aded02002d7ede4479438a43a/src/frontend/cellprofiler/modules/identifytertiaryobjects.py#L304) at all.

Based on the current test set, the only scenario when the `parent_of_child` array is not a sequence from 1 to n is when:
1. A child object's labels are not present in `objects.segmented` i.e. a ghost object exists
2. A child object's labels are outside the parent or inside another parent

Do know any other edge cases that could cause `parent_of_child` to not be a sequence?

bethac07:
Added a discussion for the renumbering stuff - #4999

tests/frontend/modules/test_identifytertiaryobjects.py
```diff
@@ -357,17 +400,17 @@ def test_do_not_shrink_identical():
...
+        for child in (1, 2, 3, 4):
             assert parent_of[child - 1] == child
 
     for location_feature in (
         M_LOCATION_CENTER_X,
         M_LOCATION_CENTER_Y,
     ):
         values = measurements.get_current_measurement(TERTIARY, location_feature)
-        assert numpy.all(numpy.isnan(values) == [False, True, False])
+        assert numpy.all(numpy.isnan(values) == [False, False, False, False])
```
sspathak, line 413: This test (merged 6 years ago) checked for ghost objects, expected the object to disappear, and have nan parent object numbers

bethac07:
Note here - we have decided rather than trying to patch 0-pixel-object measurements, to remove ability to generate them entirely, hence the pivot!

src/frontend/cellprofiler/modules/identifytertiaryobjects.py
```diff
@@ -304,6 +317,7 @@ def run(self, workspace):
         child_count_of_secondary, secondary_parents = secondary_objects.relate_children(
```
sspathak, line 317: Considering the assumption that object numbers are continuous and each secondary object is expected to produce exactly one tertiary object, we could replace the `.relate_children()` call here with the expected outputs since it will always be `([1, 1, ...., 1], [1, 2, ..., n])` where `n` is the number of objects.

src/frontend/cellprofiler/modules/identifytertiaryobjects.py
```diff
@@ -304,6 +317,7 @@ def run(self, workspace):
...
             tertiary_objects
         )
+
         if self.shrink_primary:
             child_count_of_primary, primary_parents = primary_objects.relate_children(
```
sspathak, line 322: see the comment above on replacing the `.relate_children()` call

sspathak:
#4852 should be mreged into main first, as it will introduce merge conflicts that will need to be resolved here first.

gnodar01:
Incorporates #4852

gnodar01 (approved):

gnodar01:
Closing in favor of #5023

Issue #4989, opened 2025-04-11
Change default color for image display in Workspace viewer


emiglietta:
Blue is a terrible color for a default image display in the workspace viewer. 

Could you make the default white? pretty please?
Thanks!

bethac07:
Closed by #4994

PR #4993, opened 2025-04-14
Issue: #4989 Updated the default color assigned to the image in workspace view

sspathak:
Cells are no longer blue by default! Changed from blue to white on popular demand.


sspathak:
Invalid PR. Branch from main instead of issues/<other branch>

PR #4994, opened 2025-04-14
Issue: #4989 Default color of image in workspace view changed to white

sspathak:
Cells are no longer blue by default! Changed from blue to white on popular demand.

### DIFF
--- a/src/frontend/cellprofiler/gui/workspace_view/_workspace_view.py
+++ b/src/frontend/cellprofiler/gui/workspace_view/_workspace_view.py
@@ -282,6 +282,16 @@ class WorkspaceView:
                 else get_tertiary_outline_color() if row == 3
                 else wx.WHITE
             )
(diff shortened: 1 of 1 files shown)


gnodar01 (approved):

src/frontend/cellprofiler/gui/workspace_view/_workspace_view.py
```diff
@@ -282,6 +282,16 @@ def add_row(self, row_class, rows, grid_sizer, can_delete):
                 else get_tertiary_outline_color() if row == 3
                 else wx.WHITE
             )
+        elif row_class == WorkspaceViewImageRow:
```
bethac07, line 285: Just FMI, why add a new thing rather than just change the order? Fine either way just curious

gnodar01 (approved):

Issue #4509, opened 2022-02-15
In FilterObjects Relabel additional object set just matches object numbers in the filtered set


rsenft1:
**Describe the bug**
In FilterObjects, you can optionally relabel an additional object set to match the filtered set to keep relationships between parents/children (e.g., cells and nuclei) consistent. When I was using this, I realized that the second object set is not filtered based on parent-child relationships, but just by object number. So if I have a set of objects 1 to n and I remove object #2 through filtering,  relabeling an additional set will also remove object #2, which might not relate at all to #2 in the first object set. 

If this is the intended behavior, I think that needs to be clarified in the documentation with an example. I would expect this to respect parent-child relationships or potentially be overlap-based, but inclusion of a RelateObjects before FilterObjects does not change this behavior. There are plenty of situations where you might generate cells and nuclei from modules other than IDPrimary and IDSecondary and the current wording of the 'relabel additional object set' feature makes it sound like you could use FilterObjects to help establish a 1:1 relationship between object sets but it doesn't appear to work that way. 

**To Reproduce**
Will add pipeline with example images

**Screenshots**
Filtering reduces objects to #s 15 and 25 in the original image
<img width="1284" alt="Pasted Graphic 1" src="https://user-images.githubusercontent.com/28116530/154130830-3395be15-775f-4e36-a582-2037d14749f9.png">

Relabeling an additional set just keeps #s 15 and 25 from that set: 
<img width="1288" alt="Pasted Graphic 2" src="https://user-images.githubusercontent.com/28116530/154130881-5c85ce03-54d2-432e-9720-9d4d62812100.png">

If we filter based on object number and there are no objects of that number in the relabeling image, it gets rid of all objects in set 2.
<img width="1298" alt="Pasted Graphic 3" src="https://user-images.githubusercontent.com/28116530/154130916-0c48376f-6954-448d-a5b1-4f2781ae0fa6.png">

CellProfiler installed from source on Windows VM.

rsenft1:
Here's the pipeline as a text file. You should be able to change the extension to "cppipe" and open it in CellProfiler.
[example_bug.txt](https://github.com/CellProfiler/CellProfiler/files/8074126/example_bug.txt)

bethac07:
I'm not sure this issue is closed - Rebecca's issue explicitly states that adding a RelateObjects step doesn't fix the problem, at least as of then. Can we confirm now, that if relationship is present, filtering happens correctly?

bethac07:
[...]

PR #5004, opened 2025-05-12
FilterObjects relabel mismatch Issues/4509

sspathak:
As discussed with @bethac07 last week, the #4509 issue bug occurs when there is a mismatch in the labels assigned to the object being filtered and the `Additional Object`s that the user wishes to relabel. A decision was made to add a note in the docs, clarifying that the user needs to `RelateObjects` prior to using this module to ensure that the labels assigned to the newly created object sets match the labels of the object being filtered. 
We chose this option instead of automatically relabeling the `Additional Object` in order to avoid redundant functionality among modules.


sspathak:
Screenshot 2025-05-12 at 10 13 40 AM

Preview of the updated docstring

gnodar01 (approved):

PR #5007, opened 2025-05-20
Make dark-mode/light-mode logo switch

### LINKED ISSUE #5006: Logo background color inivisible in dark mode
**When in dark mode, the Logo is invisible. It could use a different image as logo with a background or at least a stroke on the font.**

Image

## Proposal:
for this specific image to load this version:

Image

as  splash_alt.pgn at  `/CellProfiler/CellProfiler/blob/main/src/frontend/cellprofiler/data/images/`

gnodar01:
Resolves #5006


bethac07 (approved):
I suppoooooose we can support darkmode

PR #5012, opened 2025-05-23
Move `convertimagetoobjects.py` to library

sspathak:


src/subpackages/library/cellprofiler_library/types.py
```diff
@@ -91,6 +91,8 @@ def validate_object_label_set(label_set: list) -> list:
 ImageGrayscale = Union[Image2DGrayscale, Image3DGrayscale]
 ImageGrayscaleMask = Union[Image2DGrayscaleMask, Image3DGrayscaleMask]
 
+ImageAny = Union[Image2DColor, Image3DColor, Image2DGrayscale, Image3DGrayscale]
```
sspathak, line 94: ImageAny added but not used. The breaking change was that this module previously accepted any image as input. However, it has unexpected behavior for color images. So the module has been updated to support only grayscale images

gnodar01 (dismissed):

PR #5018, opened 2025-06-10
Moved `correctilluminationapply` to library

sspathak:
Merge #5012 before this PR

### DIFF
--- a/src/frontend/cellprofiler/modules/correctilluminationapply.py
+++ b/src/frontend/cellprofiler/modules/correctilluminationapply.py
@@ -37,9 +37,11 @@ from cellprofiler_core.setting.do_something import DoSomething
 from cellprofiler_core.setting.do_something import RemoveSettingButton
 from cellprofiler_core.setting.subscriber import ImageSubscriber
 from cellprofiler_core.setting.text import ImageName
+from cellprofiler_library.opts.correctilluminationapply import Method
+from cellprofiler_library.modules._correctilluminationapply import correct_illumination_apply
 
-DOS_DIVIDE = "Divide"
-DOS_SUBTRACT = "Subtract"
+DOS_DIVIDE = Method.DIVIDE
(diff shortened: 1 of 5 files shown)


src/frontend/cellprofiler/modules/correctilluminationapply.py
```diff
@@ -248,50 +250,33 @@ def run_image(self, image, workspace):
...
-        # Optionally, clip high and low values
-        #
-        if self.truncate_low.value:
-            output_pixels = numpy.where(output_pixels < 0, 0, output_pixels)
-        if self.truncate_high.value:
-            output_pixels = numpy.where(output_pixels > 1, 1, output_pixels)
-        
+        except ValueError as e:
+            # Update error message to be more descriptive if the error is due to image and illum function being different shapes
+            if "This module requires that the image and illumination function have equal dimensions" in str(e):
```
sspathak, line 266: Capture and raise added here to provide a better error message for a specific error (to include the name of the images in the error message)

src/frontend/cellprofiler/modules/correctilluminationapply.py
```diff
@@ -248,50 +245,33 @@ def run_image(self, image, workspace):
...
+                    "If they are paired correctly you may want to use the Resize or Crop module to make them the same size."
+                    % (
+                        image_name,
+                        illum_correct_name,
+                        orig_image.pixel_data.shape,
+                        illum_function_pixel_data.shape,
+                    )
+                )
+            else:
+                raise e
```
sspathak: Keeping the validation (and auto-fix) of illum function in the FE. Fixing an invalid illum function is not the library's responsibility and would be unfair to other modules that do not auto-fix invalid inputs. Instead, I've added an assert in the library to validate that the shapes match and throw a message otherwise.

gnodar01 (approved):

PR #5023, opened 2025-06-16
Dissallow ghost objects

gnodar01:
- all objects at least 1 pixel


gnodar01:
Added a discussion for the renumbering stuff - https://github.com/CellProfiler/CellProfiler/discussions/4999

The Breaking Change here is that ghost objects cannot exist, and are always replaced by a 1 pixel object, which will show up in measurements.

PR #5025, opened 2025-07-01
Bug fix: FilterObjects parent-child relationship

### LINKED ISSUE #4509: In FilterObjects Relabel additional object set just matches object numbers in the filtered set

sspathak:
Bug:
FilterObjects did not respect parent-child relationship when filtering.

Changes:
1. Introduced a new setting to allow keeping/removing unassociated child objects.
2. Updated logic to utilize parent object number from measurements (if available) to determine which child objects to keep
3. Added new tests to capture the bug

### DIFF
--- a/src/frontend/cellprofiler/modules/filterobjects.py
+++ b/src/frontend/cellprofiler/modules/filterobjects.py
@@ -141,7 +141,7 @@ PO_ALL = [PO_BOTH, PO_PARENT_WITH_MOST_OVERLAP]
 class FilterObjects(ObjectProcessing):
     module_name = "FilterObjects"
 
-    variable_revision_number = 10
+    variable_revision_number = 11
 
     def __init__(self):
         self.rules = Rules()
@@ -519,6 +519,17 @@ value will be retained.""".format(
         )
 
         group.append(
+            "keep_unassociated_objects",
+            Binary(
+                "Keep unassociated objects?",
+                False,
+                doc="""\
+                Select *{YES}* to keep objects that do not have a parent object. Select *{NO}* to only keep child objects of the objects that get selected by the filter.
+                """
+            )
+        )
+
+        group.append(
             "remover",
             RemoveSettingButton(
                 "", "Remove this additional object", self.additional_objects, group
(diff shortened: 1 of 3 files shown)


src/frontend/cellprofiler/modules/filterobjects.py
```diff
@@ -742,19 +753,56 @@ def run(self, workspace):
...
         ]
         m = workspace.measurements
         first_set = True
-        for src_name, target_name in object_list:
+        for src_name, target_name, keep_unassociated_objects in object_list:
             src_objects = workspace.get_objects(src_name)
             target_labels = src_objects.segmented.copy()
+
+            # Parent relation is used if it exists (use RelateObjects module)
+            parent_relation_exists = len([i for i in m.get_measurement_columns() if i[0] == src_name and i[1] == f'Parent_{self.x_name.value}']) > 0
```
sspathak, line 766: Potentially hard-coded 'Parent_'

sspathak (commented):
Added comments/hints for changes

PR #5038, opened 2025-08-07
Move `morph` to library

sspathak:
NOTE: Merge #5087 `dilateobject` before this

### DIFF
--- a/src/frontend/cellprofiler/modules/morph.py
+++ b/src/frontend/cellprofiler/modules/morph.py
@@ -268,7 +268,7 @@ from cellprofiler_core.setting.subscriber import ImageSubscriber
 from cellprofiler_core.setting.text import ImageName, Integer
 
 from cellprofiler_library.modules._morph import apply_morphological_operations
-from cellprofiler_library.opts.morph import MorphFunction, F_ALL
+from cellprofiler_library.opts.morph import MorphFunction, MorphFunctions_All as F_ALL
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


src/frontend/cellprofiler/modules/morph.py
```diff
@@ -310,9 +311,9 @@
     F_VBREAK,
 ]
 
-R_ONCE = "Once"
-R_FOREVER = "Forever"
-R_CUSTOM = "Custom"
+R_ONCE = RepeatMethod.ONCE
+R_FOREVER = RepeatMethod.FOREVER
+R_CUSTOM = RepeatMethod.CUSTOM
```
gnodar01: Let's remove this from library opts, and keep them frontend only. They don't affect library behaviour, and library never sees these values.

src/subpackages/library/cellprofiler_library/modules/_morph.py
```diff
@@ -0,0 +1,148 @@
...
+            MorphFunction.FILL,
+            MorphFunction.HBREAK,
+            MorphFunction.MAJORITY,
+            MorphFunction.REMOVE,
+            MorphFunction.SHRINK,
+            MorphFunction.SKELPE,
+            MorphFunction.SPUR,
+            MorphFunction.THICKEN,
+            MorphFunction.THIN,
+            MorphFunction.VBREAK,
```
gnodar01: This can be simplified to `not is_binary and function_name != MorphFunction.OPENLINES`.

gnodar01 (approved):

PR #5040, opened 2025-08-12
Be more careful on dtype conversion

### LINKED ISSUE #5039: Beginner Segmentation broken in cp5
- load `segmentation_final.cppipe` with the `images_ilum-corrected` image set
- click "Update" in NamesAndTypes
- error:
> Encountered unrecoverable error in NamesAndTypes during startup: int() argument must be a string, a bytes-like object or a number, not 'NoneType'

The error has to do with the `aggregate_metadata` step added in CP5

gnodar01:
resolves #5039


bethac07 (approved):
Nice catch!

PR #5050, opened 2025-09-24
Update to better reflect current dev standards

gnodar01:


gnodar01:
Nothing at all wrong with PyCharm, but I'd hardly go so far as *recommending* it over the other options 😉

gnodar01:
I know that markdown supports auto-numbering when you use all `1.`, but the newlines were messing it up, and autonumbering was not always actually applied.

gnodar01:
Releasing new version is programmatic now, so no longer relevant.

bethac07 (approved):
TY for the cleanup!

PR #5059, opened 2025-10-24
Move `identifyprimaryobjects`, `identifysecondaryobjects`, `identifytertiaryobjects` to library

sspathak:
TODO: ~InputType for `number_of_deviaitons` needs to be changed to `float` instead of `int`~ DONE

### DIFF
--- a/src/subpackages/library/cellprofiler_library/functions/image_processing.py
+++ b/src/subpackages/library/cellprofiler_library/functions/image_processing.py
@@ -382,7 +382,7 @@ def get_threshold_robust_background(
     upper_outlier_fraction: float = 0.05,
     averaging_method:       Threshold.AveragingMethod = Threshold.AveragingMethod.MEAN,
     variance_method:        Threshold.VarianceMethod = Threshold.VarianceMethod.STANDARD_DEVIATION,
-    number_of_deviations:   int = 2,
+    number_of_deviations:   float = 2.0,
 ) -> float:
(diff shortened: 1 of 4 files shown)


gnodar01:
Supercedes #4783, with changes rolled in (credit to @callum-jpg).

src/frontend/cellprofiler/modules/threshold.py
```diff
@@ -788,9 +787,9 @@ def run(self, workspace):
         self.add_threshold_measurements(
             self.get_measurement_objects_name(),
             workspace.measurements,
-            final_threshold,
-            orig_threshold,
-            guide_threshold,
+            self.final_threshold,
+            self.orig_threshold,
+            self.guide_threshold,
```
gnodar01: This works, but is confusing on first read.

`get_threshold` sets `self.final_threshold` etc, and also returns the values it sets. On line 782 we create a local `final_threshold`, and line 790 we reference not the returned one (`final_threshold`) but the one we set earlier (`self.final_threshold`). On line 813 we do the opposite, reference the returned `final_threshold` rather than `self.final_threshold`.

All the way through they happen to be the same values anyway, but that can easily be messed up by a careless refactor.

sspathak: Makes sense. `get_threshold()` should not be setting anything. I've updated the logic in the next commit

PR #5075, opened 2025-11-20
Move `measureobjectneighbors` to library

sspathak:
- [ ] Review changes related to functions that were copied from `core/object.py` to `library/functions/object_processing.py`
- [ ] Review the MeasurementScale Enum and its uses
- [ ] Review the names given to functions in `measurement.py` and text inside type Annotated in `_measureobjectneighbors.py`


src/subpackages/library/cellprofiler_library/functions/measurement.py
```diff
@@ -2992,3 +3000,537 @@ def make_objskeleton_graph(
...
+        NDArray[numpy.int_],
+        Optional[NDArray[numpy.int_]],
+    ]:
+    labels: ObjectSegmentation = objects_small_removed_segmented.copy()
+    neighbor_labels: ObjectSegmentation = neighbor_small_removed_segmented.copy()
+    
+    kept_label_set = cast_labels_to_label_set(kept_labels)
+    neighbor_kept_label_set = cast_labels_to_label_set(neighbor_kept_labels)
+    kept_label_ijv = convert_label_set_to_ijv(kept_label_set, validate=False)
+    has_pixels = areas_from_ijv(kept_label_ijv) > 0
```
sspathak: These functions from functions/segmented can be expensive. Library user should be able to pass these as args in order to avoid having to re-compute these values

sspathak (commented):
Added comments on cacheing and computing parameters that can be derived from other parameters

gnodar01:
superseded by #5092

PR #5076, opened 2025-11-21
Move `measureimagequality` to library

sspathak:


src/frontend/cellprofiler/modules/measureimagequality.py
```diff
@@ -1084,421 +1044,65 @@ def post_run(self, workspace):
...
-        result = []
-        for image_name in self.images_to_process(image_group, workspace):
+        if image_names is None:
+            logging.warning("[MeasureImageQuality]: No images to process in the image group")
+            return []
+        
+        scale_groups = [scale_group.scale.value for scale_group in image_group.scale_groups]
+        all_threshold_groups = self.get_all_threshold_groups(image_group)
+        
+        for image_name in image_names:
```
sspathak: This is a significant change in the code for this module. Previous version iterated over images for every feature. As a result, the same image would load for each feature. The new version optimizes this by loading the image once and uses the same pixel data for all features.

sspathak (commented):
Added some notes

gnodar01:
superseded by #5114

PR #5088, opened 2026-01-06
Move `rescaleintensity` to library

sspathak:
- [x] Note removal of unused legacy opts
- [x] `CUSTOM_VALUE` variable belongs to both minimum and maximum settings and was re-used in places. Now the opts for each of minimum and maximum have their own `CUSTOM_VALUE` enum value 
- [x] `test_rescaleintensity.py` generates a UInt8 image for testing. Seems intentional but I can't find any justification in the docs. For now, I've added a new type exclusively for this module. If uint is an error, the new type `ImageUInt` can be removed
- [ ] Measurement related code to be updated once measurements structure is finalized
- [x] Discuss pydantic type validator and shared dictionary interaction (Pydantic creates a copy so any code that assumes that the dict is passed by reference and updated by some function needs to be refactored)
- [x] Note the change in acceptable dtypes for the Pixel type. (I had to create a `Pixel_` type for compatibility with Pydantic's `AfterValidator`)
- [x] Update the parameter and return types of `RemoveHoles` in library to be `ImageAny` instead of a `Union` of `ImageAny` and `NDArray[np.uint8]`.


sspathak:
Continuing our discussion on the `ImageUInt` type:
I initially had `float32` the only acceptable dtype for the `ImageAny` type. CellProfiler loads images as `float32`, most modules work with `float32` images and the image dtype almost never changes. Most of CellProfiler's test cases use a `float32` image as input as well. 
However, the underlying algorithms of CellProfiler do not mandate `float32` and most of them work with other dtypes perfectly (exception include `MeasureTexture` which has a dependency for its Haralick features that requires a specific dtype). So, there is no reason to force library users to input `float32` images.
I've updated the `Pixel` type to include other dtypes like `int` and `uint`.

gnodar01 (approved):

PR #5098, opened 2026-01-14
Move `makeprojection` to library

sspathak:
This PR has 2 changes
1. Splitting the ImageProvider into smaller, single-purpose classes
2. Refactoring the ImageProvider to make it stateless.

The original ImageProvider class was very large and had several conditionals at each class method which branched depending on the accumulation method selected. Since the FE only uses one accumulation method at a time, it makes sense to split up the monolith class into smaller classes that are easier to read and understand.

To make the code stateless, I've refactored the accumulator functions to accept a `state` variable that serves as a replacement of the instance variables that the ImageProvider instances had.

### DIFF
--- a/src/frontend/cellprofiler/modules/makeprojection.py
+++ b/src/frontend/cellprofiler/modules/makeprojection.py
@@ -53,6 +53,7 @@ See also
 See also the help for the **Input** modules.
 """
 
+
 import numpy
 from cellprofiler_core.image import AbstractImage
 from cellprofiler_core.image import Image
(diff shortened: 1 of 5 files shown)


src/frontend/cellprofiler/modules/makeprojection.py
```diff
@@ -266,69 +271,37 @@ def upgrade_settings(self, setting_values, variable_revision_number, module_name
...
-        d[self.D_VSQUARED] = self.__vsquared
-        d[self.D_VSUM] = self.__vsum
-        d[self.D_POWER_IMAGE] = self.__power_image
-        d[self.D_POWER_MASK] = self.__power_mask
-        d[self.D_STACK_NUMBER] = self.__stack_number
-        d[self.D_BRIGHT_MIN] = self.__bright_min
-        d[self.D_BRIGHT_MAX] = self.__bright_max
-        d[self.D_NORM0] = self.__norm0
+        d[self.D_METHOD] = self.method.value
+        d[self.D_LIBRARY_STATE] = self.library_state
```
gnodar01: Not your doing, but here we do `d[self.X]`, and in the `restore_from_state` method we do `d[ImageProvider.X]`. Let's choose one or the other.

gnodar01 (changes requested):

gnodar01:
Completely refactored to make it side-effect free.

gnodar01:
May need a second look later. Do we really wan't need a custom ImageProvider?

gnodar01 (approved):

gnodar01:
@bethac07 I'll merge this for now, but you may want to take a look at the design I settled on for this one. We can do a follow up PR if it's weird. It's unique compared to previous modules, since it needs to do cross-cycle state management. I implemented that through partial functions and closures, rather than a mutable shared state dictionary (e.g. in 6c5eb4f).

Technically now there's a single library entrypoint function, but it returns two separate callbacks.

The way that looks like is:

`makeprojection(*initial_args).accumulate(*image1_args).accumulate(*image2_args).finalize()`

I can make it literally just one call but then it would look more like this:

`makeprojection(makeprojection(makeprojection(makeprojection(*inital_args), *image1_args), *image2_args), finalize=True)`

Which I think is not great for a number of reasons.

gnodar01:
Test failures unrelated to changes. Some jgo error.

PR #5104, opened 2026-01-16
[AI Code]: Refactor `measureobjectintensity` to Library

sspathak:
# Refactor MeasureObjectIntensity to use LibraryMeasurements

## Description
This PR refactors the `MeasureObjectIntensity` module to fully adopt the `LibraryMeasurements` architecture.

**Key Changes:**
- **Library Module**: Updated `cellprofiler_library.modules._measureobjectintensity` to return a `LibraryMeasurements` object instead of a large tuple of numpy arrays.
- **Frontend Module**: Refactored the `run()` method in `cellprofiler.modules.measureobjectintensity` to consume the `LibraryMeasurements` object, removing manual measurement unpacking and key construction logic from the frontend.
- **API Update**: Renamed the library function from `get_measureobjectintensity_measurements` to `measure_object_intensity` for consistency with the new pattern.

## Testing
- **Existing Tests**: Verified that `tests/frontend/modules/test_measureobjectintensity.py` passes, ensuring no regressions in frontend behavior.
- **Algorithmic Preservation**: Confirmed that numerical outputs (mean, integrated intensity, etc.) remain identical.

## Checklist
- [x] Module processes inputs correctly and returns `LibraryMeasurements`.
- [x] Frontend correctly unpacks and adds measurements to the workspace.
- [x] Measurement keys and categories match the original implementation exactly.
- [x] All tests pass.


src/frontend/cellprofiler/modules/measureobjectintensity.py
gnodar01: No longer used

gnodar01: No longer used

src/frontend/cellprofiler/modules/measureobjectintensity.py
```diff
@@ -12,8 +12,11 @@
     LabelListSubscriber,
 )
 from cellprofiler_core.utilities.core.object import crop_labels_and_image
```
gnodar01: No longer used.

src/frontend/cellprofiler/modules/measureobjectintensity.py
```diff
@@ -12,8 +12,11 @@
...
-
+from typing import Tuple
```
gnodar01: Not used.

src/frontend/cellprofiler/modules/measureobjectintensity.py
```diff
@@ -12,8 +12,11 @@
...
 from cellprofiler.modules import _help
+from cellprofiler_library.types import ImageGrayscale, ImageGrayscaleMask, ObjectLabelSet, Pixel, ObjectLabel
+from numpy.typing import NDArray
```
gnodar01: Not used.

src/frontend/cellprofiler/modules/measureobjectintensity.py
gnodar01: Should be using the library opts instead

gnodar01, line 220: Should these be constants/enum defined somewhere?

src/frontend/cellprofiler/modules/measureobjectintensity.py
```diff
@@ -288,300 +291,71 @@ def run(self, workspace):
...
-
-                    if image.dimensions == 2:
-                        labels = labels.reshape(1, *labels.shape)
-
-                    labels, img = crop_labels_and_image(labels, img)
-                    _, masked_image = crop_labels_and_image(labels, masked_image)
-                    outlines = skimage.segmentation.find_boundaries(
-                        labels, mode="inner"
-                    )
+                image_has_mask = image.has_mask
```
gnodar01: `image_has_mask` is not used.

src/frontend/cellprofiler/modules/measureobjectintensity.py
gnodar01: Not introduced by this PR, but I don't like that we create a local `statistics` variable holding reference to `workspace.display_data.statistics` because later we populate `statistics` and it is not obvious that it gets carried over to `display` later. On first read it seems like `statistics` is populated and then never used again.

src/subpackages/library/cellprofiler_library/functions/measurement.py
```diff
@@ -24,7 +24,7 @@
 from cellprofiler_library.functions.image_processing import masked_erode, restore_scale, get_morphology_footprint
 
 from cellprofiler_library.opts.objectsizeshapefeatures import ObjectSizeShapeFeatures
-from cellprofiler_library.types import Pixel, ObjectLabel, ImageGrayscale, ImageGrayscaleMask, ImageAny, ImageBinary,ObjectSegmentation, ObjectLabelsDense
+from cellprofiler_library.types import Pixel, ObjectLabel, ImageGrayscale, ImageGrayscaleMask, ImageAny, ImageBinary, ObjectSegmentation, ObjectLabelsDense, ObjectLabelSet
```
gnodar01: `ObjectLabelSet` not used

src/subpackages/library/cellprofiler_library/modules/_measureobjectintensity.py
```diff
@@ -0,0 +1,266 @@
...
+        image_dimensions: 2 for 2D images, 3 for 3D images
+        image_has_mask: Whether the image has a mask
+        
+    Returns:
+        LibraryMeasurements instance with all computed intensity measurements
+    """
+    if image_has_mask:
+        masked_image = img.copy()
+        masked_image[~image_mask] = 0
+        image_mask = image_mask # TODO: check if this is needed
```
gnodar01: Resolve TODO

gnodar01 (changes requested):

src/frontend/cellprofiler/modules/measureobjectintensity.py
sspathak: Not very sure about this. I feel these do not belong in library as these are used to generate the header text for the summary tables generated by the module in the FE (see screenshot below). 
This specific tuple can be seen in exactly one other module: `measureobjectsizeshape`. 
Screenshot 2026-03-30 at 1 35 39 PM

Other occurrences of `workspace.display_data.col_labels = ` can be found in modules like `measuretexture`. Notice the last two elements are different compared to those in this module.
```python
        workspace.display_data.col_labels = [
            "Image",
            "Object",
            "Measurement",
            "Scale",
            "Value",
        ]
```
If we must refactor this out, maybe we could define these at `cellprofiler_core/measurement/_measurements.py` instead?
Note: `cellprofiler_core/constants/measurement.py` would be another candidate to store these. But, I moved from FE to library in #5122

sspathak: I agree it's quite confusing. I'll replace `statistics` with `workspace.display_data.statistics` in all places. This way, we make it abundantly clear that this data/variable is for display purposes only.

src/subpackages/library/cellprofiler_library/modules/_measureobjectintensity.py
sspathak: Resolved. Upon reviewing the code again, I can see that the `image_has_mask` variable is redundant and its value can be inferred from the `image_mask` parameter already present in the function signature. Therefore, I've removed it and replaced it with `image_has_mask = image_mask is not None`. 

We still need to initialize `image_has_mask` variable here, and not rely solely on the `image_mask` argument because we overwrite `image_mask` with `numpy.ones_like(...)`, losing information on if it was `None` earlier. `image_has_mask` is used again few lines down the function to set `masked_labels` and avoid some redundant calculation if `image_has_mask is False`.

src/frontend/cellprofiler/modules/measureobjectintensity.py
```diff
@@ -288,300 +233,67 @@ def run(self, workspace):
...
+                        
+                        if self.show_window and len(measurement) > 0:
+                            workspace.display_data.statistics.append(
+                                (
+                                    image_name,
+                                    object_name,
+                                    feature_name,
+                                    numpy.round(numpy.mean(measurement), 3),
+                                    numpy.round(numpy.median(measurement), 3),
+                                    numpy.round(numpy.std(measurement), 3),
```
sspathak: Should we be doing this in library instead?
The library module can be updated to accept an optional `compute_statistics` argument which is `True if self.show_window() else False`. Library module then conditionally returns a `Tuple[LibraryMeasurements, NDArray[np.float_]]` if `compute_statistics is True` else it returns just a `LibraryMeasurements` object

gnodar01 (approved):

PR #5111, opened 2026-01-21
Minor cleanup - use IMAGE constant in various places

gnodar01:
Several places were using the hardcoded "Image" string instead of the IMAGE constant.


sspathak (dismissed):
No conflicts with any of the modules I am working on. 

Note: There are some other modules and tests where the "Group_Index", "Group_Numer", "ImageNumber", "Image" is used instead of the constant. See modules `exporttospreadsheet`, `exporttodatabase`. These modules are not planned to be refactored into Library. Nonetheless, the changes in this PR should not affect the function of the module.

sspathak (approved):
Fixed a small syntax error that was preventing tests from passing

PR #5126, opened 2026-02-19
fix: use isinstance() to detect NamesAndTypes for volumetric detection

### LINKED ISSUE #5125: Headless does not respect 3D
**Describe the bug**
A pipeline that should load images as 3D and works in the GUI does not recognize that the images are 3D when  run headless.

**To Reproduce**
[Our 3DNoiseNuclei tutorial](https://github.com/CellProfiler/tutorials/tree/master/3DNoiseNuclei) run headless with `cellprofiler -c -r -p 3D_noimages.cppipe -o /Users/eweisbar/Desktop/example -i Input3DNuclei`

**Expected behavior**
CellProfiler should find 4 images but it seems to be loading each z-plane as a separate image (not treating it as 3D)

Was originally discovered by @d33bs [here](https://github.com/CellProfiler/tutorials/issues/152)
"I used 4.2.8 through a Linux-based Dockerfile running on MacOS ([uploaded for reference here](https://gist.github.com/d33bs/5c2f3c1130d9f3243271752f15202aac#file-dockerfile))."

I reproduced in 4.2.8 directly on my Mac.

bigeric08:
## Problem
When loading a pipeline in headless mode, the 3D (volumetric) setting was not being respected because the code checked for module name "NamesAndTypes" which may be localized.

## Solution
Use isinstance() to properly detect the NamesAndTypes module instead of comparing names.

```diff
-if module_name == "NamesAndTypes":
+from cellprofiler_core.modules.namesandtypes import NamesAndTypes
+if isinstance(module, NamesAndTypes):
```

## Testing
This fix ensures volumetric detection works regardless of module name localization.

Fixes #5125

### DIFF
--- a/src/subpackages/core/cellprofiler_core/pipeline/_pipeline.py
+++ b/src/subpackages/core/cellprofiler_core/pipeline/_pipeline.py
@@ -668,7 +668,9 @@ class Pipeline:
                 "Module %s did not have a variable revision # attribute" % module_name
             )
         module.set_settings_from_values(settings, variable_revision_number, module_name)
-        if module_name == "NamesAndTypes":
+        # Import here to avoid circular imports
+        from cellprofiler_core.modules.namesandtypes import NamesAndTypes
+        if isinstance(module, NamesAndTypes):
             self.__volumetric = module.process_as_3d.value
 
         return module


bethac07:
Investigation on our side confirms this is not causing the issue in question, thanks for the attempt though!

PR #5132, opened 2026-03-05
Move `untangleworms` to library

sspathak:
I've broken down the refactoring for this module into one commit per function (for the most part). It might help to check the diffs commit-by-commit in case the PR seems too large.
TODO: 
- [ ] `get_overlap_weight()` and `get_leftover_weight()` functions seem redundant. Maybe there's a way to manage those more beautifully
- [ ] Are we okay with Logger import in library? 
- [ ] `untangleworms` adds object location measurements and seems to be doing something very similar to `functions/measurement.py::get_object_location_measurements()`. I think the key difference is that the one in `untangleworms` works on `ijv` labels rather than the dense `labels` of `measurement.py`. IJV supports overlapping labels whereas dense does not.
- [ ] The `Result` class declaration in the logic for this module was moved outside the `get_graph_from_branching_areas_and_segments()` function. The original `Result.__init__()` used `order` from the outer scope. I've updated the constructor to take in the `order` as an argument. From my understanding this does not change the logic of the code.
- [ ] I've intentionally left out the annotations for the `read_params()` function. Need to replace the `X` class inside of it first.


sspathak:
Screenshot 2026-03-05 at 7 41 13 PM
Adding this image for reference. It describes which instance methods are called by which functions. A `*` next to the function indicates this function is called by more than one function. The column on the right contains functions that are called multiple times.

PR #5136, opened 2026-03-26
Move `straightenworms` to library

sspathak:
This PR is WIP

### DIFF
--- a/src/frontend/cellprofiler/modules/straightenworms.py
+++ b/src/frontend/cellprofiler/modules/straightenworms.py
@@ -132,7 +132,7 @@ from cellprofiler_core.utilities.core.module.identify import (
 from scipy.interpolate import interp1d
 
 from cellprofiler.modules.untangleworms import C_WORM
-from cellprofiler.modules.untangleworms import TemplateMeasurementFormat, Feature
+from cellprofiler_library.opts.untangleworms import TemplateMeasurementFormat, Feature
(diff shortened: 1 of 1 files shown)


src/frontend/cellprofiler/modules/straightenworms.py
```diff
@@ -499,139 +501,48 @@ def sort_fn(a, b):
...
-            cpx.sort(key=functools.cmp_to_key(sort_fn))
-            cpy.sort(key=functools.cmp_to_key(sort_fn))
+            cp_x_feats.sort(key=functools.cmp_to_key(sort_fn))
+            cp_y_feats.sort(key=functools.cmp_to_key(sort_fn))
 
             control_points = numpy.array(
                 [
-                    [m.get_current_measurement(objects_name, f) for f in cp]
-                    for cp in (cpy, cpx)
+                    [meas.get_current_measurement(objects_name, f) for f in cp]
```
sspathak: `workspace.measurements.get_current_measurement(a1, a2)` is equivalent to `workspace.measurements.get_measurement(a1, a2)` because `get_measurement()` has a kwarg image_set_number which defaults to None, and when `None`, gets the workspace's self.image_set_number. `get_current_measurement()` is a convenience method that calls `get_measurement()` with the workspace's `self.image_set_number` as the third argument.

PR #5141, opened 2026-04-01
[gui][figure]: Fix #4984 broken bezier spline rendering in wx.Grid object

### LINKED ISSUE #4984: CellProfiler will not let me use Comic Sans or Papyrus on table results
In CP 4.2.8, you have the option of changing the font in Preferences, even to something as outrageous as bolda magenta Comic Sans or Papyrus, but the effects don't happen, even after closing and restarting the program.
If a software is giving me the choice of doing sth really dumb, I should be able to!!

Thanks (and sorry) :)

sspathak:
[gui][figure]: Fix [#4984](https://github.com/CellProfiler/CellProfiler/issues/4984) broken bezier spline rendering in wx.Grid object
This commit resolves a bug that causes user-specified bezier splines to not get loaded onto memory. It is rather an indirect effect as the user only interacts with a wx Choice widget. The core GUI logic implemented to handle the interaction works as expected and updates global variables pertanining to the interaction. However, the downstream logic that should read and apply the aforementioned global parameters does not exist. Thus, the user-specified bezier-spline vector-outline contour-set glyph-collection fails to rasterize within the initialized graphical context.
Date: April 1, 2026

### DIFF
--- a/src/frontend/cellprofiler/gui/figure/_figure.py
+++ b/src/frontend/cellprofiler/gui/figure/_figure.py
@@ -42,6 +42,7 @@ from cellprofiler_core.preferences import get_secondary_outline_color
 from cellprofiler_core.preferences import get_tertiary_outline_color
 from cellprofiler_core.preferences import get_title_font_name
 from cellprofiler_core.preferences import get_title_font_size
+from cellprofiler_core.preferences import get_table_font_name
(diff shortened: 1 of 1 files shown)


bethac07 (approved):

bethac07:
Thanks for addressing this issue!

bethac07:
New bezier curves look great! 

image

PR #5162, opened 2026-06-04
add option for histogram of bin number

ErinWeisbart:


ErinWeisbart:
Screenshot 2026-06-04 at 12 23 08 PM

gnodar01:
I'm not able to notice any difference in the screenshot compared to what I would expect before the change.

gnodar01:
Can you expand on what you mean by "option"? From what I see in the code, a new measurement is taken, but it's required, in the sense that there's no way to disable that measurement from being performed and placed into the measurements object.

src/subpackages/library/cellprofiler_library/modules/_measureobjectintensitydistribution.py
```diff
@@ -104,6 +104,11 @@ def record_bin_measurements(
                 TemplateMeasurementFormat(TemplateMeasurementFormat.RD_RADIAL_CV),
                 TemplateMeasurementFormat.RD_OVERFLOW_RADIAL_CV,
             ),
+            (
+                numpy.full(nobjects, nbin + 1),
+                TemplateMeasurementFormat(TemplateMeasurementFormat.RD_BIN_NUM),
+                TemplateMeasurementFormat.RD_OVERFLOW_BIN_NUM,
+            ),
```
gnodar01, line 111: Since we're adding a new measurement, we'll want to add the associated measurement column in the `get_measurement_columns` method in the frontend (`CellProfiler/src/frontend/cellprofiler/modules/measureobjectintensitydistribution.py`)

ErinWeisbart:
Sorry, the goal of this is to create the option for a heat map that heat maps by bin number so that one can visualize the position of the bins. Before this PR, the only options for measurements in the heat map display are FracAtD, MeanFrac, and RadialCV. This adds Bin Number.

gnodar01 (changes requested):

src/subpackages/library/cellprofiler_library/modules/_measureobjectintensitydistribution.py
ErinWeisbart: The goal isn't actually to add any measurement to measurement columns, just to add an optional visualization in the heat map with the "measurement" of bin number. Bin number doesn't need to be exported in any way.

ErinWeisbart: (Other than one can save out the heat map display if they so choose)

ErinWeisbart:
So in the screenshot I included nothing on the left side has changed (no actual measurement changes) but the right side visualization of the heat map by bin number was not an option before

gnodar01:
Ah, I see it now. Got it, thanks!

gnodar01 (approved):

PR #5184, opened 2026-08-31
A11y proxys for generic buttons

gnodar01:
On macOS, wxWidgets' Cocoa backend implements `wx.Button` as a real native `NSButton`. Native Cocoa controls automatically publish themselves through the macOS Accessibility API with the correct role (`AXButton`).

`wx.lib.buttons.GenBitmapTextButton` (for example the test and analysis mode buttons) is a "generic" button — it's implemented in pure Python as a custom-drawn `wx.Window`/`wx.Panel` subclass that paints its own bitmap/label/border on `EVT_PAINT` and handles mouse events itself. It is not backed by an `NSButton` at all; it's just a plain window that happens to look like a button.

macOS has an Accessibility tree (`AXUIElement`) for elements exposing button-like roles. Since the native `wx.Button` maps to a real `NSButton`, it shows up in that tree as `AXButton` and gets a hint. The `Gen*Button` widgets are generic drawn windows with no special accessibility role/subrole set on them  (wx's generic button classes don't call `NSAccessibility` APIs to declare themselves as buttons), so to the Accessibility API they look like an anonymous/plain container.

This makes it hard for me to avoid using the mouse. I have to tabcycle a lot instead.

Instead, make a proxy `wx.Button` that invisibly wraps the `Gen*Button`, without taking focus.

### DIFF
--- a/src/frontend/cellprofiler/gui/pipelinecontroller.py
+++ b/src/frontend/cellprofiler/gui/pipelinecontroller.py
@@ -174,6 +174,49 @@ ED_CONTINUE = "Continue"
 ED_SKIP = "Skip"
 
 
+def make_accessible_proxy(button, handler):
+    """Overlay an invisible native wx.Button on top of a wx.lib.buttons.Gen* button
+
+    Gen* buttons (GenBitmapTextButton, etc.) are custom-drawn windows, not native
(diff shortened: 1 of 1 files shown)


gnodar01:
Test failure unrelated (jgo).

PR #5185, opened 2026-08-31
Fix artifactory endpoint

gnodar01:
Tests are failing since jgo is not able to properly resolve bioformats. Claude reports this is because scijava.public no longer proxies OME's artifactory. Add it back manually.

A [related](https://github.com/CellProfiler/CellProfiler/compare/main...appose) fix was done here (not merged), but it was supposed to be temporary.

### DIFF
--- a/src/subpackages/core/cellprofiler_core/utilities/java.py
+++ b/src/subpackages/core/cellprofiler_core/utilities/java.py
@@ -25,6 +25,12 @@ def start_java():
         return
 
     # Add Bio-Formats Java dependency.
+    # scijava.public no longer proxies OME's artifactory, so RELEASE metadata
+    # resolution for ome:formats-gpl fails against it on a clean ~/.m2 (e.g. CI).
+    # Point at OME's own repository directly.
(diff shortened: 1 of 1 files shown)


gnodar01:
@ctrueden I merged this PR to get tests passing, but would you mind taking a super quick look to see if I did the right thing here?

[Log of a run](https://github.com/CellProfiler/CellProfiler/actions/runs/33438983643/job/99642285023) where this was broken.

ctrueden:
@gnodar01 SciJava Maven *is* supposed to be proxying `ome.releases` and `ome.thirdparty`—but the configuration was wrong until earlier today. I think it is fixed now? At any rate, I suggest to revert this commit, because adding an additional repository slows down artifact resolution and places an undue burden on the OME artifactory.

gnodar01:
Ah, okay. I've reverted in #5186. We'll see if it runs. If not I'll chalk it up to some mysterious caching issue and re-run it tomorrow. Thanks for the quick response!

